PR #2, opened 2019-12-03
Play seed with jenkins

MancunianSam:


Jenkinsfile-build
```diff
@@ -0,0 +1,34 @@
...
+        agent {
+            ecs {
+                inheritFrom 'ecs'
+            }
+        }
+        steps {
+            checkout scm
+            sh 'sbt dist'
+            stash includes: 'Dockerfile', name: 'Dockerfile'
+            stash includes: 'target/universal/tdr-transfer-frontend-1.0-SNAPSHOT.zip', name: 'tdr-transfer-frontend-1.0-SNAPSHOT.zip'
```
suzannehamilton: Today I learned about `stash includes`. :+1: 

Can we use a wildcard so it doesn't depend on a specific version number? Ditto for the `unstash` step below.

Jenkinsfile-build
```diff
@@ -0,0 +1,34 @@
+pipeline {
```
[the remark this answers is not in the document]
MancunianSam: It's a fair point, plus this job will probably update the ecs service to run the new image as well so deploy makes more sense there. I've changed it.

Jenkinsfile-build
```diff
@@ -0,0 +1,34 @@
...
+    stage('Docker') {
+            agent {
+                label 'master'
+            }
+            steps {
+                unstash 'tdr-transfer-frontend-1.0-SNAPSHOT.zip'
+                unstash 'Dockerfile'
+                sh 'docker build -t nationalarchives/tdr-transfer-frontend .'
+                withCredentials([usernamePassword(credentialsId: 'docker', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
+                    sh 'echo $PASSWORD | docker login --username $USERNAME --password-stdin'
```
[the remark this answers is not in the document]
MancunianSam: It would obfuscate it because it comes from a password variable but this command wouldn't print it anyway, the output of echo is piped to the docker login command.

Jenkinsfile-build
MancunianSam: The includes can use a wildcard which I've changed. The name can be anything you like which I didn't know so I can call it something without the version and reference that in the unstash step.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM openjdk:8-slim
+WORKDIR play
+COPY target/universal/tdr-transfer-frontend-1.0-SNAPSHOT.zip .
+RUN apt-get update && apt-get install unzip && unzip -qq tdr-transfer-frontend-1.0-SNAPSHOT.zip
+CMD AWS_ACCESS_KEY_ID=$ACCESS_KEY_ID \
+      AWS_SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY \
+      transfer-digital-records-1.0.1-SNAPSHOT/bin/transfer-digital-records \
+      -Dplay.http.secret.key=$PLAY_SECRET_KEY \
+      -DTDR_GRAPHQL_URI=$TDR_GRAPHQL_URI \
```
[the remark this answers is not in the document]
MancunianSam: This and the sendgrid one were me being lazy and copying from the prototype docker file. I've removed them all now apart from PLAY_SECRET_KEY, as you say, we can add them as an when we need them if we need them.

Jenkinsfile-build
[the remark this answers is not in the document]
suzannehamilton: Cheers. :+1:

Jenkinsfile-build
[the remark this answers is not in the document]
suzannehamilton: :+1:

Jenkinsfile-deploy
```diff
@@ -5,7 +5,7 @@ pipeline {
     stage('Test') {
         agent {
             ecs {
-                inheritFrom 'ecs'
+                inheritFrom 'sbt'
```
suzannehamilton: Does the other Jenkinsfile need this line as well?

Out of curiosity, where is this template defined?

MancunianSam: I've changed it in the other file. They're defined in the Jenkins project here https://github.com/nationalarchives/tdr-jenkins/tree/master/docker. I've started with one folder per template.

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM openjdk:8-slim
+WORKDIR play
+COPY target/universal/tdr-transfer-frontend*.zip .
+RUN apt-get update && apt-get install unzip && unzip -qq tdr-transfer-frontend-1.0-SNAPSHOT.zip
```
suzannehamilton: The version on line 3 is a wildcard but it's hard-coded on line 4.

Jenkinsfile-deploy
```diff
@@ -0,0 +1,59 @@
...
+    }
+    stage("Docker") {
+        agent {
+            label "master"
+        }
+        steps {
+            unstash "tdr-transfer-frontend-zip"
+            unstash "Dockerfile"
+            sh "docker build -t nationalarchives/tdr-transfer-frontend:${params.STAGE} ."
+            withCredentials([usernamePassword(credentialsId: "docker", usernameVariable: "USERNAME", passwordVariable: "PASSWORD")]) {
```
suzannehamilton: Where do the `USERNAME` and `PASSWORD` environment variables come from, and should we be using something like the [docker-workflow](https://jenkins.io/doc/pipeline/steps/docker-workflow/) plugin to avoid referencing the credentials directly?

Jenkinsfile-deploy
```diff
@@ -0,0 +1,59 @@
...
+        agent {
+            ecs {
+                inheritFrom "aws"
+            }            
+        }
+        steps {
+            withCredentials([[
+                $class: "AmazonWebServicesCredentialsBinding",
+                credentialsId: "integrationTerraform",
+                accessKeyVariable: "AWS_ACCESS_KEY_ID",
```
suzannehamilton: I was talking to @TomJKing last week about how feasible it would be to use the [ECS task role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_IAM_role.html) to avoid having to generate an AWS access key for pushing things from Jenkins. What do you think?

conf/application.conf
```diff
@@ -1 +1,35 @@
 # https://www.playframework.com/documentation/latest/Configuration
+play.http.secret.key=${PLAY_SECRET_KEY}
+play.http.port=9000
+# The load balancer url is just while testing, it will be removed once we have the dns sorted
+play.filters.hosts {
+    allowed = [
+        "tdr-frontend-load-balancer-intg-1439549459.eu-west-2.elb.amazonaws.com",
+        "transfer-intg.nationalarchives.gov.uk",
+        "transfer-staging.nationalarchives.gov.uk",
+        "transfer-prod.nationalarchives.gov.uk",
```
suzannehamilton: This seems like it should be configured differently in each environment (including localhost), rather than whitelisting everything for all environments.

conf/application.conf
```diff
@@ -1 +1,35 @@
...
+        "10.0.0.21:9000",
+        "10.0.0.22:9000",
+        "10.0.0.23:9000",
+        "10.0.0.24:9000",
+        "10.0.0.25:9000",
+        "10.0.0.26:9000",
+        "10.0.0.27:9000",
+        "10.0.0.28:9000",
+        "10.0.0.29:9000",
+        "10.0.0.30:9000",
```
suzannehamilton: What are all these IP addresses?

Jenkinsfile-deploy
MancunianSam: USERNAME and PASSWORD are created temporarily by the withCredentials plugin. They only exist for this job inside that block. 
We could use something like that but it's a lot of overhead just to avoid printing out obfuscated credentials.

Jenkinsfile-deploy
MancunianSam: We could do, this user is the user that jenkins uses to terraform the environments which obviously has permissions to update ECS services so it's not like we're creating a new one for this purpose.

conf/application.conf
MancunianSam: Maybe, this is unlikely to even be the final DNS names to be honest, it was mostly meant as a placeholder to remind us to sort it at some point. I can take it out if you want because we'll remember to add it later anyway because it won't work without it.

conf/application.conf
MancunianSam: IP addresses from the subnets which the task runs in. These are created by terraform and are always the same but we don't know which one it's going to get in advance so we have to whitelist all of them. It doesn't look like you can whitelist a range in Play.

Jenkinsfile-deploy
MancunianSam: I've changed my mind, using the terraform user violates the principle of least privilege. I'll create a new user and add it to Jenkins for the ECS deployments.

MancunianSam: I've just realised what you mean in that we don't need a user at all, we can pass a role through to the fargate container which will then create a temporary user based on that role and use that to update the service. I've created these roles in the other terraform projects and this is now working without creating an explicit user. 
Also, this job now is not running using the terraform user which has far too many privileges, the policy against the role only allows the container to update the ECS service.

Jenkinsfile-deploy
MancunianSam: It's actually not a lot of effort to use the plugin and it's cleaner so I've added it back in.

Dockerfile
MancunianSam: Well spotted :smile:

conf/application.conf
MancunianSam: I've added in separate files for different environments. There's no prod yet but we can add that later. Which file we need is specified in the Dockerfile when it runs based on an environment variable passed down from ECS. I know this isn't the only way to do it but this seems quite clean.

Jenkinsfile-deploy
suzannehamilton: Great! 👍

Jenkinsfile-deploy
suzannehamilton: Cheers - locking down the role used by the tasks sounds good.

Is https://github.com/nationalarchives/tdr-terraform-backend/pull/8 the change you mean?

conf/application.conf
suzannehamilton: Oh, are the IP addresses used by the load balancer health checks? Could you add a comment explaining that (or whatever the real reason is, if I've misunderstood)? I don't think it's very obvious why the host isn't just the domain name.

suzannehamilton: Actually, I still don't understand why you need to include "localhost:9000" in the deployed environments. This might be a silly question, but what calls the app on localhost except in dev?

Jenkinsfile-deploy
MancunianSam: That's the one yeah.

conf/application.conf
MancunianSam: I didn't mean to, I've taken localhost out of the deployed environments conf files

suzannehamilton: 👍 

What do you think about adding a comment to help people understand why the list of IP addresses is there?

MancunianSam: I've added in a comment to the staging and intg ones.

suzannehamilton: Cheers. 🙂 

Sorry to keep dragging this out, but from your comment it sounds like the app won't actually start if the container's IP address isn't in this let. But the [Play docs](https://www.playframework.com/documentation/2.7.x/AllowedHostsFilter) suggest that `play.filters.hosts.allowed` is used when handling a request, rather than at startup. Is the app starting up OK but being shut down by the health check because the load balancer can't access it? Or am I misunderstanding the Play docs?

MancunianSam: Play will start but the load balancer uses the IP to do the health checks and it's these that are failing if the IP address isn't listed, which kills the task so the app never starts up properly. I've changed the description to mention the load balancer.

suzannehamilton: Thanks!

PR #10, opened 2020-01-09
Tk tdr home page

TomJKing:


Jenkinsfile-deploy
```diff
@@ -56,11 +64,17 @@ pipeline {
...
+        "staging": env.STAGING_ACCOUNT,
+        "prod": env.PROD_ACCOUNT
     ]
 
-    return stageToAccountMap.get(params.STAGE)
+    if (stageToAccountMap.get(params.STAGE)) {
+        accountNumber = stageToAccountMap.get(params.STAGE)
+    }
+
+    return accountNumber
```
suzannehamilton: Is it possible to use the Java map method `.getOrDefault` in Groovy to simplify this?

Also, do we want a default? It seems like it might be better for the job to fail if the environment is misconfigured, rather than assuming it means integration.

TomJKing: Good point about it should be failing if the environment is misconfigured

PR #11, opened 2020-01-10
Pac4j auth

MancunianSam:


app/modules/FrontendHttpActionAdaptor.scala
```diff
@@ -0,0 +1,19 @@
...
+class FrontendHttpActionAdaptor extends PlayHttpActionAdapter {
+
+  override def adapt(code: Int, context: PlayWebContext): Result = {
+    if (code == HttpConstants.UNAUTHORIZED) {
+      Results.unauthorized(views.html.index.render().toString()).as(HttpConstants.HTML_CONTENT_TYPE)
+    } else if (code == HttpConstants.FORBIDDEN) {
+      Results.forbidden(views.html.index.render().toString()).as(HttpConstants.HTML_CONTENT_TYPE)
+    } else {
+      super.adapt(code, context)
+    }
```
MancunianSam: This will change once we have some proper unauthorised pages but it's ok for a placeholder for now I think.

build.sbt
```diff
@@ -5,7 +5,7 @@ version := "1.0-SNAPSHOT"
 
 lazy val root = (project in file(".")).enablePlugins(PlayScala)
 
-scalaVersion := "2.13.1"
+scalaVersion := "2.12.8"
```
MancunianSam: Pac4j has no support for 2.13 as yet, there is an open issue.

app/modules/FrontendHttpActionAdaptor.scala
suzannehamilton: I've [added it to the backlog](https://github.com/orgs/nationalarchives/projects/20#card-31548460) so we don't forget. :+1:

app/modules/SecurityModule.scala
```diff
@@ -0,0 +1,59 @@
...
+  override def configure(): Unit = {
+
+    bind(classOf[PlaySessionStore]).to(classOf[PlayCacheSessionStore])
+    bind(classOf[SecurityComponents]).to(classOf[DefaultSecurityComponents])
+
+    bind(classOf[Pac4jScalaTemplateHelper[CommonProfile]])
+
+    // callback
+    val callbackController = new CallbackController()
+    callbackController.setDefaultUrl("/?defaulturlafterlogout")
```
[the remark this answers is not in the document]
MancunianSam: It should be /, I've changed it.

conf/application.conf
```diff
@@ -0,0 +1,16 @@
+# https://www.playframework.com/documentation/latest/Configuration
+play.http.port=9000
+
+auth.url="http://localhost:8080"
+auth.callback="http://localhost:9000/callback"
+auth.secret="set_me"
```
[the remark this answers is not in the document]
MancunianSam: There's no fallback. If you specify config.resource, it will use that file, if you don't, it will use this one. This won't work if we accidentally deploy it to prod without the config.resource param.

PR #12, opened 2020-01-14
Mock keycloak server and redis cache.

MancunianSam:
For an unauthorised user, the tests were trying to contact the keycloak
server which doesn't exist when the tests are running. I've mocked this
out in the unauthorised tests.
There were also exceptions relating to redis because there was no redis
instance to connect to. These weren't causing anything to fail but they
were annoying. I've mocked out the redis module so these exceptions
don't happen.

I've moved the overrides to the test helper trait and changed the tests
to use this.

### DIFF
--- a/test/util/FrontEndTestHelper.scala
+++ b/test/util/FrontEndTestHelper.scala
@@ -86,7 +86,6 @@ trait FrontEndTestHelper extends PlaySpec with MockitoSugar with Injecting with 
     // Mock the init method to stop it calling out to the keycloak server
     doNothing().when(configuration).init()
 
-
     // Mock the redirect return when the client is unauthorised
     val client = mock[KeycloakOidcClient]
     doAnswer(_ => "OidcClient").when(client).getName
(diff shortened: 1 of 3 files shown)


test/util/FrontEndTestHelper.scala
```diff
@@ -39,7 +55,47 @@ trait FrontEndTestHelper {
...
+    //There is a null check for the action adaptor.
+    testConfig.setHttpActionAdapter(new PlayHttpActionAdapter())
+
+    // There is a check to see whether an OidcClient exists. The name matters and must match the string passed to Secure in the controller.
+    val clients = new Clients()
+    val configuration = mock[OidcConfiguration]
+
+    // Mock the init method to stop it calling out to the keycloak server
+    doNothing().when(configuration).init()
+
```
TomJKing: Remove unnecessary line break

MancunianSam: :smile:

PR #23, opened 2020-02-03
add nvm + npm instructions to README.md

Lholbourn:
Added additional steps to ensure front-end aspects work when repo is pulled/cloned & inserted instructions on how to make CSS work if user does not have npm installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,15 @@ Repository for TDR transfer code
 
 * In the client settings, change the login theme to govuk.
 
+* Set "Root URL" to `http://localhost:9000`
+
+* Set "Valid redirect URIs" to `http://localhost:9000/*`
+
 * Generate a [secret](https://www.keycloak.org/docs/latest/server_admin/#_client-credentials)
 
-* Set the secret in auth.secret in [application.conf](conf/application.conf).
+* In the [application.conf](conf/application.conf) file find the auth.secret key and set its value to the secret you have just generated
+
+### User accounts
 
 * Create a new [user](https://www.keycloak.org/docs/latest/getting_started/index.html#_create-new-user) in the tdr realm.
 
@@ -24,9 +30,7 @@ Repository for TDR transfer code
 
 `docker run -d --name redis -p 6379:6379 redis`
 
-* Start the application using `sbt run`
-
-* If css is not displaying on localhost:9000/dashboard run `npm run build`
+### Static assets
 
 * If npm is not installed install [nvm](https://github.com/nvm-sh/nvm) in root directory.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,5 +26,16 @@ Repository for TDR transfer code
 
 * Start the application using `sbt run`
 
+* If css is not displaying on localhost:9000/dashboard run `npm run build`
```
suzannehamilton: I think it's fine to skip the "If css is not displaying". People will have to run this regularly anyway. :slightly_smiling_face:

README.md
```diff
@@ -26,5 +26,16 @@ Repository for TDR transfer code
...
+
+* If npm is not installed install [nvm](https://github.com/nvm-sh/nvm) in root directory.
```
suzannehamilton: I'd move the nvm/npm installation above the previous line, so that there's a clear order:

1. Set up nvm/npm if neccessary
2. Install npm dependencies
3. Build the assets

README.md
```diff
@@ -26,5 +26,16 @@ Repository for TDR transfer code
...
+
+* Once nvm is installed:
+    `nvm install 13.6`
+
+* `cd` into tdr-transfer-frontend in terminal
+
+* run  `npm install` again, then `npm run build`. CSS should now appear on localhost.
```
suzannehamilton: Just noticed that this README didn't actually say how to run the application.

Is it just `sbt run` at the moment, or do you need to set any environment variables? Either way, could you add another subsection to explain how to start Play?

Lholbourn:
> A few comments about making the setup even clearer, but otherwise this looks good. Thanks for doing this!

Thanks for checking it out! I'll let you know once I've made the changes 👍

README.md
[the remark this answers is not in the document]
Lholbourn: Added these 👌

Lholbourn: Removed wording 👌

README.md
Lholbourn: ordering of document changed to reflect these steps 👌

README.md
Lholbourn: No environmental variables needed. Instructions on how to run have now been moved to the correct place as well

README.md
```diff
@@ -14,17 +14,38 @@ Repository for TDR transfer code
...
+### Static assets
+
+* If npm is not installed install [nvm](https://github.com/nvm-sh/nvm) in root directory.
+
+* Once nvm is installed:
+    `nvm install 13.6`
+
+* `cd` into tdr-transfer-frontend in terminal
+
+* run  `npm install` again, then `npm run build`
```
suzannehamilton: We can drop the "again", because this is now the first time `npm install` is mentioned.

Lholbourn: Ah, I'll change that! 👌

PR #24, opened 2020-02-04
Tk series page

techncl:
1. Removed the hard-coded asterisk from the series page form label
2. Wrote a function to return a asterisk if input option is required
3. Added 3 tests in order to verify that function (mentioned above) works


public/stylesheets/main.css
```diff
@@ -0,0 +1 @@
...ast-child,.govuk-radios__item:last-of-type{margin-bottom:0}.govuk-radios__input{cursor:pointer;position:absolute;z-index:1;top:-2px;left:-2px;width:44px;height:44px;margin:0;opacity:0}.govuk-radios__label{display:inline-block;margin-bottom:0;padding:8px 15px 5px;cursor:pointer;-ms-touch-action:manipulation;touch-action:manipulation}.govuk-radios__label::before{content:"";box-sizing:border-box;position:absolute;top:0;left:0;width:40px;height:40px;border:2px solid currentColor;border-radius:50%;background:0 0}.govuk-radios__label::after{content:"";position:absolute;top:10px;left:10px;width:0;height:0;border:10px solid currentColor;border-radius:50%;opacity:0;background:currentColor}.govuk-radios__hint{display:block;padding-right:15px;padding-left:15px}.govuk-radios__input:focus+.govuk-radios__label::before{border-width:4px;box-shadow:0 0 0 4px #fd0}.govuk-radios__input:checked+.govuk-radios__label::after{opacity:1}.govuk-radios__input:disabled,.govuk-radios__input:disabled+.govuk-radios__label{cursor:default}.govuk-radios__input:disabled+.govuk-radios__label{opacity:.5}@media (min-width:40.0625em){.govuk-radios--inline:after{content:"";display:block;clear:both}.govuk-radios--inline .
```
suzannehamilton: @Lholbourn and I were discussing this file yesterday - should it be committed? There's going to be a lot of churn, and the Jenkins deployment will have to run npm anyway.

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -0,0 +1,27 @@
...
+import play.api.test.FakeRequest
+import play.api.test.Helpers._
+import util.FrontEndTestHelper
+
+class SeriesDetailsControllerSpec extends FrontEndTestHelper {
+  "SeriesDetailsController GET" should {
+
+    "render the series details page with an authenticated user" in {
+      val controller = new SeriesDetailsController(getAuthorisedSecurityComponents())
+      val home = controller.seriesDetails().apply(FakeRequest(GET, "/"))
```
suzannehamilton: Maybe I'm misunderstanding how these controller tests work, but why is this variable called `home`? And why do you make a request to "/" (which looks like the homepage route) to the SeriesDetailsController?

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -0,0 +1,27 @@
...
+      status(home) mustBe OK
+      contentType(home) mustBe Some("text/html")
+      contentAsString(home) must include ("seriesDetails.header")
+      contentAsString(home) must include ("seriesDetails.title")
+    }
+
+    "return a redirect to the auth server with an unauthenticated user" in {
+      val controller = new SeriesDetailsController(getUnauthorisedSecurityComponents())
+      val home = controller.seriesDetails().apply(FakeRequest(GET, "/"))
+      redirectLocation(home).get should include ("/auth/realms/tdr/protocol/openid-connect/auth")
```
suzannehamilton: Very minor thing, but is there a ScalaTest matcher you can use to avoid calling `.get` (which will cause a confusing error message if it fails)?

conf/application.conf
```diff
@@ -3,7 +3,7 @@ play.http.port=9000
 
 auth.url="http://localhost:8080"
 auth.callback="http://localhost:9000/callback"
-auth.secret="set_me"
+auth.secret="[set me]"
```
suzannehamilton: I know this isn't really your change, but is it actually necessary to set this secret? I think it gets overridden in all of the deployed environments, and it doesn't matter what you set it to in development.

Maybe setting it to `dev-environment-placeholder` might make it clearer, so that it's obvious that it should only be used in development and you don't have to change it when you set up the project?

conf/routes
```diff
@@ -6,6 +6,8 @@
 # An example controller showing a sample home page
 GET     /                           controllers.HomeController.index
 GET     /dashboard                  controllers.DashboardController.dashboard
+GET     /seriesDetails              controllers.SeriesDetailsController.seriesDetails
+POST    /seriesDetails/submit       controllers.SeriesDetailsController.seriesSubmit()
```
suzannehamilton: Not a big deal, but why do we have parentheses in the POST route but not the GET one?

app/controllers/SeriesDetailsController.scala
```diff
@@ -0,0 +1,32 @@
...
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, Request}
+
+@Singleton
+class SeriesDetailsController @Inject()(val controllerComponents: SecurityComponents)
+                                         extends Security[CommonProfile] with I18nSupport  {
+
+  val form = Form(
+    mapping(
+      "seriesNo" -> number
```
suzannehamilton: Maybe call it seriesId or seriesName rather than seriesNo, since it's not a number? How do archivists refer to it?

app/controllers/SeriesDetailsController.scala
```diff
@@ -0,0 +1,32 @@
...
+    )(SelectedSeriesData.apply)(SelectedSeriesData.unapply)
+  )
+
+  val mockedData: Seq[(String, String)] = Seq(("series1", "Series 1"), ("series2", "Series 2"), ("series3", "Series 3"))
+
+  def seriesDetails(): Action[AnyContent] = Secure("OidcClient") { implicit request: Request[AnyContent] =>
```
suzannehamilton: Is there a way of commonising this `Secure("OidcClient")` wrapper so we don't have to set a String every time we add  a new controller action?

app/views/inputDropdown.scala.html
```diff
@@ -0,0 +1,27 @@
...
+    <label class="govuk-label" for="@elements.id">
+        @elements.args.get('_label)
+        @elements.args.requiredLabelSuffix()
+    </label>
+    <select
+            class="govuk-select"
+            id="@elements.id"
+            name="@elements.id"
+            @elements.args.requiredInput()>
+        <option value="" selected>Please choose...</option>
```
suzannehamilton: Did Pia have wording for this on her wireframes? I'm just wondering if something like "Select a series" might be more direct.

app/views/inputDropdown.scala.html
```diff
@@ -0,0 +1,27 @@
...
+
+        @for(dropdownOption <- dropdownOptions) {
+            <option value="@dropdownOption._1">@dropdownOption._2</option>
```
suzannehamilton: Are you planning to change this to a case class when we have real data? Tuples are generally a bit hard to follow.

app/views/inputDropdown.scala.html
TomJKing: This is the wording that is used on the wireframes: https://projects.invisionapp.com/share/S4TVATFN82M#/screens/383373576

conf/application.conf
TomJKing: In the local environment it needs to be set to the tdr client crendentials secret that is setup in the local keycloak application.

app/views/inputDropdown.scala.html
TomJKing: Yes. At the moment not clear what the data will be. Added comment to the template about this.

conf/application.conf
suzannehamilton: Should it be read from an environment variable, then? Otherwise (as far as I can tell) you'll have changes that you have to avoid committing.

app/views/inputDropdown.scala.html
suzannehamilton: OK, let's leave it then. It just sounds odd to me. :slightly_smiling_face: 

And I can't find any examples in GOV.UK because [they usually prefer radio buttons](https://design-system.service.gov.uk/components/select/), which might not be appropriate if we have to show any long series lists.

conf/routes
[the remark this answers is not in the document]
TomJKing: What path scheme do you think we should adopt?

conf/messages
```diff
@@ -7,4 +7,7 @@ home.use.transfer.resume=resume a transfer
 home.use.transfer.history=view your transfer history
 home.start=Start now
 dashboard.header=Dashboard
-dashboard.title=Welcome to the dashboard
\ No newline at end of file
+dashboard.title=Welcome to the dashboard
+seriesDetails.header=SeriesDetails
```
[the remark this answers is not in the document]
TomJKing: Resolved

test/controllers/SeriesDetailsControllerSpec.scala
TomJKing: Resolved

app/controllers/SeriesDetailsController.scala
TomJKing: Tried to do something here but not sure if its any good

conf/routes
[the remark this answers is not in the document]
suzannehamilton: Just discussed in person. Let's change it to this for now:

```
GET   /series
POST  /series
```

Could you put a card in the tech backlog so we can plan the routes soon?

conf/application.conf
```diff
@@ -3,7 +3,8 @@ play.http.port=9000
 
 auth.url="http://localhost:8080"
 auth.callback="http://localhost:9000/callback"
-auth.secret="set_me"
+auth.secret="secretValue"
+auth.secret=${?AUTH_SECRET}
```
suzannehamilton: What will happen if you forget to set the environment variable? Is the error obvious, or does login just mysteriously fail?

app/controllers/SeriesDetailsController.scala
```diff
@@ -0,0 +1,39 @@
...
+
+@Singleton
+class SeriesDetailsController @Inject()(val controllerComponents: SecurityComponents)
+                                         extends Security[CommonProfile] with I18nSupport  {
+
+  private val secureAction = Secure("OidcClient")
+
+  val selectedSeriesForm = Form(
+    mapping(
+      "seriesId" -> number
```
suzannehamilton: How does `number` work here, given that the mock IDs are strings like `mockedSeries1`?

app/controllers/SeriesDetailsController.scala
suzannehamilton: Yep, that's the kind of thing I had in mind (assuming it's fine to use the same `Secure` instance for each request). You could also move it to its own file so it can be reused in other controllers, or we can do that later.

conf/application.conf
TomJKing: Error: [CompletionException: org.pac4j.core.exception.TechnicalException: Bad token response, error=unauthorized_client]

conf/routes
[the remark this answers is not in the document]
TomJKing: Card added and routes changed

conf/application.conf
suzannehamilton: That's probably good enough for now. If it trips people up, we can put some more checks around it.

conf/routes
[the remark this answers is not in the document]
suzannehamilton: :+1:

PR #27, opened 2020-02-11
Tag jenkins builds with build number

suzannehamilton:
Add a version tag to every master build so that we can specify which version to deploy to each AWS environment.

This requires the jenkins build to use SSH to connect to github.


suzannehamilton:
Yes, I'll write it all up when it's done. This just adds a version tag - we still need to update the deployment jobs so that they update the `integration` branch.

PR #38, opened 2020-02-26
Series details submit

MancunianSam:


app/viewsapi/FormFunctions.scala
```diff
@@ -1,5 +1,8 @@
 package viewsapi
 
+import play.twirl.api.Html
```
TomJKing: Should add some unit tests for the form functions

app/controllers/ErrorController.scala
```diff
@@ -0,0 +1,14 @@
...
+import javax.inject.{Inject, Singleton}
+import play.api.i18n.{I18nSupport, MessagesApi}
+import play.api.mvc.{AbstractController, Action, AnyContent, ControllerComponents, Request}
+
+import scala.concurrent.ExecutionContext
+
+@Singleton
+class ErrorController @Inject()(val cc: ControllerComponents)(implicit val ec: ExecutionContext)  extends AbstractController(cc) with I18nSupport {
+  def error(message: String): Action[AnyContent] = Action { implicit request: Request[AnyContent] =>
+    Ok(views.html.error(""))
```
suzannehamilton: Should the `Ok` be a server error, or handle multiple types of errors? It's a bit confusing to return a 200 OK if there's an error, and it can lead to things like error pages being cached.

app/controllers/SeriesDetailsController.scala
```diff
@@ -19,33 +25,59 @@ class SeriesDetailsController @Inject()(val controllerComponents: SecurityCompon
...
     )(SelectedSeriesData.apply)(SelectedSeriesData.unapply)
   )
 
-  def seriesDetails(): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
-    val userTransferringBody = keycloakConfiguration.verifyToken(request.token.getValue)
-      .map(t => t.getOtherClaims.get("body").asInstanceOf[String])
-    val variables: Variables = new Variables(userTransferringBody)
+  private def getSeriesDetails(request: Request[AnyContent], status: Status, form: Form[SelectedSeriesData])(implicit requestHeader: RequestHeader) = {
+    val userTransferringBody: Option[String] = keycloakConfiguration.verifyToken(request.token.getValue)
+      .flatMap(t => t.getOtherClaims.asScala.get("body").asInstanceOf[Option[String]])
```
suzannehamilton: `asInstanceOf[Option[String]]` looks much nicer than the empty string alternative, thanks. :+1:

app/controllers/SeriesDetailsController.scala
```diff
@@ -19,33 +25,59 @@ class SeriesDetailsController @Inject()(val controllerComponents: SecurityCompon
...
+  def seriesSubmit(): Action[AnyContent] =  secureAction.async { implicit request: Request[AnyContent] =>
+    val formValidationResult: Form[SelectedSeriesData] = selectedSeriesForm.bindFromRequest
+
+    val errorFunction: Form[SelectedSeriesData] => Future[Result] = { formWithErrors: Form[SelectedSeriesData] =>
+      getSeriesDetails(request, BadRequest, formWithErrors)
+    }
+
+    val successFunction: SelectedSeriesData => Future[Result] = { formData: SelectedSeriesData =>
+      val userId = keycloakConfiguration.verifyToken(request.token.getValue)
+        .map(t => t.getOtherClaims.get("user_id").asInstanceOf[String])
```
suzannehamilton: Might be worth extracting this logic and the transferring body logic to another module, so that the token parsing is reusable and in one place.

app/viewsapi/FormFunctions.scala
```diff
@@ -13,6 +16,28 @@ object FormFunctions {
...
+      if(elements.hasErrors) {
+        "govuk-form-group--error"
+      } else {
+        ""
+      }
+    }
+
+    def renderErrorMessage(): Html = {
+      if(elements.hasErrors) {
+        Html("<span id=\"error\" class=\"govuk-error-message\"><span class=\"govuk-visually-hidden\">Error:</span>" +
```
suzannehamilton: Is there any way to move this to the templates, so we don't have to hard-code HTML in a Scala object?

suzannehamilton: I'm a bit wary of having HTML in the Scala classes. I know it's a choice between having HTML in the Scala, or logic in the templates, but I think having some logic is OK if it's very simple like an if statement.

app/views/errorMessage.scala.html
```diff
@@ -0,0 +1,5 @@
+@import views.html.helper._
+@(elements: FieldElements)
+@if(elements.hasErrors) {
+    <span id="error" class="govuk-error-message">@elements.errors.mkString(", ")</span>
```
suzannehamilton: Just in case it wasn't an intentional change - I think the previous version of this had "Error: " before the list of errors.

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -98,5 +101,84 @@ class SeriesDetailsControllerSpec extends FrontEndTestHelper {
...
+      wiremockServer.stubFor(post(urlEqualTo("/graphql"))
+        .willReturn(okJson(dataString)))
+
+      val controller = new SeriesDetailsController(getAuthorisedSecurityComponents(), new GraphQLConfiguration(app.configuration), getValidKeycloakConfiguration)
+      val seriesSubmit = controller.seriesSubmit().apply(FakeRequest().withFormUrlEncodedBody(("series", seriesId.toString)).withCSRFToken)
+      playStatus(seriesSubmit) mustBe SEE_OTHER
+      redirectLocation(seriesSubmit) must be(Some(s"/consignment/$consignmentId/transfer-agreement"))
+    }
+
+    "redirect to the error page when a valid form is submitted but there is an error from the api" in {
```
suzannehamilton: Does this test still pass? Does this mean that we're still redirecting to error pages?

(Note to self: this is where it would be really useful for us to fix the GitHub build reporting.)

MancunianSam: No they don't, there were a few errors which I've fixed.

app/views/errorMessage.scala.html
MancunianSam: I've put it back

test/controllers/SeriesDetailsControllerSpec.scala
suzannehamilton: Nitpicky, but the test name still says "redirect".

MancunianSam: I've changed this to not say redirect.

PR #39, opened 2020-02-27
Upgrade to scala 2 13

MancunianSam:


app/views/seriesDetails.scala.html
```diff
@@ -20,8 +20,9 @@ <h1 class="govuk-heading-xl">@Messages("seriesDetails.title")</h1>
             @inputDropdown(
                 seriesDetailsForm("series"),
                 allSeriesData,
-                '_name -> "Series Details",
-                '_label -> Messages("seriesDetails.chooseSeries"),
+                Symbol("_error") -> seriesDetailsForm.errors("error").headOption.getOrElse(None),
```
suzannehamilton: Is this error being used yet? I was trying to work out what type this needs to be, but I can't see any references to it in inputDropdown.scala.html.

build.sbt
```diff
@@ -1,24 +1,24 @@
...
-scalaVersion := "2.12.8"
+scalaVersion := "2.13.0"
 
 libraryDependencies += guice
 libraryDependencies += "org.scalatestplus.play" %% "scalatestplus-play" % "4.0.3" % Test
 
-val playPac4jVersion = "8.0.1"
-val pac4jVersion = "3.7.0"
+val playPac4jVersion = "9.0.0-RC3"
+val pac4jVersion = "4.0.0-RC3"
```
suzannehamilton: It's probably OK at this early stage of the project, but in general it would be good to use stable versions rather than release candidates. Unless RCs in Scala are generally quite stable?

conf/application.conf
```diff
@@ -9,6 +9,9 @@ play.cache.redis.timeout=20s
 
 consignmentapi.url="http://localhost:8080/graphql"
 
+akka.actor.allow-java-serialization="on"
+akka.actor.warn-about-java-serializer-usage="off"
```
suzannehamilton: Where are we using Java serialization?

build.sbt
MancunianSam: I don't know, I generally just go for the most recent version and I've only had a few problems with it so far.

conf/application.conf
MancunianSam: The library which connects the play cache to redis needs it.

app/views/seriesDetails.scala.html
MancunianSam: It is being used but as you suggested I've changed this to `'_error -> seriesDetailsForm.errors("error").headOption,`

conf/application.conf
MancunianSam: I've added a comment to the first application.conf file

build.sbt
suzannehamilton: Yeah, in this case we'll be using the dependencies in development for long enough to be confident that they're stable, but I wouldn't upgrade a production system to an RC unless there was a really strong reason.

build.sbt
```diff
@@ -1,24 +1,24 @@
 name := """tdr-transfer-frontend"""
 organization := "tna"
+maintainer := "tdr@nationalarchives.gov.uk"
```
suzannehamilton: Is this compulsory? If you haven't already, can you talk to David about what email address we should use?

MancunianSam: It's not mandatory but it was the last warning the build was giving. I've asked David and he's given me the proper email for this.

suzannehamilton: +1 for fixing compiler warnings. Thanks!

PR #40, opened 2020-03-03
Transfer agreement form

TomJKing:


app/controllers/TransferAgreementController.scala
```diff
@@ -3,12 +3,46 @@ package controllers
...
+      "publicRecord" -> boolean
+        .verifying("Must answer yes", b => b),
+      "crownCopyright" -> boolean
+        .verifying("Must answer yes", b => b),
+      "english" -> boolean
+        .verifying("Must answer yes", b => b),
+      "digital" -> boolean
+        .verifying("Must answer yes", b => b),
+      "droAppraisalselection" -> boolean
+        .verifying("Must answer yes", b => b),
```
MancunianSam: You need another capital in `droAppraisalselection` There's a few places in the file where this is so I won't comment on them all

test/viewsapi/FormFunctionSpec.scala
```diff
@@ -60,21 +60,37 @@ class FormFunctionSpec extends FrontEndTestHelper {
...
+    }
+
+    "not render disabled when no _disabledOption" in {
+      val args: Map[Symbol, Any] = Map(Symbol("_madeUpOption") -> "true")
+
+      val result = FormFunctions.inputRenderOptions(args).disabledInput("true")
+      result shouldBe ()
+    }
+
   "setErrorClass function" should {
```
MancunianSam: There's a new `selectedInput` method in the form functions but there's no test for these in here.

app/controllers/TransferAgreementController.scala
TomJKing: fixed

test/viewsapi/FormFunctionSpec.scala
TomJKing: fixed

Issue #34, opened 2020-02-24
(S6) 4.01 Task: **Handle transfer agreement form submission**


TomJKing:
Call API, redirect to placeholder page (for now), create error handling. Might need to link new wireframes of error handling

PR #41, opened 2020-03-06
Submit transfer agreement

techncl:
The Transfer Agreement page already existed. This PR handles the form submit and improve the error messages.

### DIFF
--- a/app/views/inputSingleCheckbox.scala.html
+++ b/app/views/inputSingleCheckbox.scala.html
@@ -9,7 +9,7 @@ args: (Symbol, Any)*
 @value = {@elements.args.get(Symbol("_value"))}
 @label = {@elements.args.get(Symbol("_label"))}
 
-<div class="govuk-checkboxes">
+<div class="govuk-checkboxes @elements.setErrorClass">
     <div class='govuk-checkboxes__item @if(elements.args.contains(Symbol("_smallCheckbox"))){govuk-checkboxes--small}'>
         <input
                 @elements.args.requiredInput()
@@ -23,4 +23,5 @@ args: (Symbol, Any)*
             @label
         </label>
     </div>
+    @errorMessage(elements)
 </div>
--- a/app/views/inputSingleCheckbox.scala.html
+++ b/app/views/inputSingleCheckbox.scala.html
@@ -9,19 +9,21 @@ args: (Symbol, Any)*
 @value = {@elements.args.get(Symbol("_value"))}
 @label = {@elements.args.get(Symbol("_label"))}
 
-<div class="govuk-checkboxes @elements.setErrorClass">
-    <div class='govuk-checkboxes__item @if(elements.args.contains(Symbol("_smallCheckbox"))){govuk-checkboxes--small}'>
-        <input
-                @elements.args.requiredInput()
-                class="govuk-checkboxes__input"
-                id="@elements.id"
-                name="@elements.id"
-                type="checkbox"
-                value="@value"
-        />
-        <label class="govuk-label govuk-checkboxes__label" for="@elements.id">
-            @label
-        </label>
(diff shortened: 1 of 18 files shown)


suzannehamilton:
Resolves #34

suzannehamilton:
We spotted a couple of frontend bugs when I showed this to. I'm fixing them now.

app/views/transferAgreement.scala.html
```diff
@@ -10,61 +10,64 @@
...
-                Symbol("_requiredOption") -> true,
-                Symbol("_disabledOption") -> "false"
-                )
-            </div>
+            @inputRadioGroup(
+            transferAgreementForm("publicRecord"),
+            agreementOptions,
+            Symbol("_name") -> Messages("transferAgreement.publicRecord"),
+            Symbol("_requiredOption") -> true,
+            Symbol("_checkedOption") -> transferAgreementForm.data.getOrElse("publicRecord", ""),
```
suzannehamilton: Out of curiosity, what are the issues with the Play templates? And how will we remember to fix them?

app/modules/SecurityModule.scala
```diff
@@ -56,6 +59,28 @@ class SecurityModule extends AbstractModule {
...
+If you don't provide an authoriser, it gives you the csrf one by default
+so the trick is to give it a dummy authoriser to stop it checking the
+csrf one. Which is what this does.
+
+This has no effect on whether logged out users can see or not see
+certain pages. This is authorisation which we're not doing in the front
+end, although if we do decide to do front end authorisation, we can
+expand the CustomAuthoriser class to something more useful.
+*/
+class CustomAuthoriser extends Authorizer[CommonProfile] {
```
suzannehamilton: This might be a moot point if we add actual authorization in the Pac4J layer, but I think you can delete this authoriser if you change the secureAction in OidcSecurity from this:

```
Secure("OidcClient", authorizers = "custom")
```

to this:

```
Secure("OidcClient", authorizers = DefaultAuthorizers.NONE)
```

app/views/transferAgreement.scala.html
Lholbourn: Within the checkboxes partial specifically, we couldn't find a way for it to allow us to use the checked option as is, so we had to convert it to a string. It doesn't feel like that's the best way of doing it, but we couldn't find another way to make it work properly - we found this when doing the bug fix to keep the checkboxes checked even when the form is submitted but not complete.

app/views/inputSingleCheckbox.scala.html
```diff
@@ -8,19 +8,23 @@
 @elements = @{new FieldElements(field.id, field, null, args.toMap, messages) }
 @value = {@elements.args.get(Symbol("_value"))}
 @label = {@elements.args.get(Symbol("_label"))}
+@isSelected = {@elements.args.selectedInput(value.toString)}
```
Lholbourn: @suzannehamilton this is the area which doesn't feel like its the best way to get access to checked/unchecked in when rendering the checkboxes.

Lholbourn:
This PR branch has issues with the play templates and rendering our partial radio groups & checkboxes. They work and render, but the back-end code is confusing and a little complicated. We will be returning to fix it, but having the branch un-merged is stalling work on other aspects of the project so we will be merging it and fixing it in an ongoing basis so other work can continue.

app/modules/SecurityModule.scala
Lholbourn: Suzanne will be returning to this issue outwith this branch as it is a separate problem.

suzannehamilton: Fixed here: https://github.com/nationalarchives/tdr-transfer-frontend/pull/52

PR #46, opened 2020-03-13
Records page

### LINKED ISSUE #45: (S7) 0.05 Task: **Records placeholder page**

Lholbourn:
This pr creates the records placeholder page upon completion of successful upload. Right now it is only accessible through the link. 
I was thinking of adding a button to the upload placeholder page but have left that for now as I wasn't 100% sure no one was working on that page atm.

closes #45

### DIFF
--- a/test/controllers/RecordsControllerSpec.scala
+++ b/test/controllers/RecordsControllerSpec.scala
@@ -1,21 +1,34 @@
 package controllers
 
+import configuration.GraphQLConfiguration
 import org.scalatest.Matchers._
 import play.api.test.FakeRequest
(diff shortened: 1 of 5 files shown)


Lholbourn:
> I would include a unit test for the controller as for the other pages.

Awesome, I'll do that and then push the new changes

test/controllers/RecordsControllerSpec.scala
```diff
@@ -0,0 +1,22 @@
+package controllers
```
TomJKing: Should add a test to make sure an unauthorised user is redirected to the login page as well. The SeriesDetailsControllerSpec test has an example of this.

PR #48, opened 2020-03-16
Initial js setup

MancunianSam:
When you run the application using `sbt run`, it automatically runs `webpack --watch` which means that any `.ts` file changes will automatically reload in the browser without having to run a parallel command. 
When you run `sbt dist` it does a webpack production build and outputs it to the public folder.
`sbt test` will now run the jest tests as well.
There is linting using es-lint and code formatting using prettier. es-lint runs on a pre-commit hook which will fail if there are any linting errors. We can turn it off if its annoying but I've seen it work well in the past.


npm/css-src/main.css
```diff
@@ -0,0 +1,4337 @@
+.govuk-link {
```
TomJKing: Not sure we should be merging the main.css file as it is generated from the sass files.

MancunianSam: Good point, I've deleted it.

MancunianSam: I've removed main.js while I was at it as that's the same problem.

PR #51, opened 2020-03-17
Remove npm commands from Jenkins build

MancunianSam:


Jenkinsfile-testing
```diff
@@ -39,8 +39,6 @@ pipeline {
           }
           steps {
             checkout scm
-            sh "npm ci"
```
suzannehamilton: Don't we need to do these tasks before deployment, so that the static code is included in the Docker image?

MancunianSam: They've been moved so they run as part of sbt dist so we don't need specific npm commands any more.

suzannehamilton: I see, thanks!

Could you update NpmCommands to run `npm ci` instead of `npm install`? It makes CI builds more consistent: http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

MancunianSam: Yep, makes sense, I've changed it.

PR #54, opened 2020-03-18
(S7) 4.02a Task: Create HTML form with folder selection on page

rsulxo:
Ticket: https://github.com/orgs/nationalarchives/projects/20#card-34420173

Notes:
- Added components on the placeholder page for upload selection
- Replicated the prototypes
- No functionality added yet, only back button works


app/views/upload.scala.html
```diff
@@ -1,6 +1,47 @@
...
+                  <span class="govuk-details__summary-text">
+                    Which file extensions are not permitted?
+                  </span>
+                </summary>
+                <div class="govuk-details__text">
+                  Cannot upload the following extensions: .exe, thumbs.db and .zip
+                </div>
+              </details>
+
+            <!-- <div class="govuk-warning-text">
```
suzannehamilton: Is this old code? Can it be deleted?

css-src/main.css
```diff
@@ -0,0 +1,4337 @@
+.govuk-link {
```
suzannehamilton: Has this file been copied from the npm directory? I think all the CSS and JS code should be in that directory now. @MancunianSam is that right?

package-lock.json
```diff
@@ -0,0 +1,2795 @@
+{
```
suzannehamilton: Similar to the new CSS file - does this need to be here, or has it been copied by accident?

app/views/upload.scala.html
```diff
@@ -1,6 +1,47 @@
 @()(implicit messages: Messages)
 
 @main(Messages("upload.header")) {
-    <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
-    <p>Placeholder page for upload</p>
+  <div class="govuk-grid-row">
```
suzannehamilton: Minor thing, but this file has a mix of 2 and 4 space indents. It should be 4 spaces everywhere.

@MancunianSam - I think you set up the linter. Do you know if we can apply it to Play templates so we don't have to worry about indentation ourselves? 🙂 

You can also configure IntelliJ to make the spacing consistent in Preferences->Editor->Code Style. I'm not sure whether it considers these files to be HTML, but you could try updating those settings.

app/views/upload.scala.html
```diff
@@ -1,6 +1,47 @@
...
+    <div class="govuk-grid-column-two-thirds">
+        <span>
+            <a href="/dashboard" class="govuk-back-link">Back</a>
+            <span class="govuk-caption-xl">Step 3 of 3</span>
```
suzannehamilton: We've got some code that lets us autogenerate these "Step X of Y" messages without having to update them everywhere when the total changes. See https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/views/seriesDetails.scala.html#L13 for example.

app/views/upload.scala.html
```diff
@@ -1,6 +1,47 @@
...
+            <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
+        </span>
+        <div class="govuk-form-group">
+            <label class="govuk-label" for="file-upload-1">
+              Select a folder
+            </label>
+            <input class="govuk-file-upload" id="file-upload-1" name="file-upload-1" type="file">
```
suzannehamilton: Is the `-1` suffix necessary? You could call it "file-selection" or something, if you want to differentiate it from the rest of the page, since it's all about file upload?

suzannehamilton: Is this limited to just directories, rather than individual files? I forget exactly how we did this in the prototype - it might be the `webkitdirectory` attribute here: https://github.com/nationalarchives/tdr-prototype-mvc/blob/29332ab8b1d0fdbbcc3ca186ec88cfdc0545e2d2/app/views/upload.scala.html#L39

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

app/views/upload.scala.html
```diff
@@ -1,6 +1,47 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <span>
+            <a href="/dashboard" class="govuk-back-link">Back</a>
+            <span class="govuk-caption-xl">Step 3 of 3</span>
+            <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
+        </span>
+        <div class="govuk-form-group">
+            <label class="govuk-label" for="file-upload-1">
+              Select a folder
```
suzannehamilton: One more thing, sorry: we're adding user-facing text to the [messages file](https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/conf/messages) so that it's easier to review and update it all in one place. @techncl should be able to help with this one because he's working on it at the moment. 🙂

app/views/upload.scala.html
MancunianSam: The linter is typescript only I'm afraid. There may be one for the templates, I haven't looked.

css-src/main.css
MancunianSam: Yeah, the css-src directory is inside the npm directory now. The changes are in master so if you merge in the latest from master it should be fine.

app/views/upload.scala.html
suzannehamilton: Sorry, I meant Scalastyle rather than ESLint. (Is it fair to call Scalastyle a linter?)

package-lock.json
rsulxo: Removed file as it is not in line with what's on master, not sure why it was generated.

app/views/upload.scala.html
rsulxo: Now using `progressIndicator` to generate dynamic messages from `Messages`. Can you take a look and see if I've done it inline with the other pages?

conf/messages
```diff
@@ -20,12 +20,16 @@ transferAgreement.english=The records are all in English
...
+upload.summary=Cannot upload the following file extensions: .zip, thumbs.db, .exe
 records.header=Records
 records.title=Records
 
 # Progress indicator step numbers
 seriesDetails.progress=1
 transferAgreement.progress=2
-progress.totalSteps=2
\ No newline at end of file
+progress.uploadStep=3
```
suzannehamilton: Just a rewording change: I think this should be `upload.progress` to be consistent with the other steps.

conf/messages
```diff
@@ -20,12 +20,16 @@ transferAgreement.english=The records are all in English
 transferAgreement.digital=The records are all Digital
 transferAgreement.droAppraisalSelection=I confirm that the DRO has signed off on the appraisal and selection decision.
 transferAgreement.droSensitivity=I confirm that the DRO has signed off on the sensitivity review and that all records are OPEN i.e. no FOI exemptions apply to these records.
-upload.header=Upload
-upload.title=Upload
+upload.header=Upload Records
+upload.title=Upload Records
+upload.message=Select a folder
+upload.details=Which file extensions are not permitted?
+upload.summary=Cannot upload the following file extensions: .zip, thumbs.db, .exe
```
suzannehamilton: I might give the last three slightly more descriptive names so it's more obvious what they do when you read the template code. Maybe:

- upload.selectFolder
- upload.fileExtensionTitle
- upload.fileExtensionSummary

rsulxo: I've made the changes that you suggested, not sure why I didn't keep it consistent/more clear before  🤦‍♀

app/views/upload.scala.html
```diff
@@ -1,6 +1,34 @@
...
+    @()(implicit messages: Messages)
 
 @main(Messages("upload.header")) {
-    <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
-    <p>Placeholder page for upload</p>
-}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+        @progressIndicator(Messages("progress.uploadStep"))
```
suzannehamilton: Does this need to be updated to match the new message name?

app/views/upload.scala.html
```diff
@@ -1,6 +1,34 @@
...
+                        @Messages("upload.fileExtensionTitle")
+                    </span>
+                </summary>
+                <div class="govuk-details__text">
+                    @Messages("upload.fileExtensionSummary")
+                </div>
+            </details>
+
+            <button class="govuk-button" type="submit" data-module="govuk-button" role="button">
+                Continue
```
suzannehamilton: Just spotted one more thing that should be in the messages file.

app/views/upload.scala.html
rsulxo: Apologies, unaware of how many minor mistakes I'm making today 🤦‍♀️

app/views/upload.scala.html
rsulxo: I didn't add this to the messages file because on previous pages it is hard coded, so I wasn't sure if I should follow the same pattern or not.

suzannehamilton: Nigel's in the process of moving the other ones, e.g. https://github.com/nationalarchives/tdr-transfer-frontend/blob/adc6d8bc15fa98d13bb0e5743865ac7834b6b1fb/conf/messages#L20

So it would be great if you could move this one too, since it's new.

rsulxo: No worries, that's done too now 😄

conf/messages
```diff
@@ -20,12 +20,17 @@ transferAgreement.english=The records are all in English
...
 transferAgreement.droAppraisalSelection=I confirm that the DRO has signed off on the appraisal and selection decision.
 transferAgreement.droSensitivity=I confirm that the DRO has signed off on the sensitivity review and that all records are OPEN i.e. no FOI exemptions apply to these records.
-upload.header=Upload
-upload.title=Upload
+upload.header=Upload Records
+upload.title=Upload Records
+upload.selectFolder=Select a folder
+upload.fileExtensionTitle=Which file extensions are not permitted?
+upload.fileExtensionSummary=Cannot upload the following file extensions: .zip, thumbs.db, .exe
+upload.continueLink=continue
```
suzannehamilton: Should Continue have an upper case C? Sorry to keep nitpicking!

rsulxo: Not at all! This is extremely helpful because you can point out mistakes I would overlook!

suzannehamilton:
Feel free to merge! 🚀

PR #55, opened 2020-03-18
No upload without agreement

MancunianSam:


app/auth/ValidatedActions.scala
```diff
@@ -0,0 +1,39 @@
...
+
+abstract class ValidatedActions() extends TokenSecurity {
+  implicit val ec: ExecutionContext
+  val graphqlConfiguration: GraphQLConfiguration
+
+  private val getTransferAgreementClient = graphqlConfiguration.getClient[gta.Data, gta.Variables]()
+
+  private def agreed(field: Option[Boolean]) = field.isDefined && field.get
+
+  private def isTransferAgreementComplete(ta: gta.GetTransferAgreement): Boolean = {
```
suzannehamilton: Ah, I think I misunderstood how this was going to work from the API PR. I thought we were just fetching the transfer agreement to see if it existed, rather than expecting the frontend to make a judgement about whether the transfer agreement was complete.

I think this logic should be in the API, because the API's responsible for the integrity of the data and should determine whether or not a transfer agreement is "complete" or not. Other clients might need to work this out, e.g. an admin site or reporting tool.

And there will be similar decisions in future, which will be much easier to handle if they all happen in the consignment API code. For example, the export client will need to know whether the consignment is ready for export, and it shouldn't have to fetch all the consignment data to determine that.

Sorry, I know I'm asking you to change a lot of work that you've already done. I really think it will make things simpler to manage in future!

MancunianSam: I agree actually and also, the definition of what "complete" means may change if we ever get optional things in the agreement. 
I've moved the  logic to the API and added an `isComplete` parameter to the response and the front end is checking this.

suzannehamilton: That sounds good!

Just one thing on the [query](https://github.com/nationalarchives/tdr-generated-graphql/pull/4) - should we have a separate query for checking whether the transfer agreement is complete? One of the biggest advantages of GraphQL is that you only query what you need, but that query forces the frontend to get lots of info that it _doesn't_ need. It's not such a big deal for that particular query because it's just a list of boolean fields, but it would be good to set a precedent of only querying what you need for when we have more expensive nested queries.

MancunianSam: That's a very good point. I've put a new query in the generated graphql project https://github.com/nationalarchives/tdr-generated-graphql/pull/6 and updated this PR to use it.

app/auth/ValidatedActions.scala
```diff
@@ -0,0 +1,32 @@
+package auth
```
suzannehamilton: Sorry, I should have spotted this before: I think this validation is conceptually different to the authorisation, even if the steps are similar (i.e. check something exists in the API, throw an error if not).

This is more about the current state of the transaction. I don't quite know how we should structure the code to handle that at the moment - whether it should wrap secureAction, or if it should be at a different level of abstraction. I think that will get clearer as we build up the steps of the upload process. But it would be good to at least put it into a different module to auth, so that we don't confuse state errors with auth errors.

app/auth/ValidatedActions.scala
```diff
@@ -0,0 +1,32 @@
...
+abstract class ValidatedActions() extends TokenSecurity {
+  implicit val ec: ExecutionContext
+  val graphqlConfiguration: GraphQLConfiguration
+
+  private val getTransferAgreementClient = graphqlConfiguration.getClient[itac.Data, itac.Variables]()
+
+  def transferAgreementExistsAction(consignmentId: Long)(f: Request[AnyContent] => Result): Action[AnyContent]
+  = secureAction.async { implicit request: Request[AnyContent] =>
+    val variables = itac.Variables(consignmentId)
+    getTransferAgreementClient.getResult(request.token.bearerAccessToken, itac.document, Some(variables)).map(data => {
```
suzannehamilton: Should this client still be called `getTransferAgreement` now that the query name has changed?

test/controllers/UploadControllerSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    // This is unlikely but it's possible that they've bypassed the checks and partially agreed to things
+    "redirect to the transfer agreement page if the transfer agreement for that consignment has been partially agreed to" in {
+      implicit val ec: ExecutionContext = ExecutionContext.global
+      val consignmentId = 1
+      val controller = new UploadController(getAuthorisedSecurityComponents,
+        new GraphQLConfiguration(app.configuration), getValidKeycloakConfiguration)
+
+      stubGetTransferAgreementResponse(Some(new gta.GetTransferAgreement(
```
suzannehamilton: Is this still stubbing the right response? It looks like the [new query](https://github.com/nationalarchives/tdr-generated-graphql/pull/6/files) just returns one boolean.

app/auth/ValidatedActions.scala
MancunianSam: Yep, that makes sense. As you say, it's pretty tightly coupled into the secureAction at the moment so any major changes will take time but I've moved it into a new validation package.

app/auth/ValidatedActions.scala
MancunianSam: No, it shouldn't. I've changed it.

test/controllers/UploadControllerSpec.scala
MancunianSam: Technically it is because it just pulls the bit from the query it needs :smile: but no, it should be using the updated query. I've changed it.

PR #57, opened 2020-03-19
Setup keycloak in js

MancunianSam:


npm/test/auth.test.ts
```diff
@@ -0,0 +1,46 @@
...
+class MockKeycloakError {
+  init() {
+    return new Promise((_, reject) => reject("There has been an error"))
+  }
+}
+
+beforeEach(() => jest.resetModules())
+
+test("Returns an error if the user is not logged in", async () => {
+  ;(Keycloak as jest.Mock).mockImplementation(() => {
```
suzannehamilton: Typescript novice question: what does the semicolon at the start of the line do?

npm/test/auth.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    return new MockKeycloakUnauthenticated()
+  })
+  await expect(getToken()).rejects.toEqual("User is not authenticated")
+  jest.clearAllMocks()
```
suzannehamilton: Worth doing this in the before/after each step?

conf/routes
```diff
@@ -12,6 +12,7 @@ GET     /consignment/:consignmentId/transfer-agreement     controllers.TransferA
 POST    /consignment/:consignmentId/transfer-agreement     controllers.TransferAgreementController.transferAgreementSubmit(consignmentId: Long)
 GET     /consignment/:consignmentId/upload                 controllers.UploadController.uploadPage(consignmentId: Long)
 GET     /consignment/:consignmentId/records                controllers.RecordsController.recordsPage(consignmentId: Long)
+GET     /keycloak.json                              controllers.KeycloakConfigurationController.keycloak
```
suzannehamilton: Identation.

npm/src/auth/index.ts
```diff
@@ -0,0 +1,15 @@
+import Keycloak from "keycloak-js"
+
+export const getToken: () => Promise<string | undefined> = async () => {
```
suzannehamilton: How does this code interact with the server-side endpoint at /keycloak.json? Does it look for server settings at that URL by default?

If so, it would probably by worth a comment in the controller explaining that /keycloak.json is a standard path that the JS depends on.

app/controllers/KeycloakConfigurationController.scala
```diff
@@ -0,0 +1,22 @@
...
+
+class KeycloakConfigurationController @Inject ()(val controllerComponents: SecurityComponents,
+                                                 configuration: Configuration
+                                                ) extends OidcSecurity {
+
+  def keycloak: Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
+    val authUrl = configuration.get[String]("auth.url")
+    Ok(Json.obj(
+      "auth-server-url" -> s"$authUrl/auth",
+      "resource" -> "tdr-fe",
```
suzannehamilton: Is this the client ID? I'm just trying to work out how this all fits together. 🙂 

Presumably you have to set up a new non-confidential client in Keycloak? Could you add some instructions to the readme about how to do that?

Also, do we have a card for automating the local Keycloak setup? Could you add one to the backlog if not?

npm/test/auth.test.ts
MancunianSam: This is redundant, the `resetModules` on the beforeEach function does what's needed here. I've removed it.

npm/test/auth.test.ts
MancunianSam: This is a prettier thing rather than specifically typescript. This is the only rationale I could find for it https://prettier.io/docs/en/rationale.html#semicolons which is a bit vague.

conf/routes
MancunianSam: Fixed, thanks.

npm/src/auth/index.ts
MancunianSam: Yep, it's the default location for the settings. I've added a comment to the controller.

app/controllers/KeycloakConfigurationController.scala
MancunianSam: It is the client ID. I've added a line to the README about setting up another client and I've added a card to the backlog to add the exported keycloak config to our private configuration repository. For the time being, local setup will include donwloading and importing that file although when we have time, we can probably automate it into the docker image fully.

npm/test/auth.test.ts
suzannehamilton: Yeah, that sounds like the reason. It still looks ugly to me, but it makes sense. Ah, JavaScript.

README.md
```diff
@@ -72,6 +72,10 @@ updating the frontend.
 
 * Create a new [user](https://www.keycloak.org/docs/latest/getting_started/index.html#_create-new-user) in the tdr realm.
 
+* Set up another [client](https://www.keycloak.org/docs/latest/server_admin/#oidc-clients) called tdr-fe.
```
suzannehamilton: Can you mention what type of client it is? It won't be a confidential client like the server-side one.

npm/test/auth.test.ts
MancunianSam: It's a bit weird, I agree. One of the main reasons for using something like prettier is that it takes personal preference out of the formatting. I don't think there's any configuration that can change this odd semcolon thing but if anyone wants to change any of the prettier config to something else, longer lines or whatever, then that's fine. So long as its consistent :smile:

README.md
MancunianSam: Ok, I've added a line in explaining it's a public client.

PR #58, opened 2020-03-20
Remove user ID from GraphQL params

suzannehamilton:
The API now reads the user ID from the auth token, so this parameter can be removed.

Also update the README on how to set up Keycloak.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ updating the frontend.
 * Start the auth server
 
     `docker run -d  --name keycloak -p 8081:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin nationalarchives/tdr-auth-server:intg`
-* Go to `http://localhost:8080/auth/admin` and log in with username admin and password admin.
+* Go to `http://localhost:8081/auth/admin` and log in with username admin and password admin.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -45,32 +45,58 @@ When you log into the site, you will need to log in as a user from the Integrati
...
+  - Add an attribute called `body` with value matching one of the transferring bodies in the database, e.g.
+    `MOCK1 Department`
+  - Click the Credentials tab
+  - Set a non-temporary password for the user
+- Set AUTH_SECRET as an environment variable in IntelliJ and/or the command line (depending on how you plan to run the
+  frontend project) with the secret as its value:
+  ```
+  AUTH_SECRET=[secret value]
+  ```
+- Create a new [user](https://www.keycloak.org/docs/latest/getting_started/index.html#_create-new-user) in the tdr realm.
```
TomJKing: Is this line still needed as you give instructions to create a user above?

suzannehamilton:
@TomJKing I've tested this against the current version of the API, and it all works because the userId [is now an optional parameter](https://github.com/nationalarchives/tdr-consignment-api/pull/28).

README.md
suzannehamilton: Good point! I've removed this line.

PR #59, opened 2020-03-20
Transfer Agreement - create 'not found' page and functionality

Lholbourn:
This branch creates functionality so that we can check if the consignment id that a user queries (within the URL - /consignment/**53156413**/transfer-agreement). If the consignment id does not exist, the user will be shown a 404 page. 

Changes were also made the to the generated GraphQL project to facilitate a 'getConsignment' query within the transferAgreementController. Based on the results of the query the user is shown the correct page. I included the ability to look at the information within a consignment too as I think this will be useful to us in future (e.g. checking whether user has departmental permissions to see that specific consignment or continuing a terminated consignment etc). I have also included tests for the added functions.

The code within the TA controller is quite complex and I did struggle a little as there were many things that were fairly new concepts to me, so it definitely needs some refactoring.

In terms of the FE part of this task, I have created a placeholder page so far (without a dedicated error controller). I think it would be a good idea for @rsulxo and @sarrahdi to weight in as I don't know if we've fully decided what we want for error pages etc. I did find a small bit of [GDS guidance](https://design-system.service.gov.uk/patterns/page-not-found-pages/) but it is classed as 'experimental' so not sure how well it has been tested.


app/controllers/TransferAgreementController.scala
```diff
@@ -39,8 +43,29 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
+  private val getConsignmentClient = graphqlConfiguration.getClient[getConsignment.Data, getConsignment.Variables]()
+
+  private def getConsignmentDetails(request: Request[AnyContent],
+                                    status: Status,
+                                    consignmentId: Long)(
+                                     implicit requestHeader: RequestHeader): Future[Result] = {
+    val variables: getConsignment.Variables = new GetConsignment.getConsignment.Variables(consignmentId)
+    getConsignmentClient.getResult(request.token.bearerAccessToken, getConsignment.document, Some(variables)).map(data => {
+      if (data.data.isDefined) {
+        val consignmentData: Option[(Long, UUID)] = data.data.get.getConsignment.map(c => (c.seriesid, c.userid))
```
suzannehamilton: All we really care about here is whether the consignment exists or not - we don't actually need the series ID or user ID. It would be better to remove this in case the API changes in the future. For example, we might remove userid from the response because the frontend doesn't normally need to know it.

What's the difference between an API response with consignment data, and one without it? How can we check for that? (That's not a leading question, btw - I can't actually remember what the API returns when there's no data.)

app/controllers/TransferAgreementController.scala
```diff
@@ -39,8 +43,29 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
     )(TransferAgreementData.apply)(TransferAgreementData.unapply)
   )
 
-  def transferAgreement(consignmentId: Long): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
-    Ok(views.html.transferAgreement(consignmentId, transferAgreementForm, options))
+  private val getConsignmentClient = graphqlConfiguration.getClient[getConsignment.Data, getConsignment.Variables]()
+
+  private def getConsignmentDetails(request: Request[AnyContent],
```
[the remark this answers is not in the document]
Lholbourn: That's a good point, I'll start working on extracting it out into a service 👌

app/controllers/TransferAgreementController.scala
```diff
@@ -39,8 +43,29 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
+                                    status: Status,
+                                    consignmentId: Long)(
+                                     implicit requestHeader: RequestHeader): Future[Result] = {
+    val variables: getConsignment.Variables = new GetConsignment.getConsignment.Variables(consignmentId)
+    getConsignmentClient.getResult(request.token.bearerAccessToken, getConsignment.document, Some(variables)).map(data => {
+      if (data.data.isDefined) {
```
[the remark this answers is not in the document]
Lholbourn: Ah, will change this! I did think that I wasn't doing it in the most effective way.

app/controllers/TransferAgreementController.scala
Lholbourn: I can do that, would you like me to change the query in the generated GQL project to reflect this as well? Something like:
```query getConsignment($consignmentId: Long!) {
  getConsignment(consignmentid: $consignmentId) {
    consignmentid
  }
}

File: app/controllers/TransferAgreementController.scala (hunk 2 above)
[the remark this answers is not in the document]
Comment (Lholbourn) [reply app/controllers/TransferAgreementController.scala]: To add to this, I've seen a few projects (not by TNA) that put the services in another folder rather than the same folder as the controllers. Have we got any conventions about this or shall I just use my own judgement? 😊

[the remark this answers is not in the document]
Comment (suzannehamilton) [reply app/controllers/TransferAgreementController.scala]: In Java and Scala, packages are the same as folders, so if you put it into a separate `services` package then it will be in a separate folder. 👍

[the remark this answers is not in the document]
Comment (Lholbourn) [reply app/controllers/TransferAgreementController.scala]: Ah awesome, thanks 😊

**Comment by Lholbourn** [comment]
I'm currently working on changing the testing for the GetConsignmentService to Mockito rather than WireMock. Therefore this PR is a WIP and does not need review ATM. I will hopefully have made the change to Mockito by tomorrow evening.

**Comment by MancunianSam** [comment]
I've moved the tests over to use mockito instead of wiremock. I've also moved the call to the service from the controller to the ValidatedActions class and updated the controller to use this action.

File: test/services/GetConsignmentServiceSpec.scala [hunk 4]
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  "GetConsignmentService GET" should {
+
+    "Return true when given a valid consignment id" in {
+      val graphQLClient = mock[GraphQLClient[gc.Data, gc.Variables]]
+      val graphQLConfig = mock[GraphQLConfiguration]
+      when(graphQLConfig.getClient[gc.Data, gc.Variables]()).thenReturn(graphQLClient)
+
+      val consignmentId = 999
```
Lholbourn: How does this consignment ID work with Int rather than Long?

suzannehamilton: In most cases, the Scala compiler can automatically convert an Int to a Long because it's a safe conversion.

app/validation/ValidatedActions.scala
```diff
@@ -4,16 +4,30 @@ import auth.TokenSecurity
...
 
   private val isTransferAgreementCompleteClient = graphqlConfiguration.getClient[itac.Data, itac.Variables]()
 
+  def consignmentExists(consignmentId: Long)(f: Request[AnyContent] => Result): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    val getConsignmentService = new GetConsignmentService(graphqlConfiguration)
+    val consignmentExists = getConsignmentService.consignmentExists(consignmentId, request.token.bearerAccessToken)
+    consignmentExists.map {
+      case true => f(request)
+      case false => NotFound(views.html.notFoundError("The consignment you are trying to access does not exist"))
+    }.recover {
```
suzannehamilton: Do we need this recover case?

Presumably, if an error happens here, it's not because of anything the user did? So shouldn't we show an internal server error rather than a bad request? So we should just let the exception happen, since that will generate a 500 page. There's a card in the backlog about adding general exception handling to Play, so that exceptions get converted to the styled 500 page.

app/views/notFoundError.scala.html
```diff
@@ -0,0 +1,22 @@
+@*
+* This is a placeholder to allow us to write unit tests which show the redirect to the error page.
```
suzannehamilton: Really nitpicky, but "redirect to the error page" should be "return an error page". Just in case anyone reads this and gets confused.

test/services/GetConsignmentServiceSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+      val graphQLClient = mock[GraphQLClient[gc.Data, gc.Variables]]
+
+      val graphQLConfig = mock[GraphQLConfiguration]
+      when(graphQLConfig.getClient[gc.Data, gc.Variables]()).thenReturn(graphQLClient)
+
+      val captors = mockFailedResponse(graphQLClient)
+
+      val getConsignment = new GetConsignmentService(graphQLConfig).consignmentExists(consignmentId, new BearerAccessToken("someAccessToken"))
+
+      ScalaFutures.whenReady(getConsignment.failed) { results =>
```
suzannehamilton: You can do `getConsignment.failed.futureValue`, and then it uses the same timeouts as `getConsignment.futureValue` in the other tests.

app/services/GetConsignmentService.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  private val getConsignmentClient = graphqlConfiguration.getClient[getConsignment.Data, getConsignment.Variables]()
+
+  def consignmentExists(consignmentId: Long,
+                        token: BearerAccessToken): Future[Boolean] = {
+    val variables: getConsignment.Variables = new GetConsignment.getConsignment.Variables(consignmentId)
+    getConsignmentClient.getResult(token, getConsignment.document, Some(variables)).map(data => {
+      data.data match {
+        case Some(data) => {
+          data.getConsignment match {
```
suzannehamilton, line 23: Just spotted that this `match` statement could be replaced with `isDefined`.

app/validation/ValidatedActions.scala
MancunianSam: It's a fair point, I've taken it out

app/views/notFoundError.scala.html
MancunianSam: You're right. I've changed it.

test/services/GetConsignmentServiceSpec.scala
MancunianSam: Yep, makes sense

app/services/GetConsignmentService.scala
MancunianSam: So we need to check the data.data option is there so I've done `data.data.isDefined && data.data.get.getConsignment.isDefined`

PR #65, opened 2020-03-24
Add graphql client

MancunianSam:


npm/src/index.ts
```diff
@@ -1 +1,28 @@
...
+import { ApolloQueryResult } from "apollo-boost"
+
+// Needed to keep typescript happy
+declare var TDR_API_URL: string
+
+// Pass this class instance to any class or function that needs it
+// This means we don't create the graphql client each time we need a query or mutation
+const graphqlClient = new GraphqlClient(TDR_API_URL)
+
+// Then use the imported types like this
```
TomJKing: Is this just here as an example? If so maybe make an explicit comment that this is the case

MancunianSam: Yeah it is, I put it in the commit comment but it makes sense to put it in here as well. I've added it in.

npm/src/graphql/index.ts
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        variables,
+        context: {
+          headers: {
+            Authorization: `Bearer ${token}`
+          }
+        }
+      }
+    } catch {
+      throw Error("Cannot fetch token")
```
suzannehamilton: I've always been a bit hazy on good practices for handling errors in JavaScript, but can we include the details of the original error rather than throwing them away, so it's easier to debug?

MancunianSam: Good point. I've thrown the original error and updated the tests.

PR #68, opened 2020-03-26
Set expire session with token to be true.

MancunianSam:
When this is set to true, if the token has expired, it will get the
refresh token and call keycloak to get a new access token, and then
store that access token in the cache.

There is a brief description of this functionality at the end of this
page https://www.pac4j.org/docs/clients/openid-connect.html.

The method that refreshes the token is here
https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticator.java#L160

This is called eventually from the profileManager.get(true) line in our
TokenSecurity class.

I've tested this and it will check the expiry on the token and get a new
one silently in the background.

### DIFF
--- a/app/modules/SecurityModule.scala
+++ b/app/modules/SecurityModule.scala
@@ -45,6 +45,7 @@ class SecurityModule extends AbstractModule {
     val secret = configuration.get[String]("auth.secret")
     oidcConfiguration.setSecret(secret)
     oidcConfiguration.setDiscoveryURI(s"$authUrl/auth/realms/tdr/.well-known/openid-configuration")
+    oidcConfiguration.setExpireSessionWithToken(true)
     val oidcClient = new OidcClient[OidcConfiguration](oidcConfiguration)
     oidcClient.setCallbackUrl(callback)
     oidcClient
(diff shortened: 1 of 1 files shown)


suzannehamilton:
It sounds like you've traced this through the code and it does use the refresh token, but the Pac4j docs make it sound like it expires the whole session rather than using the refresh token:

> Additionally, it is possible to establish a behavior according to which, the local session expires when the access token does. In order to do this, just enable `ExpireSessionWithToken` in configuration.

But it definitely does use the refresh token? It doesn't take the user through the login redirect loop again? The user might not see the login page if their Keycloak session is still active, but you should still be able to see the redirects in dev tools if that's what's happening.

suzannehamilton:
Someone asked about this in the Pac4j repo a few months ago, and didn't get a reply: https://github.com/pac4j/play-pac4j/issues/202#issuecomment-549239947

MancunianSam:
Yeah, I saw that issue. I've traced this through the code and I've tested it locally, it's definitely does use the refresh token to update the access token.
I set the token expiry to one minute and added a print statement for the token.
After a minute, I refreshed the page and I got a new token with an updated expiry time.
Also, if I refreshed inside the minute, I didn't hit the breakpoints inside the refresh method I linked above. Outside of the minute, I did.

This does exactly what we need, it's just documentation isn't the pac4j developers strong point.

suzannehamilton:
Cheers. In which case, would you mind adding a code comment explaining that `setExpireSessionWithToken` uses the refresh token to get a new access token? It's a shame that it's so non-obvious. 🙁

MancunianSam:
I've added a comment in. It is a pain that the most effective way of working out how the library works is digging through the source code.

PR #69, opened 2020-03-26
Make the HTML ids for error messages unique

techncl:
Was going back and forth as to whether id should be in Kebab case like other HTML ids or Camel case like the ids on the transferAgreement page; kept it as Camel.

### DIFF
--- a/app/views/errorMessage.scala.html
+++ b/app/views/errorMessage.scala.html
@@ -1,7 +1,9 @@
 @import views.html.helper._
 @(elements: FieldElements)(implicit messages: Messages)
+
 @if(elements.hasErrors) {
-    <span id="error" class="govuk-error-message">
+
+    <span id="@elements.id-error" class="govuk-error-message">
         <span class="govuk-visually-hidden">@Messages("screenReader.error")</span> @elements.errors.mkString(", ")
-        </span>
+    </span>
 }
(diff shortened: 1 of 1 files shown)


app/views/errorMessage.scala.html
```diff
@@ -1,7 +1,9 @@
 @import views.html.helper._
 @(elements: FieldElements)(implicit messages: Messages)
+
 @if(elements.hasErrors) {
-    <span id="error" class="govuk-error-message">
+
+    <span id="@elements.id-error" class="govuk-error-message">
```
suzannehamilton: Where does `@elements.id-error` come from? I don't see it in the [Play docs](https://www.playframework.com/documentation/2.8.0/api/scala/views/html/helper/FieldElements.html).

techncl: Ah, this is just string interpolation; @element.id + the string "-error" so that the id will end up as "publicRecord-error" for example. Was not sure whether it should be "public-record-error" (like on the GDS) , publicRecordError or publicRecord-error (like what was already there) though.

suzannehamilton:
As discussed, let's see if it's possible to delete these IDs entirely so we don't have to worry about them. 👍

suzannehamilton:
Has this PR been superseded by #80? Can this one be closed?

techncl:
Yes, it has and yes it can :)

PR #73, opened 2020-04-03
Switch to postgres

MancunianSam:


build.sbt
```diff
@@ -8,6 +9,13 @@ lazy val root = (project in file(".")).enablePlugins(PlayScala)
 
 watchSources ++= (baseDirectory.value / "npm/src" ** "*").get
 
+resolvers ++= Seq[Resolver](
+  s3resolver.value("Releases resolver", s3("tdr-releases-mgmt")),
+  s3resolver.value("Snapshots resolver", s3("tdr-snapshots-mgmt"))
+)
+
+useCoursier := false
```
suzannehamilton, line 17: Just to check - why is this necessary? I hadn't heard of Coursier before, but it sounds like it's the recommended way to resolve dependencies. Does it not work with S3?

MancunianSam: Coursier is included automatically with higher versions of sbt and the original plugin doesn't support it and doesn't support private maven style repositories. I've replaced this with a resolver plugin which supports both.

suzannehamilton: Ah cool. That one looks more like it's in active development too. 👍

suzannehamilton:
Approved, but I'm guessing you'll want to hold off on merging until the API, etc. are ready?

PR #75, opened 2020-04-07
Update GraphQL client library to version 0.0.12

suzannehamilton:
This brings in the updated error handling and error codes: https://github.com/nationalarchives/tdr-graphql-client/pull/4

### DIFF
--- a/test/controllers/SeriesDetailsControllerSpec.scala
+++ b/test/controllers/SeriesDetailsControllerSpec.scala
@@ -69,7 +69,7 @@ class SeriesDetailsControllerSpec extends FrontEndTestHelper {
 
     "render the error page if the api returns errors" in {
       val client = new GraphQLConfiguration(app.configuration).getClient[gs.Data, gs.Variables]()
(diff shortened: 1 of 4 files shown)


test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -16,6 +16,7 @@ import play.api.test.CSRFTokenHelper._
 import play.api.test.FakeRequest
 import play.api.test.Helpers.{status => playStatus, _}
 import uk.gov.nationalarchives.tdr.GraphQLClient
+import uk.gov.nationalarchives.tdr.error.GraphQlError
 import util.FrontEndTestHelper
```
MancunianSam, line 20: This import isn't being used

suzannehamilton: Fixed, thanks.

suzannehamilton:
@MancunianSam Yeah, in the authorisation branch I've added a [Play ErrorHandler](https://www.playframework.com/documentation/2.8.x/ScalaErrorHandling) which means that we can return a different status code depending on the error, and will eventually make it easy for us to return the branded error pages without having to add them to every controller action.

PR #76, opened 2020-04-08
Handle authorisation errors from the API

suzannehamilton:
This change:

- Checks for authorisation errors in the GraphQL response, and throws an AuthorisationException if it finds any
- Adds a [custom Play ErrorHandler](https://www.playframework.com/documentation/2.8.x/ScalaErrorHandling), which catches AuthorisationException and converts it to a 403 Forbidden response

The custom error handling will make it easier for us to add branded error pages in future, because we can add them to this file rather than to all of the controllers.

I've extracted services from a few of the controllers, so that I wasn't cluttering up the controllers even more. That makes this change quite large, but hopefully it's worth it - take a look at the SeriesDetailsController, for example. I think it's a bit easier to follow [on this branch](https://github.com/nationalarchives/tdr-transfer-frontend/blob/add-authorisation/app/controllers/SeriesDetailsController.scala) than it does on [master](https://github.com/nationalarchives/tdr-transfer-frontend/blob/02c22f7ce9dea83819001367e5437183fd405731/app/controllers/SeriesDetailsController.scala) - let me know what you think!

I've also tried to simplify the tests where I was adding more setup that would have made them more complicated.

### DIFF
--- a/app/controllers/TransferAgreementController.scala
+++ b/app/controllers/TransferAgreementController.scala
@@ -3,7 +3,6 @@ package controllers
 import java.util.UUID
 
 import configuration.{GraphQLConfiguration, KeycloakConfiguration}
-import errors.AuthorisationException
 import graphql.codegen.AddTransferAgreement.AddTransferAgreement
 import graphql.codegen.types.AddTransferAgreementInput
 import javax.inject.{Inject, Singleton}
@@ -12,7 +11,7 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{I18nSupport, Lang, Langs}
 import play.api.mvc.{Action, AnyContent, Request, Result}
-import uk.gov.nationalarchives.tdr.error.NotAuthorisedError
+import services.ApiErrorHandling.sendApiRequest
 import validation.ValidatedActions
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -66,13 +65,8 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
 
       val variables: AddTransferAgreement.Variables = AddTransferAgreement.Variables(addTransferAgreementInput)
 
-      addTransferAgreementClient.getResult(request.token.bearerAccessToken, AddTransferAgreement.document, Some(variables)).map(data => {
-        data.errors match {
-          case Nil => Redirect(routes.UploadController.uploadPage(consignmentId))
-          case List(authError: NotAuthorisedError) => throw new AuthorisationException(authError.message)
-          case errors => throw new RuntimeException(errors.map(e => e.message).mkString)
-        }
-      })
+      sendApiRequest(addTransferAgreementClient, AddTransferAgreement.document, request.token.bearerAccessToken, variables)
+        .map(_ => Redirect(routes.UploadController.uploadPage(consignmentId)))
     }
 
     val formValidationResult: Form[TransferAgreementData] = transferAgreementForm.bindFromRequest
--- a/app/controllers/TransferAgreementController.scala
+++ b/app/controllers/TransferAgreementController.scala
@@ -12,7 +12,6 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{I18nSupport, Lang, Langs}
 import play.api.mvc.{Action, AnyContent, Request, Result}
-import services.ConsignmentService
 import uk.gov.nationalarchives.tdr.error.NotAuthorisedError
 import validation.ValidatedActions
 
@@ -22,7 +21,6 @@ import scala.concurrent.{ExecutionContext, Future}
 class TransferAgreementController @Inject()(val controllerComponents: SecurityComponents,
                                             val graphqlConfiguration: GraphQLConfiguration,
                                             val keycloakConfiguration: KeycloakConfiguration,
-                                            val getConsignmentService: ConsignmentService,
                                             langs: Langs)
                                            (implicit val ec: ExecutionContext) extends ValidatedActions with I18nSupport {
 
--- a/app/controllers/TransferAgreementController.scala
+++ b/app/controllers/TransferAgreementController.scala
@@ -12,7 +12,7 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{I18nSupport, Lang, Langs}
 import play.api.mvc.{Action, AnyContent, Request, Result}
-import services.GetConsignmentService
+import services.ConsignmentService
 import uk.gov.nationalarchives.tdr.error.NotAuthorisedError
 import validation.ValidatedActions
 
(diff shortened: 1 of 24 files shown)


app/controllers/TransferAgreementController.scala
```diff
@@ -21,7 +22,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class TransferAgreementController @Inject()(val controllerComponents: SecurityComponents,
                                             val graphqlConfiguration: GraphQLConfiguration,
                                             val keycloakConfiguration: KeycloakConfiguration,
-                                            val getConsignmentService: GetConsignmentService,
+                                            val getConsignmentService: ConsignmentService,
```
TomJKing, line 25: Is the ConsignmentService needed?

app/controllers/TransferAgreementController.scala
```diff
@@ -12,7 +12,8 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{I18nSupport, Lang, Langs}
 import play.api.mvc.{Action, AnyContent, Request, Result}
-import services.GetConsignmentService
+import services.ConsignmentService
+import uk.gov.nationalarchives.tdr.error.NotAuthorisedError
```
TomJKing, line 16: This import doesn't appear to work when I try this locally?

app/validation/ValidatedActions.scala
```diff
@@ -5,22 +5,19 @@ import java.util.UUID
...
 abstract class ValidatedActions() extends TokenSecurity with I18nSupport {
   implicit val ec: ExecutionContext
   val graphqlConfiguration: GraphQLConfiguration
 
-  private val isTransferAgreementCompleteClient = graphqlConfiguration.getClient[itac.Data, itac.Variables]()
-
   def consignmentExists(consignmentId: UUID)(f: Request[AnyContent] => Result): Action[AnyContent] = secureAction.async {
     implicit request: Request[AnyContent] =>
-    val getConsignmentService = new GetConsignmentService(graphqlConfiguration)
+    val getConsignmentService = new ConsignmentService(graphqlConfiguration)
```
TomJKing, line 20: Maybe rename to val consignmentService to match the renamed class

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -122,39 +135,54 @@ class SeriesDetailsControllerSpec extends FrontEndTestHelper {
...
       val seriesSubmit = controller.seriesSubmit().apply(FakeRequest(POST, "/series").withFormUrlEncodedBody(("series", seriesId.toString)).withCSRFToken)
-      playStatus(seriesSubmit) mustBe BAD_REQUEST
+
+      seriesSubmit.failed.futureValue shouldBe a[GraphQlException]
     }
 
     "display errors when an invalid form is submitted" in {
       val client = new GraphQLConfiguration(app.configuration).getClient[gs.Data, gs.Variables]()
-      val data: client.GraphqlData = client.GraphqlData(Option.empty, List(GraphQLClient.Error("Error", Nil, Nil, None)))
+      val uuid = UUID.fromString("c2efd3e6-6664-4582-8c28-dcf891f60e68")
```
TomJKing, line 155: Might be worth abstracting this value out as its used in a couple of different tests

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -60,42 +65,46 @@ class SeriesDetailsControllerSpec extends FrontEndTestHelper {
...
 
-      playStatus(seriesDetailsPage) mustBe BAD_REQUEST
-      contentType(seriesDetailsPage) mustBe Some("text/html")
-      //This test is based on the placeholder error page, it will need to be changed when we implement a new error page.
-      contentAsString(seriesDetailsPage) must include ("Error")
-      contentAsString(seriesDetailsPage) must include ("govuk-error-message")
-
-      wiremockServer.verify(postRequestedFor(urlEqualTo("/graphql")))
+      val failure = seriesDetailsPage.failed.futureValue
+      failure mustBe an[Exception]
```
MancunianSam, line 93: Can this be more specific? I'm guessing if this is a general API error then this should be a GraphQlException

MancunianSam:
Looks really nice and clean. Just a couple of questions on things

app/errors/ErrorHandler.scala
```diff
@@ -0,0 +1,34 @@
...
+    )
+  }
+
+  override def onServerError(request: RequestHeader, exception: Throwable): Future[Result] = {
+    logger.error(s"Internal server error at path '${request.path}'", exception)
+
+    val response = exception match {
+      case authException: AuthorisationException =>
+        Forbidden(views.html.error(s"Not authorised: ${authException.getMessage}")(request2Messages(request)))
+      case e =>
```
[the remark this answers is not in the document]
suzannehamilton: Yeah, it's because it's a custom exception. I couldn't find a better way to handle the exception that would let us return a 403 status code _and_ log all the error details without duplicating the error handling in every controller that might hit an authorisation error.

Do you know if there's a more idiomatic way of doing this in Play? I'm mostly trying to remember how other frameworks do things and then trying to find the equivalent in Play, so I'm probably missing things. 🙂

app/controllers/TransferAgreementController.scala
suzannehamilton: Good spot. Deleted. 👍

app/controllers/TransferAgreementController.scala
suzannehamilton: Do you have the latest version? It's in version 0.0.12.

app/validation/ValidatedActions.scala
suzannehamilton: Done, thanks.

test/controllers/SeriesDetailsControllerSpec.scala
suzannehamilton: Done, thanks.

Thinking out loud: at some point I'd quite like to tidy these tests up a bit more. The controller and its dependencies are stateless, so we ought to be able to construct them at the top of the test rather than for each test, but it will require a bit of dependency juggling to get it to work.

app/controllers/TransferAgreementController.scala
```diff
@@ -68,10 +69,10 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
       addTransferAgreementClient.getResult(request.token.bearerAccessToken, AddTransferAgreement.document, Some(variables)).map(data => {
-        if(data.data.isDefined) {
-          Redirect(routes.UploadController.uploadPage(consignmentId))
-        } else {
-          InternalServerError(views.html.error(data.errors.map(e => e.message).mkString))
+        data.errors match {
+          case Nil => Redirect(routes.UploadController.uploadPage(consignmentId))
+          case List(authError: NotAuthorisedError) => throw new AuthorisationException(authError.message)
+          case errors => throw new RuntimeException(errors.map(e => e.message).mkString)
         }
```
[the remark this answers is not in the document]
suzannehamilton: Good spot - I forgot to extract this one. Done. 👍

test/controllers/SeriesDetailsControllerSpec.scala
suzannehamilton: You're right. Fixed.

app/errors/ErrorHandler.scala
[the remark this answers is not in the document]
MancunianSam: I don't know of any other way, this is cleaner than duplicating the logic everywhere so I'm happy.

suzannehamilton:
Just rebased to fix a merge conflict in application.intg.conf.

PR #77, opened 2020-04-08
Progress Bar (TDR-16)

rsulxo:
#### Jira card: https://national-archives.atlassian.net/browse/TDR-16?atlOrigin=eyJpIjoiYjQ5MWMxMTNmZDFhNDc3MmJjZWZlYTBkN2IyYTY4YWQiLCJwIjoiaiJ9

[TDR-21 & TDR-19]

*In order for you to be able to run the progress bar and see the metadata extraction it is important to note that you may need to comment out line 75 in `npm/src/clientfileprocessing/index.ts` - the line being: `this.s3Upload.uploadToS3(consignmentId, tdrFiles, callback, stage)`*

This PR is showing that upon selecting a file on the upload page and clicking continue, the file-upload div is hidden and the progress-bar div is visible. This also shows the progress bar extracting the metadata up until 50% as the other half will be completed with s3 uploads.


app/controllers/UploadController.scala
```diff
@@ -19,6 +19,10 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
                                 (implicit val ec: ExecutionContext) extends ValidatedActions with I18nSupport {
 
   def uploadPage(consignmentId: UUID): Action[AnyContent] = transferAgreementExistsAction(consignmentId) { implicit request: Request[AnyContent] =>
-    Ok(views.html.upload())
+    Ok(views.html.upload(consignmentId))
+  }
+
+  def uploadSubmit(consignmentId: UUID): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
```
suzannehamilton, line 25: I think I was wrong about this action on the call! I thought we could remove it because the JavaScript will handle the redirect, but I was misremembering what we did in the prototype. I thought we did a pure-JavaScript redirect, but actually we just submitted the form when the upload had finished: https://github.com/nationalarchives/tdr-prototype-mvc/blob/master/js-src/upload/index.ts#L54

We'll still be using JavaScript to handle the submit click (see [Tom's open PR](https://github.com/nationalarchives/tdr-transfer-frontend/pull/74/files#diff-c03c04aaddc8881dc3dcf93441a5eed9R25)), but we will eventually submit the form.

So ignore what I said about deleting this controller action! It should stay. 🙂

app/views/upload.scala.html
```diff
@@ -33,4 +33,13 @@ <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
         </form>
     </div>
 </div>
+<div id="progress-bar" class="govuk-grid-row" style="visibility: hidden">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">@Messages("fileProgress.header")</h1>
+        <p class="govuk-body">@Messages("fileProgress.title")</p>
+        <div>
+            <progress class="progress-bar" value="" max="50"></progress>
```
TomJKing, line 41: Little bit confusing having an id and class with the same value: progress-bar. I would rename the class something like: progress-display, perhaps?

npm/src/clientfileprocessing/index.ts
```diff
@@ -40,6 +40,30 @@ export class ClientFileProcessing {
             "Percent of metadata extracted: " +
               progressInformation.percentageProcessed
           )
+          const fileUpload: HTMLDivElement | null = document.querySelector(
```
TomJKing, line 43: I would abstract the progress bar function to a named function instead of having it as an anonymous call back function. This will make is more reusable and make testing easier

app/views/checkingRecords.scala.html
```diff
@@ -0,0 +1,6 @@
+@()(implicit messages: Messages)
```
TomJKing, line 1: Is this page being used? I don't think there is a redirect to it once the upload has completed? How does this relate to the changes Loki is making in PR: https://github.com/nationalarchives/tdr-transfer-frontend/pull/84/files. I think there is a bit of overlap. Maybe this should be removed from your PR?

npm/src/clientfileprocessing/index.ts
```diff
@@ -40,6 +40,30 @@ export class ClientFileProcessing {
...
+          )
+          const progressBar: HTMLDivElement | null = document.querySelector(
+            "#progress-bar"
+          )
+          const progressBarElement: HTMLDivElement | null = document.querySelector(
+            ".progress-bar"
+          )
+
+          if (fileUpload && progressBar) {
+            fileUpload.style.display = "none"
```
TomJKing, line 54: Shouldn't be using style in the HTML

app/views/upload.scala.html
```diff
@@ -33,4 +33,13 @@ <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
         </form>
     </div>
 </div>
+<div id="progress-bar" class="govuk-grid-row" style="visibility: hidden">
```
TomJKing, line 36: Don't use "style" in HTML. Use a class and css instead. For example class="govuk-grid-row hide". With CSS for the hide class to keep the element hidden.

suzannehamilton:
When I run this locally, I still see the unstyled version of the progress bar:

Screenshot 2020-05-04 at 08 55 37

How do the custom styles get loaded in? Is there a missing npm library or have I missed a build step? I've run `npm install` and `npm run build:local`.

rsulxo:
Hi Suzanne, have you tried commenting this line out? 

https://github.com/nationalarchives/tdr-transfer-frontend/blob/5d52d73f875c27c0a5a78917a22a34bfe9e8e682/npm/src/clientfileprocessing/index.ts#L74

suzannehamilton:
@rsulxo I've been testing it with a large file, so I have time to see the progress bar for the metadata before it reaches the S3 upload step.

npm/package.json
```diff
@@ -39,17 +39,19 @@
   },
   "homepage": "https://github.com/nationalarchives/tdr-transfer-frontend#readme",
   "dependencies": {
+    "@nationalarchives/file-information": "1.0.4",
+    "@nationalarchives/tdr-generated-graphql": "1.0.12",
+    "@types/mocha": "^7.0.2",
```
TomJKing, line 44: Is there a reason that @types/mocha and ajv dependencies have been added

rsulxo: I added that as my jest types were not being recognised by vscode, but I can remove them as it works now.

app/views/upload.scala.html
```diff
@@ -48,4 +49,13 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
         </form>
     </div>
 </div>
+<div id="progress-bar" class="govuk-grid-row hide">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">@Messages("fileProgress.header")</h1>
+        <p class="govuk-body">@Messages("fileProgress.title")</p>
+        <div>
+            <progress class="progress-display" value="" max="50"></progress>
```
suzannehamilton, line 57: I've been thinking about the classes and IDs and BEM. I think this element should be "progress-bar" rather than "progress-display" because "progress-display" sounds like it might include other elements, such as text that describes the progress.

npm/css-src/sass/overrides/_progress.scss
```diff
@@ -0,0 +1,32 @@
+.progress-display[value] {
+
+    /* Get rid of the default appearance */
+    appearance: none;
+    border: none;
+    /* Dimensions */
+    width: 100%; height: 20px;
+    background-color: rgb(245,245,245);
```
suzannehamilton, line 8: The design system comes with a [colour palette](https://frontend.design-system.service.gov.uk/sass-api-reference/#helpers-colour). I think we should use the palette by default so that the progress bar is consistent with the rest of the site.

If we want to use a colour that's not in the palette, I'd suggest extracting the colour to a sass file for our extended palette, so that all the colours can be referenced in one place.

suzannehamilton: Also, if we are using our own colours, it might be good to add them as hex values rather than RGB, again for consistency with the design system: https://github.com/alphagov/govuk-frontend/blob/v3.6.0/src/govuk/settings/_colours-applied.scss#L135

Unless there's a good reason to use RGB/RGBA instead?

npm/css-src/sass/overrides/_progress.scss
```diff
@@ -0,0 +1,32 @@
...
+    border-radius: 10px;
+    box-shadow: 0 2px 3px rgba(0,0,0,.5) inset;
+    color: rgb(65, 105, 225);
+    position: relative;
+    margin: 0 0 1.5em;
+}
+
+.progress-display[value]::-webkit-progress-bar {
+    background-color: rgb(245,245,245);
+    border-radius: 10px;
```
suzannehamilton, line 18: Could you create variables for things like the border-radius which are repeated in the different blocks? That will mean it only has to change in one place if we need to tweak it later.

npm/css-src/sass/overrides/_progress.scss
```diff
@@ -0,0 +1,32 @@
+.progress-display[value] {
```
suzannehamilton, line 1: This file should probably be in the sass folder rather than the overrides folder. I think /overrides is for where we're overriding basic HTML defaults or the design system, whereas this is totally new code. 🙂

app/views/upload.scala.html
rsulxo: I had initially named it progress-bar but Tom mentioned that it can be confusing having an id and class with the same value, so I changed it to progress-display, could there possibly be a better word besides display that I could use here?

npm/css-src/sass/overrides/_progress.scss
suzannehamilton: Thanks for converting the colours to hex values. Did you take a look at the design system's colour palette? Can we use colours from there or define our own?

rsulxo: That's all done, I've updated our colours to match the gov uk colour palette and so now they are more consistent to the design standards and look more cleaner too.

npm/css-src/sass/main.scss
```diff
@@ -1,6 +1,44 @@
...
+    border: none;
+    /* Dimensions */
+    width: 100%;
+    height: 20px;
+    background-color: govuk-colour("light-grey", $legacy: "grey-1");
+    border-radius: $borderRadius;
+    box-shadow: $boxShadow;
+    color: govuk-colour("blue", $legacy: "govuk-blue");
+    position: $position;
+    margin: 0 0 1.5em;
```
suzannehamilton, line 21: Just noticed - this size should be in px for consistency with the other ones (and with the GOV.UK design system, which [seems to use px](https://github.com/alphagov/govuk-frontend/blob/v3.6.0/src/govuk/settings/_measurements.scss#L16-L16) too).

rsulxo: No worries, that's now been updated!

PR #81, opened 2020-04-17
Add javascript for upload

MancunianSam:


npm/src/auth/index.ts
```diff
@@ -17,3 +21,33 @@ export const getKeycloakInstance: () => Promise<
...
+  keycloak: Keycloak.KeycloakInstance<"native">
+) => Promise<string> = async keycloak => {
+  const token = await refreshOrReturnToken(keycloak)
+  const credentials = new CognitoIdentityCredentials({
+    IdentityPoolId: TDR_IDENTITY_POOL_ID,
+    Logins: {
+      [TDR_IDENTITY_PROVIDER_NAME]: token
+    }
+  })
+  AWS.config.update({ region: "eu-west-2", credentials })
```
TomJKing, line 49: Maybe region should be set as env variable?

npm/src/s3upload/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    tdrFile,
+    callback,
+    progressInfo
+  ) => {
+    const { file, fileId } = tdrFile
+    console.log(`${this.identityId}/${fileId}`)
+    const progress: S3.ManagedUpload = this.s3.upload({
+      Key: `${this.identityId}/${fileId}`,
+      Body: file,
+      Bucket: "tdr-upload-files-intg"
```
TomJKing, line 39: Should the bucket name be hardcoded? Would be different depending on the environment

npm/src/auth/index.ts
MancunianSam: I've set it as a webpack variable but then hard coded it into the package json. I don't think we'll ever need to change it but if we do, I can set it in the Jenkins container.

npm/src/s3upload/index.ts
MancunianSam: Good spot, I've added the stage variable and used it here.

PR #84, opened 2020-04-20
Redirect to records page after successful upload

Lholbourn:
This change makes the file upload asynchronous so that we can wait until
it is completed, then redirect the user to the 'processing records' placeholder page.

**As the upload does not fully work yet, this code is untested and I'm happy to leave it
as a WIP if recommended to.**

Rather than utilising an additional form submission in the browser as the alpha version does.
This change uses the existing uploading JS functions to redirect the user to the right page, as these functions already hold the information about the progress of the upload.

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -5,8 +5,6 @@ import java.util.UUID
 import configuration.{FrontEndInfoConfiguration, GraphQLConfiguration, KeycloakConfiguration}
 import javax.inject.{Inject, Singleton}
 import org.pac4j.play.scala.SecurityComponents
-import play.api.data.Form
-import play.api.data.Forms._
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, Request}
 import validation.ValidatedActions
@@ -24,10 +22,4 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
     Ok(views.html.upload(consignmentId, frontEndInfoConfiguration.frontEndInfo))
   }
 
-  def uploadRedirect(consignmentId: UUID): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
-    Redirect(routes.RecordsController.recordsPage(consignmentId))
-  }
-
 }
-
-case class UploadData(consignmentId: UUID)
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -5,7 +5,8 @@ import java.util.UUID
 import configuration.{FrontEndInfoConfiguration, GraphQLConfiguration, KeycloakConfiguration}
 import javax.inject.{Inject, Singleton}
 import org.pac4j.play.scala.SecurityComponents
-import play.api.Configuration
+import play.api.data.Form
+import play.api.data.Forms._
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, Request}
 import validation.ValidatedActions
@@ -22,4 +23,11 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
   def uploadPage(consignmentId: UUID): Action[AnyContent] = transferAgreementExistsAction(consignmentId) { implicit request: Request[AnyContent] =>
     Ok(views.html.upload(consignmentId, frontEndInfoConfiguration.frontEndInfo))
   }
+
+  def uploadRedirect(consignmentId: UUID): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
+    Redirect(routes.RecordsController.recordsPage(consignmentId))
+  }
+
 }
+
+case class UploadData(consignmentId: UUID)
--- a/app/views/upload.scala.html
+++ b/app/views/upload.scala.html
@@ -25,9 +25,6 @@
                 </ul>
             </div>
         </div>
-        @form(routes.UploadController.uploadRedirect(consignmentId), Symbol("id") -> "upload-data-form") {
-            @CSRF.formField
-        }
         <form id="file-upload-form" data-consignment-id="@consignmentId">
             <div class="govuk-form-group">
                 <label class="govuk-label" for="file-selection">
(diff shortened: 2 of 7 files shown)


npm/src/upload/index.ts
```diff
@@ -48,11 +48,12 @@ export class UploadFiles {
             throw Error("No files selected")
           }
 
-          this.clientFileProcessing.processClientFiles(
+          await this.clientFileProcessing.processClientFiles(
             consignmentId,
             files,
             progress => console.log(progress.percentageProcessed)
           )
+          location.replace(`/${consignmentId}/records`)
```
TomJKing, line 56: Does using the replace function have any consequences for other behaviours? For example the "back" button? https://developer.mozilla.org/en-US/docs/Web/API/Location

Lholbourn: Yeah, I wasn't 100% sure on whether we wanted users to be able to click the back button, so I went for replace which doesn't allow it. I had a feeling being able to go back might add some more complexity to our form logic since we don't want multiple uploads of the same thing. Would assign() - which allows user to press back - be a better choice?

TomJKing: Probably want to check the requirements around what should be happening with this

suzannehamilton: If we do this as a form submissions, like in the prototype, does it preserve the back button behaviour?

If so, I think we should go with that for now. I've got a note to discuss back buttons with Kirsten and Sarra - both the "Back" link on the page, and the back button in the browser.

suzannehamilton: I talked to Kirsten about the Back link and the browser's back button, and we're going to have a discussion about all the possible scenarios to decide what should happen. That discussion might not happen for a while, though.

@Lholbourn @TomJKing What do you think about changing this to a form submission, to try to preserve the back button behaviour for now?

Lholbourn: @suzannehamilton How big a change do you think it would be to make it a form submission?

suzannehamilton: I don't think it's a huge change. You'll need to:

- Add a new controller endpoint which handles form submission - all it needs to do is redirect to the next page
- Update the `<form>` element in the HTML template so that form submission links to the new controller endpoint
- Capture the form submission event, and submit it when the upload is finished. Here's how we did it in the prototype: https://github.com/nationalarchives/tdr-prototype-mvc/blob/167a45dcb820777ae8d24859147fadb01ce194d2/js-src/upload/index.ts#L54

Lholbourn: Ah cool, I'll give it a go today!

app/views/upload.scala.html
```diff
@@ -24,6 +25,9 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
                 </ul>
             </div>
         </div>
+        @form(routes.UploadController.uploadRedirect(consignmentId), Symbol("id") -> "upload-data-form") {
```
TomJKing, line 28: I would suggest moving this form to after the main upload form, to give some indication that this form is used after the main form. Probably should add a comment about why it is being used as well

Have you tried to call the RecordController.recordsPage(...) directly here, instead of calling the UploadController, which then redirects to the RecordsController.recordsPage(...). So you would have something looking like this:

@form(routes.RecordsController.recordsPage(consignmentId)) {@CRSF.formField}

app/controllers/UploadController.scala
```diff
@@ -22,4 +23,11 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
...
     Ok(views.html.upload(consignmentId, frontEndInfoConfiguration.frontEndInfo))
   }
+
+  def uploadRedirect(consignmentId: UUID): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
+    Redirect(routes.RecordsController.recordsPage(consignmentId))
+  }
+
 }
+
+case class UploadData(consignmentId: UUID)
```
TomJKing, line 33: This case class is not needed

conf/routes
```diff
@@ -11,6 +11,7 @@ POST    /series                                            controllers.SeriesDet
 GET     /consignment/:consignmentId/transfer-agreement     controllers.TransferAgreementController.transferAgreement(consignmentId: java.util.UUID)
 POST    /consignment/:consignmentId/transfer-agreement     controllers.TransferAgreementController.transferAgreementSubmit(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/upload                 controllers.UploadController.uploadPage(consignmentId: java.util.UUID)
+POST    /consignment/:consignmentId/upload                controllers.UploadController.uploadRedirect(consignmentId: java.util.UUID)
```
TomJKing, line 14: Regarding the above comments, this route won't be needed.

npm/src/upload/index.ts
```diff
@@ -51,12 +55,15 @@ export class UploadFiles {
...
           }
 
-          this.clientFileProcessing.processClientFiles(
+          await this.clientFileProcessing.processClientFiles(
             consignmentId,
             files,
             progress => console.log(progress.percentageProcessed),
             this.stage
           )
+          if (uploadDataFormRedirect) {
```
TomJKing, line 64: Does the redirect happen, only if the await function completes?

app/views/upload.scala.html
Lholbourn: I've given this a go and since it's now a GET rather than a POST we don't need the CSRF token :+1:

npm/src/upload/index.ts
Lholbourn: Yeah, it should only redirect if the upload has successfully finished not if it errors out or stalls for any reason :raised_hands:

app/views/upload.scala.html
```diff
@@ -47,6 +48,8 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
                 </button>
             </div>
         </form>
+<!--        Form to redirect user once upload has completed. It sends consignmentId to record processing placeholder page -->
+        @form(routes.RecordsController.recordsPage(consignmentId), Symbol("id") -> "upload-data-form") { }
```
TomJKing, line 52: Could this be rewritten without the additional mapping, eg: @form(routes.RecordsController.recordsPage(consignmentId)) {}

TomJKing: As discussed, this is resolved. As this is needed to find the form with js

PR #85, opened 2020-04-21
Batch metadata upload

TomJKing:


npm/test/clientfilemetataupload.test.ts
```diff
@@ -194,6 +203,41 @@ test("saveClientFileMetadata fails to upload client file metadata", async () =>
...
   )
 })
+
+test("createMetadataInputBatches generates batches of the defined size", () => {
+  const input1 = generateMockMetadataInput()
+  const input2 = generateMockMetadataInput()
+  const input3 = generateMockMetadataInput()
+  const input4 = generateMockMetadataInput()
+  const input5 = generateMockMetadataInput()
+
```
MancunianSam, line 216: The only thing I'm not sure of in this test is that each of these inputs has the same values, even with the Date.now() which is weird. This means that you can swap out the order of the inputs in the expect statement and the test still passes. I don't know if this matters because the point is that it splits them into arrays of the right size, not that they're in the right order. So if you think the order doesn't matter, then ignore me. If it does, you can make them unique by putting `Math.random().toString(36)` in the fileId and then the ordering matters.

npm/src/clientfilemetadataupload/index.ts
```diff
@@ -50,51 +57,82 @@ export class ClientFileMetadataUpload {
...
-    return {
-      input: {
+  generateClientFileData(
+    fileIds: string[],
+    metadata: IFileMetadata[]
+  ): IClientFileData[] {
+    const clientFilesData: IClientFileData[] = []
+
+    metadata.forEach((value, index) => {
+      const fileId = fileIds[index]
```
MancunianSam, line 95: This is personal preference but you can do a scala style `map` here to convert from from one to the other. 
So instead of pushing to the clientFilesData variable, you can change the foreach to map and return the object. Something like:
```
return metadata.map((value, index) => {
      const fileId = fileIds[index]
     ................
     ................
      return {
        metadataInput,
        tdrFile
      }
    })

```

npm/test/clientfilemetataupload.test.ts
TomJKing: Yeah good point. The order isn't particular important, but I think it should be checked as there is quite a lot of iterating through, and its good to keep the ordering consistent, I think. Made the change here to make the inputs unique to test the ordering corectly.

npm/src/clientfilemetadataupload/index.ts
TomJKing: Ah, that's pretty cool!

PR #88, opened 2020-04-23
Create groovy function to post jenkins build status to github

Lholbourn:
This function utilises the GitHub webhooks API to post Jenkins status information to Github for all branches/commits.

After the change from GitHub branch source plugin to the git branch source plugin,
we lost previous functionality that enabled up to push Jenkins build status to Github. Rather than add another plugin which would require a lot of set up (e.g. verifying which users can trigger build post), I decided to use the GitHub API as we already pull information from GitHub with other plugins, meaning adding another plugin may have resulted in multiple plugins which do very similar actions and could lead to confusion.

This function currently uses a temporary GitHub personal access key, which will need to be changed to make the process more secure and safe.

### DIFF
--- a/Jenkinsfile-testing
+++ b/Jenkinsfile-testing
@@ -1,8 +1,15 @@
 def versionTag = "v${env.BUILD_NUMBER}"
 
+def setBuildStatus(String status, String desc) {
+  withCredentials([string(credentialsId: 'github-jenkins-access-token', variable: 'GITHUB_ACCESS_TOKEN')]) {
+    String sha = sh(script: 'git rev-parse HEAD', returnStdout: true).trim()
+    String url = "https://api.github.com/repos/nationalarchives/tdr-transfer-frontend/statuses/${sha}"
+    sh "curl -XPOST '${url}' -H 'Authorization: bearer ${env.GITHUB_ACCESS_TOKEN}' --data '{\"state\":\"${status}\",\"target_url\":\"${env.BUILD_URL}\",\"description\":\"${desc}\",\"context\":\"Jenkins TDR Front end test\"}'"
+  }
+}
+
 pipeline {
   agent none
-
   stages {
     stage('Test') {
       agent {
@@ -11,6 +18,7 @@ pipeline {
         }
       }
       steps {
+        setBuildStatus("pending", "Jenkins build has started")
         checkout scm
         sh 'sbt -no-colors test scalastyle'
       }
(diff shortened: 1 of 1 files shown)


Jenkinsfile-testing
```diff
@@ -1,8 +1,15 @@
 def versionTag = "v${env.BUILD_NUMBER}"
 
+def setBuildStatus(String status, String desc) {
+  withCredentials([string(credentialsId: 'github-jenkins-access-token', variable: 'GITHUB_ACCESS_TOKEN')]) {
+    String sha = sh(script: 'git rev-parse HEAD', returnStdout: true).trim()
+    String url = "https://api.github.com/repos/nationalarchives/tdr-transfer-frontend/statuses/${sha}"
+    sh "curl -XPOST '${url}' -H 'Authorization: bearer ${env.GITHUB_ACCESS_TOKEN}' --data '{\"state\":\"${status}\",\"target_url\":\"${env.BUILD_URL}\",\"description\":\"${desc}\",\"context\":\"Jenkins TDR Front end test\"}'"
```
suzannehamilton, line 7: Did you look into whether any of the git plugins would do this step for us? The looks like it will work - I'm just wondering if Jenkins has a built-in tool.

Lholbourn: I did yeah, the plugins that I could find usually required a username/password rather than SSH/access keys or they required all GitHub users be 'whitelisted' to enable them to kick off a request (Which I thought didn't quite meet what we wanted). A few plugins also didn't work for multi-branch pipelines. I put together some notes of what I found and what they need if you'd like to look at them (Though I didn't take note of any that require a different Jenkins structure since I didn't feel it was necessary)?

suzannehamilton: If you have some notes, can you stick them on the Jira card? That might be the best place for them if we ever need to revisit this.

Lholbourn: Will do! 😄

suzannehamilton:
@Lholbourn I've just removed the temporary access key, and added one properly using Docker. It's called `github-jenkins-api-key`. Could you switch to that one and check it works?

Lholbourn:
> @Lholbourn I've just removed the temporary access key, and added one properly using Docker. It's called `github-jenkins-api-key`. Could you switch to that one and check it works?

Awesome will do! 😄

Lholbourn:
> OK, let's merge this and start using it! There's another card to extract the function to a shared Jenkins library, at which point we can come back and update this Jenkinsfile.

Awesome I'll merge it now! :smile:

PR #89, opened 2020-04-24
Cognito pool env var

MancunianSam:
The name of this branch is wrong because I realised after I started that all of the stage specific webpack variables needed to be replaced.

### DIFF
--- a/app/configuration/FrontEndInfoConfiguration.scala
+++ b/app/configuration/FrontEndInfoConfiguration.scala
@@ -1,0 +1,10 @@
+package configuration
+
+import javax.inject.Inject
+import play.api.Configuration
+import viewsapi.FrontEndInfo
+
+class FrontEndInfoConfiguration @Inject ()(configuration: Configuration) {
+  private def get(location: String) = configuration.get[String](location)
+  def frontEndInfo: FrontEndInfo = FrontEndInfo(get("consignmentapi.url"), get("cognito.identityProviderName"), get("cognito.identitypool"), get("environment"), get("region"))
+}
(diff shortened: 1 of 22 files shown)


app/configuration/FrontEndInfoConfiguration.scala
```diff
@@ -0,0 +1,10 @@
+package configuration
+
+import javax.inject.Inject
+import play.api.Configuration
+import viewsapi.FrontEndInfo
+
+class FrontEndInfoConfiguration @Inject ()(configuration: Configuration) {
```
TomJKing, line 7: Wonder whether "...Info..." is the best word to use in the name? Seems a bit vague. Although can't really think of a better one.

conf/application.conf
```diff
@@ -6,12 +6,19 @@ auth.url=${?AUTH_URL}
 auth.callback="http://localhost:9000/callback"
 auth.secret="placeholderSecretValue"
 auth.secret=${?AUTH_SECRET}
+
+cognito.identitypool=${?IDENTITY_POOL_ID}
+cognito.identityProviderName=auth.tdr-integration.nationalarchives.gov.uk/auth/realms/tdr
```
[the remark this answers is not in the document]
MancunianSam: No, local keycloak won't work and I don' think there's a way of setting up an iam identity provider to connect to a local server which is upsetting.

app/configuration/FrontEndInfoConfiguration.scala
MancunianSam: I'm not too pleased either, it was originally authenticationDetails or something like that but there's the api url and the stage so I tried to go for something generic.

conf/application.conf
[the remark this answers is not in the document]
TomJKing: Makes sense

app/configuration/FrontEndInfoConfiguration.scala
TomJKing: Fair enough

PR #91, opened 2020-04-27
Add error handling for upload

TomJKing:


npm/css-src/sass/overrides/_base.scss
```diff
@@ -1,3 +1,7 @@
 html {
     box-sizing: border-box;
 }
+
+.hide {
```
suzannehamilton, line 5: Should this be in the main SCSS files, rather than in the overrides folder? I assumed /overrides was for anything that overrides HTML defaults or styling in the libraries we use.

npm/test/errorhandling.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    '<form id="file-upload-form">' +
+    '<div class="govuk-error-summary hide">' +
+    '<p class="errorMessage">' +
+    "</p>" +
+    "</div>" +
+    "</form>"
+}
+
+function setupNonErrorHtml() {
+  document.body.innerHTML = 'div class="some-class">' + "</div>"
```
suzannehamilton, line 83: Is the div missing a `<`?

npm/test/errorhandling.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    ".errorMessage"
+  )
+
+  expect(() => {
+    handleUploadError(mockError)
+  }).toThrowError(new Error("Upload failed: some error"))
+})
+
+function checkExpectedErrorHtmlState(expectedRenderedErrorMessage: string) {
+  const mockFormElement: HTMLFormElement | null = document.querySelector(
```
suzannehamilton, line 47: Minor point, but in what sense are these mocks? They're more like the state of the test data after running the function under test. I'd just call them `formElement`, etc.

conf/messages
```diff
@@ -42,6 +42,8 @@ upload.selectFolder=Select a folder
 upload.fileExtensionTitle=Which file extensions are not permitted?
 upload.fileExtensionSummary=Cannot upload the following file extensions: .zip, thumbs.db, .exe
 upload.continueLink=Continue
+upload.error.title=There is a problem
```
suzannehamilton, line 45: In case anyone's wondering about the wording, we've got a card to come back and replace the placeholder error messages once we've done some work on the content design: https://national-archives.atlassian.net/browse/TDR-36

npm/test/errorhandling.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    "#file-upload-form"
+  )
+  const mockErrorElement: HTMLDivElement | null = document.querySelector(
+    ".govuk-error-summary"
+  )
+  const mockErrorMessageElement: HTMLParagraphElement | null = document.querySelector(
+    ".errorMessage"
+  )
+
+  if (mockFormElement) {
```
suzannehamilton, line 57: Conditionals in test assertions are quite dangerous - this test would pass even if the elements weren't present. It also makes it hard to read the tests because the reader has to work out if it _should_ be true or false for any given test.

I don't know the best way to do this in Typescript, but the assertions should expect the elements to be there.

app/views/upload.scala.html
```diff
@@ -8,6 +8,19 @@
         @progressIndicator(Messages("upload.progress"))
 
         <h1 class="govuk-heading-xl">@Messages("upload.title")</h1>
+        <div id="uploadError" class="govuk-error-summary hide" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
```
suzannehamilton, line 11: Minor point, but we're using kebab-case for CSS class names (partly because we're using [BEM](http://getbem.com/introduction/) as a naming scheme), so HTML IDs should be in kebab-case too. There might be elements in camelCase elsewhere in the codebase, but we should be as consistent as possible.

app/views/upload.scala.html
```diff
@@ -8,6 +8,19 @@
...
+        <h2 class="govuk-error-summary__title" id="error-summary-title">
+            @Messages("upload.error.title")
+        </h2>
+        <div class="govuk-error-summary__body">
+            <ul class="govuk-list govuk-error-summary__list">
+                <li>
+                    <p>@Messages("upload.error.failure")</p>
+                    <p class="errorMessage"></p>
```
suzannehamilton, line 19: I think by BEM rules, the parent div (the one that currently has an ID of `uploadError` should have a class that's something like `upload-error`, and this element would then be `<p class=upload-error__message>`.

npm/test/errorhandling.test.ts
TomJKing: Fair point

npm/test/errorhandling.test.ts
TomJKing: Added a comment into the file to this affect.

app/views/upload.scala.html
TomJKing: Changed. Sorry I forgot we had decided on using BEM

app/views/upload.scala.html
TomJKing: Changed

conf/messages
TomJKing: Added a comment to the file to say placeholder wording

npm/css-src/sass/overrides/_base.scss
TomJKing: Moved to main.scss

npm/test/errorhandling.test.ts
suzannehamilton: They'll fail for the right reason now, which is good! Hopefully we'll learn more about how to write good Jest tests as we go. 🙂

PR #99, opened 2020-05-01
Changed favicon from GOV UK logo to TNA logo

rsulxo:
#### Jira card: https://national-archives.atlassian.net/browse/TDR-54?atlOrigin=eyJpIjoiNjNjNjU1Y2E0YjJiNDA1MGE2NWY1OWIzNjcwMGE4MGEiLCJwIjoiaiJ9

This PR is to replace the favicon from the gov.uk logo to TNA logo.

[TDR-54] TNA Favicon Logo
Items changed:
- Added TNA favicon.ico file within `public/images`.

[TDR-54]: https://national-archives.atlassian.net/browse/TDR-54


npm/package.json
```diff
@@ -7,6 +7,7 @@
     "copy-govuk-image-assets": "copyfiles -f node_modules/govuk-frontend/govuk/assets/images/* ../public/images",
     "copy-govuk-font-assets": "copyfiles -f node_modules/govuk-frontend/govuk/assets/fonts/* ../public/fonts",
     "copy-govuk-js-assets": "copyfiles -f node_modules/govuk-frontend/govuk/all.js ../public/javascripts",
+    "copy-tdr-image-assets": "copyfiles -f assets/images/*",
```
suzannehamilton, line 10: For the header logo, Nigel's tried adding the image to assets/images directly: https://github.com/nationalarchives/tdr-transfer-frontend/pull/101/files

That might be easier in this case too.

rsulxo: I tried this approach but for some reason the favicon wouldn't change, Tom also attempted it but didn't work. I had to change the file path within the `main.scala.html` file which worked!

suzannehamilton: Sorry, I mean public/images. If you copy it to there, you shouldn't need the extra script in package.json.

rsulxo: Yes that was the first thing I tried essentially to keep things simple but it didn’t work, it is the same thing Tom tried to do too but, I had to go through with it using the script approach eventually.

suzannehamilton: @TomJKing Do you know why just putting the image in the folder doesn't work? It seems like it should be simpler than adding a script to the do the same thing.

TomJKing: No idea. It seems that the favicon has some pretty hefty caching, which prevents easy updating

suzannehamilton:
@rsulxo @TomJKing I just tried adding the image to public/images, and it seems to work for me. Any idea what's different to when you tried it?

Tom, would you mind reviewing my commit if you're happy that it works?

rsulxo:
@suzannehamilton yeah it's working now! That's so odd, I have no idea why this didn't work when I tried it...quite confusing to me. But at least this way there are no extra scripts, so that's good!

PR #101, opened 2020-05-10
Replace gov.uk logo with TNA logo

techncl:


app/views/main.scala.html
```diff
@@ -26,15 +26,7 @@
...
-
-                            <image src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image"></image>
-                        </svg>
-                        <span class="govuk-header__logotype-text">
-                            @Messages("main.logo")
-                        </span>
+                        <image src="/assets/images/tna-horizontal-logo-negative.png" width="200"></image>
```
suzannehamilton, line 29: Does this image need to be added? You might have to update the .gitignore so that it's not ignored. I'd suggest changing line 7 to `public/images/govuk-*` so git ignores all GOV.UK assets (which are populated by the npm scripts) and includes all of our images.

PR #102, opened 2020-05-12
Add redirect functionality for once upload is complete

Lholbourn:
This change redirects the user to the 'processing your records' placeholder page once the s3 upload has completed. It is done using a second hidden form on the upload page, the action of which is to render the page that the user is redirected to. This hidden form is only submitted once the upload has completed.

This change still needs tests to be added to it.


Lholbourn:
Accidentally made a new branch rather than using the one I was using originally due to working on a different computer. Have now amended this.

PR #103, opened 2020-05-14
Add post build action to Jenkinsfile to run e2e tests.

Lholbourn:
This change triggers the e2e tests to run on intg after an intg deployment has been started successfully. There is a potential issue in that the deployment may not have finished and the load balancer may not be sending traffic to the newly deployed version yet. As a small workaround to this, the e2e tests Jenkinsfile will be modified to include a 'quietPeriod' in which it waits a pre-configured amount of time to start building after a build is initially triggered.

A more sophisticated solution to this problem may exist, and I will look into it to see what can be done to fix it.

### DIFF
--- a/Jenkinsfile-deploy
+++ b/Jenkinsfile-deploy
@@ -54,15 +54,15 @@ pipeline {
   }
   post {
     success {
-        script {
-            if (params.STAGE == "intg"){
-                build(
-                    job: "TDRAcceptanceTest",
-                    parameters: [ string(name: "STAGE", value: "intg") ],
-                    wait: false
-                )
-            }
+      script {
+        if (params.STAGE == "intg"){
+          build(
+            job: "TDRAcceptanceTest",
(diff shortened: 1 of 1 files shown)


Jenkinsfile-deploy
```diff
@@ -52,6 +52,19 @@ pipeline {
       }
     }
   }
+  post {
+    success {
+        script {
```
suzannehamilton, line 57: Formatting nitpick: the rest of the file is indented by two spaces.

Lholbourn:
> This looks good to me! Just one formatting comment.
> 
> We should have a think about what other jobs should trigger the end-to-end tests. Off the top of my head: the API deployment, Terraform environments, DB migrations.
> 
> This is another case where it would be handy to have this function in a shared Jenkins library. Are you planning to look into that as part of the GitHub branch work?

I was planning to start looking into that for the build status stuff but if it's useful for the e2e tests as well I can add a card to the e2e story too and look into both at the same time?

PR #105, opened 2020-05-19
Update progress bar with S3 upload progress

rsulxo:
This PR is for the progress bar on the upload page. What it does is extract metadata until 50% and then extract s3 uploads until it reaches 100%. Once it reaches completion, the page is then redirected to the records placeholder page. 

The progress bar now does not reset to 0% nor does it reach completion immediately, it shows a seamless process and looks exactly like the wireframes.

[TDR-22]: https://national-archives.atlassian.net/browse/TDR-22

### DIFF
--- a/npm/src/clientfileprocessing/index.ts
+++ b/npm/src/clientfileprocessing/index.ts
@@ -23,7 +23,21 @@ export class ClientFileProcessing {
     this.s3Upload = s3Upload
   }
 
-  progressCallback(progressInformation: IProgressInformation) {
+  /**
+   * Updates a progress bar with a given percentage
+   * @param percent The percentage to update the progress bar with  of type {@link Number}
+   */
+  updateProgressBarPercentage(percent: Number) {
+    const progressBarElement: HTMLDivElement | null = document.querySelector(
+      ".progress-display"
+    )
+
+    if (progressBarElement) {
+      progressBarElement.setAttribute("value", percent.toString())
+    }
+  }
+
+  progressBarSwitcher = (progressInformation: IProgressInformation) => {
     console.log(
       "Percent of metadata extracted: " +
         progressInformation.percentageProcessed
@@ -35,30 +49,14 @@ export class ClientFileProcessing {
     const progressBar: HTMLDivElement | null = document.querySelector(
       "#progress-bar"
     )
-    const progressBarElement: HTMLDivElement | null = document.querySelector(
-      ".progress-display"
-    )
 
     if (fileUpload && progressBar) {
       fileUpload.classList.add("hide")
       progressBar.classList.remove("hide")
     }
 
-    if (progressBarElement && progressInformation.percentageProcessed < 50) {
-      progressBarElement.setAttribute(
-        "value",
-        progressInformation.percentageProcessed.toString()
-      )
-    }
-  }
-
-  progressS3Callback() {
-    const progressBarElement: HTMLDivElement | null = document.querySelector(
-      ".progress-display"
-    )
-
-    if (progressBarElement) {
-      progressBarElement.setAttribute("value", "100")
+    if (progressInformation.percentageProcessed < 50) {
+      this.updateProgressBarPercentage(progressInformation.percentageProcessed)
     }
   }
 
@@ -75,8 +73,7 @@ export class ClientFileProcessing {
       )
       const metadata: IFileMetadata[] = await this.clientFileExtractMetadata.extract(
         files,
-        //Temporary function until s3 upload in place
-        this.progressCallback
+        this.progressBarSwitcher
       )
       const tdrFiles = await this.clientFileMetadataUpload.saveClientFileMetadata(
         fileIds,
@@ -85,7 +82,7 @@ export class ClientFileProcessing {
       this.s3Upload.uploadToS3(
         consignmentId,
         tdrFiles,
-        this.progressS3Callback,
+        () => this.updateProgressBarPercentage(100), // Update to 100 as we have finished at this point.
         stage
       )
     } catch (e) {
--- a/npm/src/clientfileprocessing/index.ts
+++ b/npm/src/clientfileprocessing/index.ts
@@ -24,6 +24,8 @@ export class ClientFileProcessing {
   }
 
   progressMetadataCallback(progressInformation: IProgressInformation) {
+    const weightedPercent = progressInformation.percentageProcessed / 2
+
     const fileUpload: HTMLDivElement | null = document.querySelector(
       "#file-upload"
     )
@@ -39,21 +41,20 @@ export class ClientFileProcessing {
       progressBar.classList.remove("hide")
     }
 
-    if (progressBarElement && progressInformation.percentageProcessed < 50) {
-      progressBarElement.setAttribute(
-        "value",
-        progressInformation.percentageProcessed.toString()
-      )
+    if (progressBarElement) {
+      progressBarElement.setAttribute("value", weightedPercent.toString())
     }
   }
 
-  progressS3Callback() {
+  s3ProgressCallback(progressInformation: IProgressInformation) {
+    const weightedPercent = 50 + progressInformation.percentageProcessed / 2
+
     const progressBarElement: HTMLDivElement | null = document.querySelector(
       ".progress-display"
     )
 
     if (progressBarElement) {
-      progressBarElement.setAttribute("value", "100")
+      progressBarElement.setAttribute("value", weightedPercent.toString())
     }
   }
 
@@ -79,7 +80,7 @@ export class ClientFileProcessing {
       await this.s3Upload.uploadToS3(
         consignmentId,
         tdrFiles,
-        this.progressS3Callback,
+        this.s3ProgressCallback,
         stage
       )
     } catch (e) {
--- a/npm/src/clientfileprocessing/index.ts
+++ b/npm/src/clientfileprocessing/index.ts
@@ -23,31 +23,15 @@ export class ClientFileProcessing {
     this.s3Upload = s3Upload
   }
 
-  /**
-   * Updates a progress bar with a given percentage
-   * @param percent The percentage to update the progress bar with  of type {@link Number}
-   */
-  updateProgressBarPercentage(percent: Number) {
-    const progressBarElement: HTMLDivElement | null = document.querySelector(
-      ".progress-display"
-    )
-
-    if (progressBarElement) {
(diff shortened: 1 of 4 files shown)


npm/src/clientfileprocessing/index.ts
```diff
@@ -65,14 +73,18 @@ export class ClientFileProcessing {
...
       )
       const tdrFiles = await this.clientFileMetadataUpload.saveClientFileMetadata(
         fileIds,
         metadata
       )
-      this.s3Upload.uploadToS3(consignmentId, tdrFiles, callback, stage)
+      this.s3Upload.uploadToS3(
+        consignmentId,
+        tdrFiles,
+        () => this.updateProgressBarPercentage(100), // Update to 100 as we have finished at this point.
```
TomJKing, line 85: I don't think this will work as this just sets the display to 100% without updating the progress of the s3 upload. 

See above comments for refactoring the functions slightly, and then call as follows. It's a tricky problem here as we have got two stages to deal with, both independently supply a percent processed, but starting a 0% each time.
```
...
try {
      const fileIds: string[] = await this.clientFileMetadataUpload.saveFileInformation(
        ...
      )

      //Set maximum progress percent to 50 as extract metadata first part of process
      this.setProgressBarDisplay(50)

      const metadata: IFileMetadata[] = await this.clientFileExtractMetadata.extract(
        files,
        this.updateProgressBar
      )
      const tdrFiles = await this.clientFileMetadataUpload.saveClientFileMetadata(
        ...
      )

      //Set maximum progress percent to 100 as upload final part of process
      this.setProgressBarDisplay(100)

      this.s3Upload.uploadToS3(
        consignmentId,
        tdrFiles,
        this.updateProgressBar,
        stage
      )
...
```

npm/src/clientfileprocessing/index.ts
```diff
@@ -23,7 +23,21 @@ export class ClientFileProcessing {
...
+    const progressBarElement: HTMLDivElement | null = document.querySelector(
+      ".progress-display"
+    )
+
+    if (progressBarElement) {
+      progressBarElement.setAttribute("value", percent.toString())
+    }
+  }
+
+  progressBarSwitcher = (progressInformation: IProgressInformation) => {
```
TomJKing, line 40: I would refactor this function to something like this. You can then set the max percent before calling the two stages that we are showing the percentage processed for:
1. Client file metadata upload
2. File upload

```
/**
   * Displays the progress bar setting the maximum percentage
   * @param maxPercent The maximum percentage to display on progress bar of type {@link Number}
   */
  setProgressBarDisplay(maxPercent: Number) {
    const fileUpload: HTMLDivElement | null = document.querySelector(
      "#file-upload"
    )
    const progressBar: HTMLDivElement | null = document.querySelector(
      "#progress-bar"
    )
    const progressBarElement: HTMLDivElement | null = document.querySelector(
      ".progress-display"
    )

    if (fileUpload && progressBar && progressBarElement) {
      fileUpload.classList.add("hide")
      progressBar.classList.remove("hide")
      progressBarElement.setAttribute("max", maxPercent.toString())
    }
  }
```

npm/src/clientfileprocessing/index.ts
```diff
@@ -23,7 +23,21 @@ export class ClientFileProcessing {
     this.s3Upload = s3Upload
   }
 
-  progressCallback(progressInformation: IProgressInformation) {
+  /**
+   * Updates a progress bar with a given percentage
+   * @param percent The percentage to update the progress bar with  of type {@link Number}
+   */
+  updateProgressBarPercentage(percent: Number) {
```
TomJKing, line 30: I would refactor this function slightly to something like this. This can be reused as the call back for the different stages of the upload process:
1. Client file metadata upload
2. File upload

```
updateProgressBar(progressInformation: IProgressInformation) {
    const progressBarElement: HTMLDivElement | null = document.querySelector(
      ".progress-display"
    )

    if (progressBarElement) {
      progressBarElement.setAttribute("value", progressInformation.percentageProcessed.toString())
    }
  }
```

npm/src/clientfileprocessing/index.ts
```diff
@@ -44,11 +41,20 @@ export class ClientFileProcessing {
...
+    const weightedPercent = 50 + progressInformation.percentageProcessed / 2
+
+    const progressBarElement: HTMLDivElement | null = document.querySelector(
+      ".progress-display"
+    )
+
+    if (progressBarElement) {
+      progressBarElement.setAttribute("value", weightedPercent.toString())
     }
   }
```
TomJKing, line 60: At line 64 the following parameter is passed into the processClientFile function: `callback: TProgressFunction`

This is no longer needed so should be removed.

npm/src/clientfileprocessing/index.ts
```diff
@@ -23,11 +23,8 @@ export class ClientFileProcessing {
     this.s3Upload = s3Upload
   }
 
-  progressCallback(progressInformation: IProgressInformation) {
-    console.log(
-      "Percent of metadata extracted: " +
-        progressInformation.percentageProcessed
-    )
+  progressMetadataCallback(progressInformation: IProgressInformation) {
```
TomJKing, line 26: I would rename this function: metadataProgressCallback.

npm/test/clientfileprocessing.test.ts
```diff
@@ -172,7 +172,7 @@ test("client file metadata successfully uploaded", async () => {
   ).resolves.not.toThrow()
 })
```
TomJKing, line 174: I would refactor the test slightly so you have tests as follows to have the following tests:
- "metadataProgressBarCallback function updates the progress bar with the percentage processed"
- "metadataProgressBarCallback function updates the progress bar to a maximum of 50 percent"
-- with this test set the percent processed to 100% to make sure the @value displays only 50
- "metadataProgress function does not change the HTML state if no progress bar present"
- "s3ProgressBarCallback function updates progress bar from a minimum of 50 percent"
-- with this test set the percent processed to 0% and test the @value displays 50
-  "s3ProgressBarCallback function updates the progress bar with the percentage processed"
- "s3ProgressBarCallback function updates the progress bar to a maximum of 100 percent"

This set of tests should then cover the different behaviours of the the two progress bar callback functions. For the s3 function tests you can write a separate check function to make sure the @value is set correctly

npm/test/clientfileprocessing.test.ts
```diff
@@ -172,12 +172,12 @@ test("client file metadata successfully uploaded", async () => {
...
 
-test("progressCallback function updates the progress bar with the percentage processed", () => {
+test("metadataProgressBarCallback function updates the progress bar to a maximum of 50 percent", () => {
```
TomJKing, line 175: The function name in the test description should match the actual function name. This is the same for all the new tests.

npm/test/clientfileprocessing.test.ts
```diff
@@ -261,6 +261,78 @@ test("file successfully uploaded to s3", async () => {
   expect(s3UploadMock.uploadToS3).toHaveBeenCalledTimes(1)
 })
 
+test("s3ProgressBarCallback function updates the progress bar with the percentage processed", () => {
+  setupUploadPageHTML()
+
+  const progressInformation: IProgressInformation = {
+    totalFiles: 1,
+    percentageProcessed: 100,
```
TomJKing, line 269: I would test with a percentage other than 100 here. You test this at the end, and with this test we are a checking the weighting calculation is working as expected. Would say try 75 percent.

npm/test/clientfileprocessing.test.ts
```diff
@@ -394,6 +466,16 @@ function checkExpectedPageState(percentage: String) {
   ).toEqual(percentage)
 }
 
+function checkS3ExpectedPageState(percentage: String) {
```
TomJKing, line 469: I would rename this function: checkS3UploadProgressBarState

npm/test/clientfileprocessing.test.ts
```diff
@@ -215,13 +214,13 @@ test("progressCallback function does not update progress bar if percentage proce
...
-  fileProcessing.progressCallback(progressInformation)
+  fileProcessing.metadataProgressCallback(progressInformation)
 
-  checkExpectedPageState("") // this field is not populated because % > 50
+  checkNoPageStateChangeExpected()
 })
 
-test("progressCallback function does not change the HTML state if no progress bar present", () => {
-  setupUploadPageHTMLWithoutProgressBar()
+test("metadataProgressBarCallback function updates the progress bar with the percentage processed", () => {
```
TomJKing, line 222: Name the function in the test description as the actual function name: "metadataProgressCallback"

npm/test/clientfileprocessing.test.ts
```diff
@@ -255,12 +254,84 @@ test("file successfully uploaded to s3", async () => {
...
   const fileProcessing = new ClientFileProcessing(metadataUpload, s3UploadMock)
   await expect(
-    fileProcessing.processClientFiles("1", [], jest.fn(), "")
+    fileProcessing.processClientFiles("1", [], "")
   ).resolves.not.toThrow()
 
   expect(s3UploadMock.uploadToS3).toHaveBeenCalledTimes(1)
 })
 
+test("s3ProgressBarCallback function updates the progress bar with the percentage processed", () => {
```
TomJKing, line 263: Name the function in the test description as the actual function name: "s3ProgressCallback"

npm/test/clientfileprocessing.test.ts
```diff
@@ -255,12 +254,84 @@ test("file successfully uploaded to s3", async () => {
...
+    metadataUpload,
+    new S3UploadMock()
+  )
+
+  fileProcessing.s3ProgressCallback(progressInformation)
+
+  checkS3UploadProgressBarState("87.5")
+})
+
+test("s3ProgressBarCallback function updates progress bar from a minimum of 50 percent", () => {
```
TomJKing, line 287: Name the function in the test description as the actual function name: "s3ProgressCallback"

npm/test/clientfileprocessing.test.ts
```diff
@@ -255,12 +254,84 @@ test("file successfully uploaded to s3", async () => {
...
+    metadataUpload,
+    new S3UploadMock()
+  )
+
+  fileProcessing.s3ProgressCallback(progressInformation)
+
+  checkS3UploadProgressBarState("50")
+})
+
+test("s3ProgressBarCallback function updates the progress bar to a maximum of 100 percent", () => {
```
TomJKing, line 311: Name the function in the test description as the actual function name: "s3ProgressCallback"

PR #109, opened 2020-06-05
Position TNA logo properly on MVC site

rsulxo:
This card consists of vertically centring the logo and also fixing the link highlighting which now overlaps with TNA logo perfectly.

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -24,7 +24,7 @@
         <div class="govuk-header__container govuk-width-container">
 
             <div class="govuk-header__tna-logo">
-                <a href="/" class="govuk-header__link govuk-header__link--homepage">
+                <a href="/" class="govuk-header__tna_link govuk-header__tna_link--homepage">
                         <image src="/assets/images/tna-horizontal-white-logo.svg" class="govuk-header__tna-logo-image" alt="TNA horizontal logo"></image>
                 </a>
             </div>
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -24,14 +24,12 @@
         <div class="govuk-header__container govuk-width-container">
 
             <div class="govuk-header__tna-logo">
-                <a href="/" class="govuk-header__link">
-                    <span>
-                        <image src="/assets/images/tna-horizontal-white-logo.svg" class="govuk-header__tna-logo-image"></image>
-                    </span>
+                <a href="/" class="govuk-header__link govuk-header__link--homepage">
+                        <image src="/assets/images/tna-horizontal-white-logo.svg" class="govuk-header__tna-logo-image" alt="TNA horizontal logo"></image>
                 </a>
             </div>
 
-            <div class="govuk-header__content">
+            <div class="govuk-header__tna-content">
                 <a href="#" class="govuk-header__link govuk-header__link--service-name">
                     @Messages("main.header")
                 </a>
--- a/npm/css-src/sass/_header.scss
+++ b/npm/css-src/sass/_header.scss
@@ -1,3 +1,5 @@
+$govuk-header-link: govuk-colour("white");
+
 .govuk-header__tna-logo {
     width: 220px;
     float: left;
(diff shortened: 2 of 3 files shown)


npm/css-src/sass/_header.scss
```diff
@@ -1,8 +1,16 @@
...
-    width: 33.33%;
+    width: 220px;
     float: left;
 }
 
 .govuk-header__tna-logo-image {
     width: 220px;
+}
+
+.govuk-header__tna-content {
```
suzannehamilton, line 10: Where is the HTML element that this rule applies to? I can't find it, but I might be looking in the wrong place. Or is there a change to the templates that hasn't been checked in?

rsulxo: This rule is applied to the HTML element on line 32 within `main.scala.html`

suzannehamilton: Looking at https://github.com/nationalarchives/tdr-transfer-frontend/blob/position-logo/app/views/main.scala.html#L32 I can see `govuk-header__tna-logo` but not `.govuk-header__tna-logo-image` or `.govuk-header__tna-content`:

```
            <div class="govuk-header__tna-logo">
                <a href="/" class="govuk-header__link">
                    <span>
                        <image src="/assets/images/tna-horizontal-white-logo.svg" class="govuk-header__tna-logo-image"></image>
                    </span>
                </a>
            </div>

            <div class="govuk-header__content">
                <a href="#" class="govuk-header__link govuk-header__link--service-name">
                    @Messages("main.header")
                </a>
            </div>
```

app/views/main.scala.html
```diff
@@ -24,14 +24,12 @@
         <div class="govuk-header__container govuk-width-container">
 
             <div class="govuk-header__tna-logo">
-                <a href="/" class="govuk-header__link">
-                    <span>
-                        <image src="/assets/images/tna-horizontal-white-logo.svg" class="govuk-header__tna-logo-image"></image>
-                    </span>
+                <a href="/" class="govuk-header__link govuk-header__link--homepage">
```
suzannehamilton, line 27: This link has an underline when you hover over it, which looks a bit odd because it's an image rather than text:

Screenshot 2020-06-10 at 11 33 34

rsulxo: This was a case of the default govuk scss files including an underline on the logo + text, so I had to override it, which is fine as it is fixed now.

npm/css-src/sass/_header.scss
```diff
@@ -1,8 +1,16 @@
...
+    width: 290px;
+    position: relative;
+    align-self: center;
+    top: 5;
```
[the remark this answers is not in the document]
rsulxo: I hadn't noticed this before but it doesn't get applied in my browser either, so I have now removed it as it is not required.

npm/css-src/sass/_header.scss
```diff
@@ -1,8 +1,16 @@
...
 }
 
 .govuk-header__tna-logo-image {
     width: 220px;
+}
+
+.govuk-header__tna-content {
+    width: 290px;
+    position: relative;
+    align-self: center;
```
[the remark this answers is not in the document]
rsulxo: Since 'align-self' property overrides a grid or flex item's align-items value but as we aren't using grid or flex items, it is not required here and it doesn't affect anything with regards to how the header is positioned so it has been removed.

npm/css-src/sass/_header.scss
```diff
@@ -1,8 +1,61 @@
...
+    font-size: 30px; // We don't have a mixin that produces 30px font size
+    line-height: 1;
+
+    &:link,
+    &:visited {
+      text-decoration: none;
+    }
+
+    &:active {
+      box-sizing: border-box;
```
suzannehamilton, line 55: Great, I'm glad to see this simplifies the styling. 👍 

I've just noticed that we've got a top-level box-sizing rule in [_base.scss](https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/npm/css-src/sass/overrides/_base.scss), but it's not being applied to every element.

It would be good to change it to the more general:

```
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```

If you do that, I think you can remove the `box-sizing` lines from this file.

I don't think this affects the elements styled with the GOV.UK design system, but you might want to check the progress bar still looks how you expect.

rsulxo: Awesome, I tried putting box-sizing at top level and it seems to be working fine and made no visible changes, the progress bar looks intact too. 😄

PR #112, opened 2020-06-19
Remove local API setup instructions that we no longer use

Lholbourn:
When running the consignment API locally, the environment variable AUTH_URL no longer needs to be set as it is defined in the application.conf files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,9 +102,6 @@ If you want to upload to the aws s3 bucket, you can no longer use your local key
 - AUTH_SECRET=\<insert the secret for the tdr client in the integration account\>
 - IDENTITY_POOL_ID=secret-from-/mgmt/identitypoolid_intg in the management account
 
-You will also need to set the AUTH_URL environment variable for your locally running API project to:
--  AUTH_URL=https://auth.tdr-integration.nationalarchives.gov.uk/auth
-
 #### Frontend project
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,9 +102,6 @@ If you want to upload to the aws s3 bucket, you can no longer use your local key
 - AUTH_SECRET=\<insert the secret for the tdr client in the integration account\>
 - IDENTITY_POOL_ID=secret-from-/mgmt/identitypoolid_intg in the management account
 
-You will also need to set the AUTH_URL environment variable for your locally running API project to:
--  AUTH_URL=https://auth.tdr-integration.nationalarchives.gov.uk/auth
```
suzannehamilton, line 106: This step wasn't necessary in your case because you were running a fully local development environment, but it is necessary if you follow the step just before it, and connect to a local API and the integration auth server.

This setup will get simpler once I've finished setting up some purely local backend checks, so it might be easiest to leave this for now.

Lholbourn: Ah whoops! I'll roll that back, should I add a note to say when it should be used?

suzannehamilton: I think it's fine to close this PR. The current docs are a bit confusing, but I'll be making some sweeping changes to them soon when I've finished the local dev environment setup.

PR #120, opened 2020-07-30
Add exit prompt to upload page

techncl:


suzannehamilton:
It looks like some of the checkbox changes from #119 have got mixed up with this branch - I'm guessing because this branch started on top of the other branch, rather than on master. The checkbox changes have been merged already, so it's a bit confusing to have them in this branch as well.

Maybe we can sort out the git history when we're pairing later today?

techncl:
Ah, yes. Sure! It's weird as the commit is exactly the same.

suzannehamilton:
@TomJKing Would you mind reviewing this one? Nigel and I paired on refactoring it, so it would be good to get a third person's opinion, especially since you're more familiar with Jest tests than me. 🙂 

@techncl You can update the title of this PR now we've edited it to remove the old SCSS commit.

npm/src/upload/index.ts
```diff
@@ -2,45 +2,49 @@ import { TdrFile } from "@nationalarchives/file-information"
...
+    window.addEventListener("beforeunload", pageUnloadAction)
+
+    try {
+      await this.clientFileProcessing.processClientFiles(
+        consignmentId,
+        files,
+        this.stage
+      )
+      // In order to prevent exit confirmation when page redirects to Records page
+      window.removeEventListener("beforeunload", pageUnloadAction) // Move to the goToNextPage method?
```
TomJKing, line 43: Is the comment: "Move to the goToNextPage method?" meant to be here? Little bit confusing whether it is meant to indicate a future enhancement?

npm/src/upload/upload-form.ts
```diff
@@ -0,0 +1,47 @@
+import { TdrFile } from "@nationalarchives/file-information"
+import { stringType } from "aws-sdk/clients/iam"
```
TomJKing, line 2: Not sure the `stringType` import is used?

npm/test/upload.test.ts
```diff
@@ -56,137 +56,44 @@ const mockUploadFailure: () => void = () => {
...
-  uploadFiles.upload()
-  if (uploadForm) {
-    await uploadForm.submit()
-    expect(mockRetrieveFiles).toBeCalledTimes(0)
-    expect(mockUploadFilesSuccess).toBeCalledTimes(0)
-  }
+  await uploadFiles.uploadFiles([dummyFile], "12345")
 
   expect(consoleErrorSpy).toHaveBeenCalled()
+  expect(mockGoToNextPage).toBeCalledTimes(0)
```
TomJKing, line 89: Probably change this to the following so its more explicit when read: `expect(mockGoToNextPage).not.toHaveBeenCalled()`

npm/src/upload/index.ts
techncl: That is weird, I thought I removed it. I'll do it now; thanks!

npm/src/upload/upload-form.ts
techncl: Correct, I'll remove it.

npm/test/upload.test.ts
techncl: Sure, I'll remove it. Thanks

PR #121, opened 2020-08-03
Upgrade file-information dependency

suzannehamilton:
The main reason is the incorporate this fix to file-information: https://github.com/nationalarchives/tdr-file-metadata/pull/24


suzannehamilton:
@TomJKing Thanks for reviewing! I had to fix a merge conflict, so would you mind re-approving (assuming you're still happy with the change)?

PR #133, opened 2020-08-21
Render checksum progress bar

techncl:


app/views/fileChecksProgress.scala.html
```diff
@@ -9,8 +9,14 @@
             @progressIndicator(Messages("checkingRecords.progress"))
             <h1 class="govuk-heading-xl">@Messages("checkingRecords.title")</h1>
             <p class="govuk-body">@Messages("checkingRecords.description")</p>
-            <p class="govuk-body">@Messages("checkingRecords.antivirus")</p>
-            <progress id="av-metadata-progress-bar" class="progress-display" value="@avMetadataPercentage" max="100"></progress>
+            <p>
+                <label for="av-metadata-progress-bar" class="govuk-body">@Messages("checkingRecords.antivirus")</label>
```
suzannehamilton, line 13: Just to check - what's the advantage of putting this header into a `label` element? My understanding was that `label` was for form elements.

app/views/fileChecksProgress.scala.html
```diff
@@ -9,8 +9,14 @@
             @progressIndicator(Messages("checkingRecords.progress"))
             <h1 class="govuk-heading-xl">@Messages("checkingRecords.title")</h1>
             <p class="govuk-body">@Messages("checkingRecords.description")</p>
-            <p class="govuk-body">@Messages("checkingRecords.antivirus")</p>
-            <progress id="av-metadata-progress-bar" class="progress-display" value="@avMetadataPercentage" max="100"></progress>
+            <p>
```
suzannehamilton, line 12: Could the `<progress>` element be removed from the `<p>` element? Normally, paragraph elements should just contain text - it will make any custom CSS less confusing if we keep the text and other visual displays separate.

app/controllers/FileChecksController.scala
```diff
@@ -35,10 +37,11 @@ class FileChecksController @Inject()(val controllerComponents: SecurityComponent
         fileChecks => Ok(views.html.fileChecksProgress(
           consignmentId,
           fileChecks.totalFiles,
```
suzannehamilton, line 39: Now that we're passing more and more things to the view, I'm wondering if it would be easier to follow if we passed the `FileChecks` object itself, rather than passing all of its contents?

app/views/fileChecksProgress.scala.html
techncl: From what I understood, the label tag enables it to be read out aloud by screen readers - https://www.w3schools.com/tags/tag_label.asp

W3schools and Mozilla both use it in conjunction with the progress tag 
- https://www.w3schools.com/tags/tag_progress.asp
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

app/views/fileChecksProgress.scala.html
techncl: Ah, yes, I added the <p> tag around them because without them, the separation between the two bars looked a bit off. Maybe I should play around with the <br> tag or line-height instead?

app/views/fileChecksProgress.scala.html
suzannehamilton: I didn't realise that, thanks!

app/views/fileChecksProgress.scala.html
suzannehamilton: Would you like to pair on this later today? I've got a few ideas for how to approach it, and it would be good to work it out together.

app/controllers/FileChecksController.scala
suzannehamilton: Any thoughts on this one?

techncl: Ah, yes, sorry, I forgot to reply. Yes, that is a good idea; I started working on it yesterday.

app/views/fileChecksProgress.scala.html
techncl: Sure!

suzannehamilton:
It looks like your latest commits have been push to the #134 branch rather than this one.

Give me a shout if you want some help with git.

techncl:
:man_facepalming:. Thank you! Yes, I think I might need your help; I tried reverting just now but ran into an issue

suzannehamilton (approved):
Looks good! 🚀

PR #134, opened 2020-08-21
Rename file progress

techncl:


suzannehamilton (commented):
I think this change makes sense, but I'll hold off on reviewing it properly until #133 is merged because the commits from #133 have been included in this PR too. Once that's done, let's clean up the git history for this branch. 👍

suzannehamilton (approved):

PR #135, opened 2020-09-02
Make the integration backend the default when running locally

suzannehamilton:
Split the local dev config files into two:

- application.conf: This is the default, and is the one that will be used when you run `sbt run`. It connects to the integration backend. You still need to set the `AUTH_SECRET` and `IDENTITY_POOL_ID` environment variables, because they're not safe to store in the config file.
- application.local-full-stack.conf: This can be selected if you run `sbt  -Dconfig.file=conf/application.local-full-stack.conf run`, and will connect to the local Keycloak, API, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ Repository for TDR transfer code
 
 There are two ways to develop this project:
 
-- Frontend development only, using the AWS integration environment for everything else
-- Full stack local development, using a local dev copy of the API, Keycloak, etc.
+- Frontend development only, using the AWS integration environment for everything else. This is the default
+- Full stack local development, using a local dev copy of the API, Keycloak, etc
 
 ### Prerequisites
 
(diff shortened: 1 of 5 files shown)


conf/application.local-base.conf
```diff
@@ -0,0 +1,38 @@
+# This base configuration file that contains values used in local development, regardless of which backend
+# is being used (local or integration).
+
+# https://www.playframework.com/documentation/latest/Configuration
+play.http.port=9000
+
+auth.callback="http://localhost:9000/callback"
+auth.secret=${AUTH_SECRET}
+
+play.cache.redis {
```
suzannehamilton, line 10: Everything below this line has been copied straight from application.conf, because it's the same in both cases (fully local dev and integration backend).

README.md
```diff
@@ -83,15 +107,9 @@ Run an [S3 ninja] Docker container, specifying a local directory in which to sav
 docker run -d -p 9444:9000 -v <some directory path>:/home/sirius/data --name=s3ninja scireum/s3-ninja:6.4
 ```
 
-Set the `S3_ENDPOINT_OVERRIDE` environment variable to the upload URL of the emulator, in this case
```
suzannehamilton, line 86: These environment variables are no longer necessary because they're set in the new config file.

TomJKing:
Looks like the build is failing for this branch: https://jenkins.tdr-management.nationalarchives.gov.uk/job/TDR%20Transfer%20Front%20End%20Test/job/local-dev/1/console

The unit tests are failing.

suzannehamilton:
Thanks for pointing that out! 😳 Fixed by adding a placeholder `cognito.identitypool` value to the test application.conf.

PR #136, opened 2020-09-03
Add keycloak secret for local setup

Lholbourn:
This change adds an additional secret value needed to run keycloak locally to the README.


TomJKing (approved):
Thanks for adding this.

PR #137, opened 2020-09-03
Tdr 292 sass linter

TomJKing:


README.md
```diff
@@ -115,15 +115,19 @@ and scan files as they are uploaded to the S3 emulator.
...
+    `nvm install 14.9`
+    
 * `cd` into tdr-transfer-frontend in terminal
 
 * run  `npm install` then `npm run build`
 
 If you're working on the Sass files and just want to regenerate the CSS without rebuilding all the JavaScript code, you
 can run `npm run sass-watch` to monitor and rebuild the CSS files, or run `npm run build-css` to run a single build.
 
+To run the sass linter (stylelint) before commits, run the following command: from the npm folder `npx stylelint **/*.scss`
```
suzannehamilton, line 127: Does this run as part of the Jenkins build?

README.md
```diff
@@ -115,15 +115,19 @@ and scan files as they are uploaded to the S3 emulator.
 * If npm is not installed install [nvm](https://github.com/nvm-sh/nvm) in root directory.
 
 * Once nvm is installed:
-    `nvm install 13.6`
-
+    `nvm install 14.9`
```
suzannehamilton, line 118: This should be fine, but can you remind people to upgrade their node versions once this PR gets merged?

README.md
TomJKing: No, not at the moment. The node js version on the Jenkins Docker image would need upgrading, as currently the transfer-frontend Jenkins docker image is using node js v12.13.1. stylelint needs at least v13.6, I think.

README.md
TomJKing: Sure, will do.

README.md
suzannehamilton: I see, thanks. Are you planning to do that as part of this card?

TomJKing: I wasn't intending to. I don't think the eslint is running on build either. It might be worth having a separate card to get the linters to run on the build.

suzannehamilton: I think it should be part of this card, since it won't save reviewers time unless it runs as part of CI.

Just noticed that the card mentions running it in Jenkins: https://national-archives.atlassian.net/browse/TDR-292

If you don't have time to do it today, could you move the card back to In Progress? I'll pick it up unless someone else does.

The ESLint step is probably worth running in Jenkins too, but that can be a separate card. We didn't add it to the CI build originally because ESLint runs in a commit hook, though I noticed recently that you can bypass it.

suzannehamilton (approved):
🙌

PR #138, opened 2020-09-10
Replace upload form html with drag and drop html

techncl:


app/views/upload.scala.html
```diff
@@ -27,25 +27,38 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
-                        </span>
-                    </summary>
-                    <div class="govuk-details__text">
-                        @Messages("upload.fileExtensionSummary")
+                <div class="drag-and-drop">
+                    <div class="govuk-summary-list"></div>
+                    <div>
+                        <div class="govuk-form-group">
+                            <div class="drag-and-drop__dropzone">
+                            <input type="file" id="file-selection" name="files" class="govuk-file-upload drag-and-drop__input" webkitdirectory>
```
suzannehamilton, line 35: This line should be indented by one more tab.

npm/css-src/sass/main.scss
```diff
@@ -14,3 +14,41 @@
...
+}
+
+.drag-and-drop__dropzone label {
+  display: inline-block;
+  margin-bottom: 0;
+  width: auto;
+}
+
+.file-warning-message {
+  margin-top: 20px;
```
suzannehamilton, line 49: I'm not 100% sure of this, but the GOV.UK design system seems to follow the principle of adding margins to the bottom of elements to give the element below enough space. If you add margins to the top _and_ bottom of elements, it makes it harder to mix and match those elements in other contexts.

Is it possible to add a margin to the bottom of the drag-and-drop instead? It looks like that would be more consistent.

npm/css-src/sass/main.scss
```diff
@@ -14,3 +14,41 @@
...
+}
+
+.file-warning-text {
+  font-weight: normal;
```
suzannehamilton, line 53: What's the reason for overriding the `<strong>` weighting?

If it's to make it look like the wireframes, then I suspect we should remove this styling and make it look like the design system instead. I don't think the wireframes are using the design system directly, so they don't always look the consistent. We should go with the default from the design system unless there's a really good reason to do something else.

npm/css-src/sass/main.scss
```diff
@@ -14,3 +14,41 @@
   .hide {
   display: none;
 }
+
```
suzannehamilton, line 17: Now that the CSS is working, I think it's worth moving the drag-and-drop elements into their own file. Make a new file starting with an underscore and import it like we do for the progress bar styles.

npm/css-src/sass/main.scss
```diff
@@ -14,3 +14,41 @@
...
+.drag-and-drop {
+  margin-bottom: 40px;
+}
+
+.drag-and-drop__dropzone {
+  display: flex;
+  outline: 3px dashed #0b0c0c;
```
suzannehamilton, line 24: I think you can replace this hex value with `$govuk-input-border-colour` because the colour is in the [design system palette](https://design-system.service.gov.uk/styles/colour/).

suzannehamilton (commented):
This looks really nice! 🎉 (Both the code and the feature.)

Just a few suggestions.

suzannehamilton:
One more thought - can you run the end-to-end tests against this version, in case there are selectors that need to be merged at same time to keep them passing?

app/views/upload.scala.html
techncl: Ah, yes; thanks!

npm/css-src/sass/main.scss
techncl: Good point! Sure!

npm/css-src/sass/main.scss
techncl: Yes, it was to look like the wireframes. Sure, I'll change it back.

npm/css-src/sass/main.scss
techncl: Thanks! I'll bookmark that link too.

techncl:
> One more thought - can you run the end-to-end tests against this version, in case there are selectors that need to be merged at same time to keep them passing?

Sure! I ran them and they passed. probably because I kept the css classes and ids for the `<input>` element, the same as the previous ones and the e2e tests seem to only make use of that and the "continue" button.

app/views/upload.scala.html
```diff
@@ -27,25 +27,38 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+                <div class="drag-and-drop">
+                    <div class="govuk-summary-list"></div>
+                    <div>
+                        <div class="govuk-form-group">
+                            <div class="drag-and-drop__dropzone">
+                                <input type="file" id="file-selection" name="files" class="govuk-file-upload drag-and-drop__input" webkitdirectory>
+                                <p class="govuk-body">@Messages("upload.dragAndDropHintText")</p>
+                                <label for="file-selection" class="govuk-button govuk-button--secondary">@Messages("upload.chooseFolderLink")</label>
+                            </div>
+                            <div class="govuk-warning-text file-warning-message">
```
suzannehamilton, line 39: Can the `file-warning-message` class be removed now there's no CSS referencing it? Same goes for `file-warning-text` below.

suzannehamilton (commented):
One tiny comment. Otherwise this looks good!

app/views/upload.scala.html
techncl: Sure! I can't remember why this was added anymore :smile:

suzannehamilton (approved):
🎉

PR #139, opened 2020-09-11
Tdr 358 update progress bars in js

MancunianSam:


npm/test/filecheckprocessing.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  mutation: (
+    query: DocumentNode,
+    variables: GetFileCheckProgressQueryVariables
+  ) => Promise<FetchResult<GetFileCheckProgressQuery>> = (_, __) => {
+    return Promise.reject("error")
+  }
+}
+
+test.each`
+  processed | total   | expected
```
Lholbourn, line 63: This is really cool! Much easier to see all the expected/controlled data :open_mouth:

npm/test/filecheckprocessing.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  getConsignmentData(client, callback)
+})
+
+test("getConsignmentId returns the correct id when the hidden input is present", () => {
+  const consignmentId = "91847ae4-3bfd-4fac-b448-68948dd95820"
+  document.body.innerHTML = `<input id="consignmentId" type="hidden" value="${consignmentId}">`
+  expect(getConsignmentId()).toBe(consignmentId)
+})
+
+test("getConsignmentId returns the correct id when the hidden input is absent", () => {
```
Lholbourn, line 137: Should this be about returning an error rather than the correct id?

npm/src/filechecks/index.ts
```diff
@@ -0,0 +1,46 @@
...
+        }/consignment/${getConsignmentId()}/records-results`
+        window.location.href = location
+      }
+    }
+  }
+
+  updateFileCheckProgress() {
+    const interval = setInterval(() => {
+      getConsignmentData(this.client, this.dataCallback)
+    }, 2000)
```
Lholbourn, line 44: I ran this locally and was thinking is it possible for us to update the processing bars more often than every 2 seconds so it moves similarly to the upload bar? Though this could just be because I tested it with quite a small folder!

Lholbourn (changes requested):
Looks really cool & all very readable! I left a couple of comments on things I had questions about :+1:

npm/test/filecheckprocessing.test.ts
MancunianSam: Yep, good spot. Looks like a bad copy paste job. I've changed it.

npm/src/filechecks/index.ts
MancunianSam: It's hard to tell what's best. The upload bar updates when the file checks are done or the s3 upload completes rather than being run on a timer like this one so it's difficult to compare. We have to find the right balance between not being too slow on the page and not hammering the api with requests. I think 2 seconds is ok for now and we can change it if the user testing shows it needs to be longer or shorter.

npm/test/filechecks.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  await fileChecks.updateFileCheckProgress()
+  jest.runOnlyPendingTimers()
+  expect(setInterval).toBeCalledTimes(1)
+})
+
+test("updateFileCheckProgress updates the progress bars correctly", () => {
+  mockConsignmentData({
+    antivirusProcessed: 1,
+    checksumProcessed: 2,
+    totalFiles: 1
```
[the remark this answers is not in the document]
MancunianSam: Good point. It sort of implies that we'll get data like this when we never should in practice. I've changed it.

npm/src/filechecks/index.ts
Lholbourn: Good point! :+1:

npm/test/filechecks.test.ts
```diff
@@ -41,7 +41,7 @@ test("updateFileCheckProgress updates the progress bars correctly", () => {
   mockConsignmentData({
     antivirusProcessed: 1,
     checksumProcessed: 2,
-    totalFiles: 1
+    totalFiles: 2
   })
   fileChecks.updateFileCheckProgress()
   jest.runOnlyPendingTimers()
```
Lholbourn, line 47: This test looks like it's failing on Jenkins, I think the expected data may also need changed

MancunianSam: Yep, my fault. I've updated the expected result.

Lholbourn (approved):
Looks good to me! :rocket:

PR #140, opened 2020-09-11
Add instructions to fix cognito error to readme

Lholbourn:
This includes adding a screenshot of the error a user may see to the public folder. This was chosen over an image hosting website so it is more reliable if viewing offline or when on a local copy of the project.


README.md
```diff
@@ -8,6 +8,11 @@ There are two ways to develop this project:
 - Frontend development only, using the AWS integration environment for everything else. This is the default
 - Full stack local development, using a local dev copy of the API, Keycloak, etc
 
+
+- When moving between full local and frontend only development you should clear your browsers' cookies and local storage as the browser can cache the local Cognito ID, which is incompatible with non-local Cognito. An example of the error you will see if this happens is below:
+
+![Cognito error example](public/images/Cognito-error-example-screenshot.png)
```
suzannehamilton, line 14: Could you include the error as text rather than an image? There a couple of reasons for this: if people search for the error text, it would be useful for this to show up as a search result, and putting an image in `public/images` means that it gets deployed with the Play site! 🙃

Lholbourn: Ah of course!

README.md
```diff
@@ -8,6 +8,15 @@ There are two ways to develop this project:
 - Frontend development only, using the AWS integration environment for everything else. This is the default
 - Full stack local development, using a local dev copy of the API, Keycloak, etc
 
+
+- When moving between full local and frontend only development you should clear your browsers' cookies and local storage as the browser can cache the local Cognito ID, which is incompatible with non-local Cognito. An example of the error you will see if this happens is below:
```
suzannehamilton, line 12: Sorry, just noticed that this looks like it's in the list of ways to develop the project. Maybe just remove the `-` from the start of this comment?

Lholbourn: Sure will do! I wasn't 100% sure on where to put this bit but I thought somewhere near this section was a safe bet.

suzannehamilton (approved):

suzannehamilton (dismissed):

suzannehamilton (approved):

PR #175, opened 2020-10-05
Add drag and drop functionality and animation

techncl:


npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,53 @@
   margin-bottom: 40px;
 }
 
+.drag-and-drop__button {
+  margin-bottom: 32px;
+  margin-top: 65px;
+  padding-left: 15px;
+  position: static;
```
suzannehamilton, line 9: This might be my CSS ignorance, but why is this line needed? I thought `static` was the default?

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,53 @@
...
+.drag-and-drop__button--highlight {
+  background-color: $govuk-focus-colour;
+  box-shadow: 0 -2px $govuk-focus-colour, 0 4px govuk-colour('black');
+  color: $govuk-text-colour;
+  outline: none;
+}
+
+.drag-and-drop--dragover {
+  background: #b1b4b6;
+  outline-color: #6f777b;
```
suzannehamilton, line 26: I just tried this in Firefox, and the outline doesn't seem to change colour for me. It stays black rather than switching to grey.

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,53 @@
...
+
+.drag-and-drop__button--highlight {
+  background-color: $govuk-focus-colour;
+  box-shadow: 0 -2px $govuk-focus-colour, 0 4px govuk-colour('black');
+  color: $govuk-text-colour;
+  outline: none;
+}
+
+.drag-and-drop--dragover {
+  background: #b1b4b6;
```
suzannehamilton, line 25: This colour is in the [palette](https://design-system.service.gov.uk/styles/colour/). The outline colour isn't - is there a similar colour that we could use instead?

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,53 @@
...
+}
+
+.drag-and-drop__button--highlight {
+  background-color: $govuk-focus-colour;
+  box-shadow: 0 -2px $govuk-focus-colour, 0 4px govuk-colour('black');
+  color: $govuk-text-colour;
+  outline: none;
+}
+
+.drag-and-drop--dragover {
```
suzannehamilton, line 24: It looks like this class goes on the `.drag-and-drop__dropzone` element, which means it should be `.drag-and-drop__dropzone--dragover` to be consistent with BEM.

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,53 @@
...
-.drag-and-drop__button {
-  display: inline-block;
-  margin-bottom: 0;
-  width: auto;
+.drag-and-drop__input {
+  height: 166px;
+  opacity: 0;
+  padding-top: 0;
+  position: absolute;
+  width: 630px;
```
suzannehamilton, line 51: I suspect that someone looking at these rules for the first time won't be able to work out the reason for some of these styling rules. It's probably quite surprising to see an input element which has such a large height and width, and which has zero opacity. Could you add a comment explaining why we're taking this approach?

suzannehamilton (commented):
Looks good! A few questions and suggestions on the CSS.

npm/css-src/sass/_drag-and-drop.scss
techncl: Yes, `static` is the default but for GOV.UK buttons, it is set to `relative` and that (for some reason) was causing the folder to open in a new tab when you dropped one on it; `static` was one of the only ones that didn't cause this behaviour.

npm/css-src/sass/_drag-and-drop.scss
techncl: Ah, yes! Sure, will do!

npm/css-src/sass/_drag-and-drop.scss
techncl: Cool, I'll use this `govuk-colour("dark-grey")`; it's probably the closest.

npm/css-src/sass/_drag-and-drop.scss
techncl: Ah, yes; good spot! I changed the order of the CSS so the black one was overwriting the grey. I've now fixed it and checked it in Firefox and Chrome and they look fine now. Thanks.

npm/css-src/sass/_drag-and-drop.scss
techncl: Sure, will do!

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -2,31 +2,55 @@
...
 
-.drag-and-drop__button {
-  display: inline-block;
-  margin-bottom: 0;
-  width: auto;
+.drag-and-drop__input {
+  /*Since input fields allow you to drag and drop folders, we've decided to make
+   the input as large as the dropzone box to avoid dealing with drop events*/
+  height: 166px;
+  opacity: 0; /*Opacity is 0 in order to keep hide the input button AND keep it functioning ('visibility' and 'display' properties can't do this)*/
```
suzannehamilton, line 50: Should "keep hide the input button" be "keep the input button hidden"?

suzannehamilton (commented):
Just one rewording comment. 👍

npm/css-src/sass/_drag-and-drop.scss
techncl: Ah :man_facepalming: . Good spot! I think I restructured my sentence and forgot to delete the word "keep". It was meant to be "in order to hide the input button". Thanks, I'll fix that.

suzannehamilton (approved):

PR #200, opened 2020-10-12
Detect non-folders/empty folders and display red error message

techncl:


app/views/upload.scala.html
```diff
@@ -29,15 +37,19 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+                        <div class="govuk-summary-list__row">
+                            <dd class="govuk-summary-list__value drag-and-drop__success hide">
                                 @greenTickMark()
                                 The folder
                                 "<span id="folder-name"></span>"
                                 (containing
                                 <span id="folder-size"></span>)
                                 has been selected
                             </dd>
+                            <dd class="govuk-summary-list__value drag-and-drop__failure hide">
```
TomJKing, line 49: When I test with an empty folder I get the error message: "You can only drop a single folder". Is that correct? I would have thought we need a different error message depending on the scenario, ie:

- User tries to drag and drop a non-folder: "You can only drop a single folder"
- User tries to drag and drop empty folder: "You cannot drop an empty folder"

Has the wording of the error message(s) been confirmed with Sarri, Rooha and Kirsten?

npm/test/folderinforetrieval.test.ts
```diff
@@ -67,7 +89,7 @@ test("folder retriever updates the page with correct folder information if there
   triggerInputEvent(folderRetriever!)
 
   const folderSummaryElement: HTMLElement | null = document.querySelector(
-    ".govuk-summary-list__row"
+    ".drag-and-drop__success"
```
TomJKing, line 92: We should add some additional unit test to make sure the page state is correct when:

- a user drags and drops an empty folder: `test("folder retriever updates the page with correct folder information if there are no files", () => { ... etc }`
- a user drags and drops a non-folder object:  `test("folder retriever updates the page with correct folder information if there are is a non file object", () => { ... etc }`

Not sure such test are possible, but they should be included as unit tests as they form part of the logic for the behaviour.

TomJKing (commented):
Looks good. Add some comments, let me know if you want to discuss.

Also not sure we have a JIRA ticket/thought about where a user tries to drag and drop more than one folder object? Do we need to display an error message for this scenario as well?

npm/test/folderinforetrieval.test.ts
```diff
@@ -67,7 +89,7 @@ test("folder retriever updates the page with correct folder information if there
   triggerInputEvent(folderRetriever!)
 
   const folderSummaryElement: HTMLElement | null = document.querySelector(
```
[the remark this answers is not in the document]
techncl: Sure! Thanks

app/views/upload.scala.html
techncl: This is something I brought up too. I think this will be decided by Sarra, Rooha and Kirsten after the user testing.

npm/test/folderinforetrieval.test.ts
techncl: Yes. I was trying to remember why I left out the failure test, but now I do. It was because when there is an error in the actual code, it throws an exception which doesn't allow the test to complete; I tried a few things but could not get past the exception. This was something that Suzanne said she was going to have have a think about but we should leave it for now.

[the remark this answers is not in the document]
techncl: Ah, yes. Thanks!

app/views/upload.scala.html
```diff
@@ -29,15 +37,19 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+                                @redWarningSign()
+                                You can only drop a single folder
```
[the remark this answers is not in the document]
techncl: Sure!

app/views/upload.scala.html
TomJKing: Yes, thanks for doing that.

npm/test/folderinforetrieval.test.ts
TomJKing: OK.

TomJKing (approved):
Looks good.

PR #215, opened 2020-10-20
Error pages UI-front-end

rsulxo:
Added the error pages in the `view` folder. 

Using content from wireframes.

Card for full story: https://national-archives.atlassian.net/browse/TDR-177?atlOrigin=eyJpIjoiNGM5YjZlOTVjMjBiNDU4NWJiZmFjNDNjNDIwOGQ1ZWMiLCJwIjoiaiJ9


app/views/forbiddenError.scala.html
```diff
@@ -0,0 +1,16 @@
+@*
+* This is a placeholder to allow us to write unit tests which return an error page.
+* It will be replaced once we have a properly designed error page
+*@
```
suzannehamilton, line 4: Is this comment out-of-date? This looks like it is a properly designed error page. 🙂 

Same goes for the other pages below.

app/views/notFoundError.scala.html
```diff
@@ -8,15 +8,13 @@
...
-                <h1 class="govuk-heading-xl">404 - Page not found</h1>
-                <p>This is a placeholder 404 error page</p>
-                <span id="general-error" class="govuk-error-message">
-                    @errorMessage
-                </span>
+                <h1 class="govuk-heading-l">Page not found</h1>
+                <p class="govuk-body">If you typed a web address, check it is correct.</p>
+                <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+                <p class="govuk-body">
+                  If the web address is correct or you selected a link or button, <a class="govuk-link" href="#">contact The National Archives</a> if you need to speak to someone about your transfers.
```
suzannehamilton, line 15: Where should this link go to? It currently just goes to the same page.

package-lock.json
```diff
@@ -0,0 +1,2795 @@
+{
```
suzannehamilton, line 1: Was this file added by mistake? There shouldn't be a package-lock.json file in the root directory.

app/views/unauthenticatedError.scala.html
```diff
@@ -0,0 +1,16 @@
...
+* It will be replaced once we have a properly designed error page
+*@
+@(errorMessage: String)(implicit messages: Messages)
+
+@main(Messages("error.header")) {
+    @defining(play.core.PlayVersion.current) { version =>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h1 class="govuk-heading-l">You must be logged in to see this page</h1>
+                <p class="govuk-body"> Please <a class="govuk-link" href="#">login</a> to continue.</p>
```
suzannehamilton, line 12: I've updated the acceptance criteria for [TDR-583](https://national-archives.atlassian.net/browse/TDR-583) to remind us to make this link work.

app/views/forbiddenError.scala.html
```diff
@@ -0,0 +1,16 @@
...
+@(errorMessage: String)(implicit messages: Messages)
+
+@main(Messages("error.header")) {
+    @defining(play.core.PlayVersion.current) { version =>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h1 class="govuk-heading-l">You are not permitted to see this page</h1>
+                <p class="govuk-body"> This page is restricted to certain users only.</p>
```
suzannehamilton, line 12: Really minor, but there's a space after the `<p>` tag. Same for the other error pages below.

app/views/forbiddenError.scala.html
```diff
@@ -0,0 +1,16 @@
...
+* This is a placeholder to allow us to write unit tests which return an error page.
+* It will be replaced once we have a properly designed error page
+*@
+@(errorMessage: String)(implicit messages: Messages)
+
+@main(Messages("error.header")) {
+    @defining(play.core.PlayVersion.current) { version =>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h1 class="govuk-heading-l">You are not permitted to see this page</h1>
```
suzannehamilton, line 11: Do you want a hand with moving this text into the messages file? @techncl might be a good person to pair with, since he's done this quite a lot recently.

suzannehamilton:
Just noticed that the Jenkins build is failing - it looks like some tests need to be updated because the wording on the error pages has changed.

app/views/notFoundError.scala.html
rsulxo: I would imagine this would go to Claire's email address as she deals with transfers, but if we want this to link to a TDR help desk we could also consider that. I wasn't too sure about this when I was designing it too.

suzannehamilton:
Looks like the Jenkins build is still failing. Could you take a look and fix the tests?

app/views/notFoundError.scala.html
suzannehamilton: Could you find out where it should go to and update it?

rsulxo:
The Jenkins test now passes, turns out the text on here needed to match the text on the error page itself. I’ve changed that now. So hopefully there shouldn’t be any further issues.

app/views/notFoundError.scala.html
rsulxo: I spoke with Kirsten and Sarra regarding this and they suggested that we have two links on this page, one to let the user return back to the home page and the other to let the user contact a system admin/transfer advisor, essentially it should go to a “TDR help desk”.

At the moment we haven’t got that set up so we’ll need to think about where this link would go for the time being.

conf/messages
```diff
@@ -55,6 +55,17 @@ checkingRecords.ffid=Identifying file formats
...
+forbiddenError.body=This page is restricted to certain users only.
+internalServerError.header=Sorry, there is a problem with the service
+internalServerError.body=Try again later.
+unauthenticatedError.header=You must be logged in to see this page
+unauthenticatedError.body=Please login to continue.
+notFoundError.header=This page doesn't exist
+notFoundError.body1=If you typed a web address, check it is correct.
+notFoundError.body2=If you pasted the web address, check you copied the entire address.
+notFoundError.body3=If the web address is correct or you selected a link or button, contact The National Archives Help Desk to speak to someone about your transfers.
+notFoundError.body4=Alternatively, you can go back to start page
```
suzannehamilton, line 68: The wording doesn't seem quite right here: "go back to start page". Also , should it be a link?

app/views/notFoundError.scala.html
suzannehamilton: Thanks for checking with them. Could you create a card so that we can update the contact link when we've created the mailing list?

rsulxo: No worries, I've added that here: https://national-archives.atlassian.net/browse/TDR-601

conf/messages
rsulxo: Kirsten and Sarra mentioned that as a non-technical user if they were on this page, they wouldn't know how to go back to the home page, so having a link as such would make it a lot easier for the user to navigate. Also, I think the term 'go back' sounds a bit informal, so I have changed it to 'return to home page' as that feels more familiar to the users. What do you think?

suzannehamilton: The new wording looks OK to me, thanks. I don't know if "return" or "go back" is preferred. I just meant that "go back to start page" wasn't grammatical. 🙂 

What about the link?

rsulxo: No worries. 

For the link I included the `<a>` tags around the relevant text in the `messages` file, and then in the HTML file where I call the text, I used:  `@Html(Messages("notFoundError.body3"))` to ensure the link works. I know it may not be the best solution but this was something that seemed to work cleaner than any other options.

rsulxo:
Just a little side note: To make visualisation easier, I also made placeholder controller pages for the error pages.

conf/messages
suzannehamilton: Yeah, I'm actually not sure how you're supposed to handle things like links within the messages file. I think this is fine for now.

It looks like the link doesn't go anywhere, though, because it links to `#`. Where is the link supposed to go, and could you update it to go there?

rsulxo: That link is supposed to go to the dashboard page, as if the user sees an error not found, they can return to the home page (which is our dashboard) where they can either start a new transfer/resume a transfer.

conf/messages
```diff
@@ -55,6 +55,17 @@ checkingRecords.ffid=Identifying file formats
...
+forbiddenError.body=This page is restricted to certain users only.
+internalServerError.header=Sorry, there is a problem with the service
+internalServerError.body=Try again later.
+unauthenticatedError.header=You must be logged in to see this page
+unauthenticatedError.body=Please <a href= "#">login</a> to continue.
+notFoundError.header=This page doesn't exist
+notFoundError.body1=If you typed a web address, check it is correct.
+notFoundError.body2=If you pasted the web address, check you copied the entire address.
+notFoundError.body3=If the web address is correct or you selected a link or button, contact <a href= "#">The National Archives Help Desk<a/> to speak to someone about your transfers.
+notFoundError.body4=Alternatively, you can <a href= "/dashboard">return</a> to the home page.
```
suzannehamilton, line 68: Should the link be on the whole "return to the home page" phrase rather than just "return"? This is good for accessibility reasons (see https://www.washington.edu/accessibility/links/ for example) and I think if I was the user, I'd expect to be able to click anywhere in that phrase.

rsulxo: Oh true! I've changed that now.

app/controllers/FileNotFoundController.scala
```diff
@@ -0,0 +1,18 @@
...
+import java.util.UUID
+
+import auth.TokenSecurity
+import configuration.KeycloakConfiguration
+import javax.inject.Inject
+import org.pac4j.play.scala.SecurityComponents
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, Request}
+
+class FileNotFoundController @Inject()(val controllerComponents: SecurityComponents,
```
suzannehamilton, line 12: I've added notes to the remaining subtasks to remind us to remove these controllers and routes once we've connected the error pages to the Play error handling. 👍

suzannehamilton (approved):
Thanks for making those changes!

app/controllers/FileNotFoundController.scala
rsulxo: Great, thanks! :)

PR #216, opened 2020-10-20
Add placeholder transfer confirmation page

techncl:
Wasn't sure if the `<caption>` and `<thead>` should be kept in the table and left blank, removed or filled in with something. [This page](https://design-system.service.gov.uk/components/table/ )  utilises them both; not sure if they apply to our table.

### DIFF
--- a/app/views/transferConfirmation.scala.html
+++ b/app/views/transferConfirmation.scala.html
@@ -1,0 +1,53 @@
+@()(implicit messages: Messages)
+@main(Messages("transferConfirmation.header")) {
+    @defining(play.core.PlayVersion.current) { version =>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>
+                @progressIndicator(Messages("transferConfirmation.progress"))
+                <h1 class="govuk-heading-xl">@Messages("transferConfirmation.title")</h1>
+                <table class="govuk-table">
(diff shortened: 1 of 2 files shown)


app/views/transferConfirmation.scala.html
```diff
@@ -0,0 +1,53 @@
+@()(implicit messages: Messages)
```
suzannehamilton, line 1: How do you get to this page? Was there already a controller and route pointing to it?

suzannehamilton:
Probably best to ask the question about the caption and table in Slack, so that Kirsten and Rooha can join the discussion.

app/views/transferConfirmation.scala.html
techncl: When testing this page, I was just changing the dashboard controller from `Ok(views.html.dashboard())` to  `Ok(views.html.transferConfirmation())`. Since the Jira task was to just "Placeholder page for transfer confirmation page", I wasn't sure if I should add more functionality to it since there is another card that is "Add Transfer Summary page in Play app". Should I also add a route for it and create a controller for it?
According to the wireframes, this page is the 8th page, but we've only built 5 (I think Loki is working on the 6th and the 7th is the metadata page), so there wouldn't be a direct page to it for now, you'd have to type the url in (if I build the controller and add a route for it).

suzannehamilton: Yeah, if we're adding a page then we should have a route and controller to navigate to it so that it's easy to link to from the other task, and so that we can add more features to it. 🙂

PR #219, opened 2020-10-21
Tdr 509 save parent folder

Lholbourn:
This change saves the parent folder name to the consignment at the same stage that we are updating the files table. As a function already existed to extract the parent folder name from the files, I used this and passed the variable through from the `upload-form.ts`, to `upload/index.ts`, `clientFileProcessing.ts` to the `clientFileMetadataUpload.ts` file where we have a function that saves files with the appropriate API endpoint. This saves duplicating an existing piece of work.

I also changed some of the existing tests to reflect the change in these functions and their parameters.

This work changes the way we save files (and adds the saving of parent folder name) but does not affect the upload itself.

### DIFF
--- a/npm/src/upload/upload-form.ts
+++ b/npm/src/upload/upload-form.ts
@@ -99,13 +99,18 @@ export class UploadForm {
   }
 
   addSubmitListener(
-    uploadFiles: (files: TdrFile[], consignmentId: string) => void
+    uploadFiles: (
+      files: TdrFile[],
+      consignmentId: string,
+      parentFolder: string
+    ) => void
   ) {
     this.formElement.addEventListener("submit", (ev) => {
       ev.preventDefault()
       const target: HTMLInputTarget | null = ev.currentTarget
       const files = this.retrieveFiles(target)
-      uploadFiles(files, this.consignmentId())
+      const parentFolder = this.getParentFolderName(files)
+      uploadFiles(files, this.consignmentId(), parentFolder)
     })
   }
 
--- a/npm/src/upload/upload-form.ts
+++ b/npm/src/upload/upload-form.ts
@@ -4,6 +4,11 @@ export interface InputElement extends EventTarget {
   files?: TdrFile[]
 }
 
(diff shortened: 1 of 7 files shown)


npm/src/upload/upload-form.ts
```diff
@@ -99,13 +99,18 @@ export class UploadForm {
   }
```
TomJKing, line 99: It might be better to add the consignmentId and parentFolder values to an object, and pass this through to the other functions, instead of passing them as separate params:
1. This will reduce the number of params being passed to functions;
2. If for example in the future we allow users to upload more than one folder at a time, just need to change the object to accommodate this
3. If we need to pass through other information taken from the upload event in the future, just need to change the object, by adding a new fields for the other information.

Would suggest a new interface in this file (not sure about the name):

```
export interface IUploadFileInfo {
  consignmentId: string;  
  parentFolder: string;  
}
```
This can then be passed around, and referenced by the other functions as needed, for example:

```
...
addSubmitListener(
    uploadFiles: (
      files: TdrFile[],
      uploadFileInfo:  IUploadFileInfo
    ) => void
  ) {
    this.formElement.addEventListener("submit", (ev) => {
      ev.preventDefault()
      const target: HTMLInputTarget | null = ev.currentTarget
      const files = this.retrieveFiles(target)
      const parentFolder = this.getParentFolderName(files)
      const uploadFileInfo: IUploadFileInfo = {
        consignmentId: this.consignmentId(),
        parentFolder: parentFolder
      }

      uploadFiles(files, uploadFileInfo)
    })
  }
...
```

TomJKing (commented):
Looks good. Made one suggestion, see the comments.

npm/src/upload/upload-form.ts
Lholbourn: That makes much more sense yeah! I'll change it to do that :+1:

TomJKing (dismissed):

TomJKing (approved):

PR #225, opened 2020-10-23
Change README auth secret instructions to get correct secret

Lholbourn:
Previously the readme set the AUTH_SECRET as `intg/auth/secret` but it should be `intg/keycloak/client/secret`. This change corrects this to ensure running FE locally works correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,12 @@ development environment for the other TDR services.
     - In the AWS console:
       - Go the Systems Manager service
       - Go the Parameter Store in the left-hand menu
-      - Find the `/intg/auth/secret` parameter
+      - Find the `/intg/keycloak/client/secret` parameter
       - Copy the parameter's value
     - With the AWS CLI:
       - Run:
         ```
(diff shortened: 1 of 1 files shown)


suzannehamilton (approved):

suzannehamilton:
I've just noticed there's a Notes section at the bottom of the Readme which I think is out of date too: https://github.com/nationalarchives/tdr-transfer-frontend/blob/f081346a70c67dcbf13a00a008bc4fb334eba644/README.md#notes

When you check with Sam whether we can delete `/intg/auth/secret`, can you check if we can delete those Notes too?

Lholbourn:
> I've just noticed there's a Notes section at the bottom of the Readme which I think is out of date too: https://github.com/nationalarchives/tdr-transfer-frontend/blob/f081346a70c67dcbf13a00a008bc4fb334eba644/README.md#notes
> 
> When you check with Sam whether we can delete `/intg/auth/secret`, can you check if we can delete those Notes too?

Oh cool, will do! :smile:

PR #246, opened 2020-11-02
Tdr 511 show processing success

Lholbourn:
These changes ensure that we can provide a UI for users to see that their initial upload and backend checks are complete and that they can now choose to move on to the next stage of the transfer. It has a description of the number of files uploaded and the parent folder name to help the user ensure the correct data has been uploaded. 

I created a function in the `consignmentService` to retrieve the consignment information from the API, and the `fileChecksResultsController` uses this to create the correct interface for the user. I also created unit tests for the new functionality. The UI was created using the wireframes provided.


conf/messages
```diff
@@ -53,8 +53,10 @@ checkingRecords.antivirus=Anti-virus scanning
 checkingRecords.checksum=Validating data integrity
 checkingRecords.ffid=Identifying file formats
 fileChecksResults.header=Record check results
-fileChecksResults.title=Record check results
-fileChecksResults.description=Here are your results
+fileChecksResults.title=Checks Completed!
+fileChecksResults.description=Your folder ''{0}'' containing ''{1} items'' has been successfully checked and uploaded. Please click ''Continue'' to proceed with your transfer.
```
suzannehamilton, line 57: This might be tricky, but can "items" be singular when there is only 1 item?

Also do we need the quotes around `{1} items`?

conf/routes
```diff
@@ -12,13 +12,13 @@ GET     /consignment/:consignmentId/transfer-agreement     controllers.TransferA
...
-GET     /consignment/:consignmentId/records-results        controllers.FileChecksResultsController.results(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/transfer-summary       controllers.TransferSummaryController.transferSummary(consignmentId: java.util.UUID)
 POST    /consignment/:consignmentId/transfer-summary       controllers.TransferSummaryController.transferSummarySubmit(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/notFoundError          controllers.FileNotFoundController.notFoundError(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/unauthenticatedError   controllers.UnauthenticatedErrorController.unauthenticatedError(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/forbiddenError         controllers.ForbiddenErrorController.forbiddenError(consignmentId: java.util.UUID)
 GET     /consignment/:consignmentId/internalServerError    controllers.InternalServerErrorController.internalServerError(consignmentId: java.util.UUID)
+GET     /consignment/:consignmentId/records-results        controllers.FileChecksResultsController.fileCheckResultsPage(consignmentId: java.util.UUID)
```
suzannehamilton, line 21: Was this line moved intentionally? I'd suggest either grouping the routes file by similar routes, or ordering them alphabetically.

test/controllers/FileChecksResultsControllerSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+      contentType(recordCheckResultsPage) mustBe Some("text/html")
+      resultsPageAsString must include("fileChecksResults.header")
+      resultsPageAsString must include("fileChecksResults.title")
+      resultsPageAsString must include("govuk-panel--confirmation")
+      resultsPageAsString must include("fileChecksResults.description")
+      resultsPageAsString must include("fileChecksResults.dashboard")
+      resultsPageAsString must include("fileChecksResults.continueLink")
+    }
+
+    "return a redirect to the auth server with an unauthenticated user" in {
```
suzannehamilton, line 64: It would be good to add a test for an authenticated user that does not own the consignment. I don't know if there are examples of this kind of thing already.

suzannehamilton (commented):
Looks good! Just a few small comments.

conf/messages
Lholbourn: Ah I didn't think about that case! I'll have a look and see if it can be changed when it's a singular. The quotation marks were around the both the number and the word items in the wireframes, but I do think it looks a bit strange. I can ask Rooha about it when she's back?

conf/routes
Lholbourn: Oh strange, I think this happened when I was fixing merge conflicts. It is meant to be where it was in the last commit, I'll change that back!

test/controllers/FileChecksResultsControllerSpec.scala
Lholbourn: I'll give it a go :+1:

suzannehamilton (approved):
Looks good to me!

PR #263, opened 2020-11-12
Add internal server error

TomJKing:
Display internal server error HTML

Remove temporary internal server error controller and route as no longer needed


app/errors/ErrorHandler.scala
```diff
@@ -26,7 +26,7 @@ class ErrorHandler @Inject() (val messagesApi: MessagesApi) extends HttpErrorHan
       case authException: AuthorisationException =>
         Forbidden(views.html.error(s"Not authorised: ${authException.getMessage}")(request2Messages(request)))
       case e =>
-        InternalServerError(views.html.error(s"A server error occurred: ${e.getMessage}")(request2Messages(request)))
+        InternalServerError(views.html.internalServerError(s"${e.getMessage}")(request2Messages(request)))
```
suzannehamilton, line 29: Does the `internalServerError` page need the message parameter? I don't think we want to print the error message on the page. Logging it is enough.

suzannehamilton (approved):

PR #264, opened 2020-11-12
Tdr 585 404 error page

TomJKing:


app/errors/ErrorHandler.scala
```diff
@@ -14,9 +14,12 @@ class ErrorHandler @Inject() (val messagesApi: MessagesApi) extends HttpErrorHan
   override def onClientError(request: RequestHeader, statusCode: Int, message: String): Future[Result] = {
     logger.error(s"Client error with status code $statusCode at path '${request.path}' with message: '$message'")
 
-    Future.successful(
-      Status(statusCode)("A client error occurred: " + message)
-    )
+    val response = statusCode match {
+      case 404 =>
```
suzannehamilton, line 18: Do we need a default case here? Otherwise you'll get a match error instead of the expected 400 or 403 or whatever it is.

TomJKing: What should the default behaviour be? Should it default to showing the views.html.error(), with a generic error message eg "A client error occurred: $message"

suzannehamilton: I'm not sure... Does this method get run in cases like 400 response from invalid form data? For example, if you submit a blank transfer agreement form, does it hit this code?

If not, I think the best thing to do might be for this match statement to show the appropriate view for every status code which has a specific error page (401, 403, 404), and return the generic server error page in every other case, preserving the status code. So the user would just see a branded "there is a problem with the service" message but our network logs, etc. would capture the real status code to help us debug the problem. "Wait, why is the frontend returning a 409 Conflict response?"

TomJKing: _Does this method get run in cases like 400 response from invalid form data?_ It is not run for form data issues, as this is handled by the individual form controllers. Passing feedback back to the invalid form .

suzannehamilton (approved):
Looks good, thanks.

PR #275, opened 2020-11-18
trigger consignment export

MancunianSam:


app/controllers/TransferSummaryController.scala
```diff
@@ -38,9 +40,11 @@ class TransferSummaryController @Inject()(val controllerComponents: SecurityComp
       Future.successful(BadRequest(views.html.transferSummary(consignmentId, formWithErrors)))
     }
 
-    val successFunction: TransferSummaryData => Future[Result] = { formData: TransferSummaryData =>
-      Future(Ok(views.html.transferConfirmation()))
-      //Code here will be replaced with a mutation to database
+    val successFunction: TransferSummaryData => Future[Result] = { _: TransferSummaryData =>
+      for {
+        exportTriggered <- consignmentExportService.triggerExport(consignmentId, request.token.bearerAccessToken.toString)
```
suzannehamilton, line 45: Should this action also call the API to mark the transfer as initiated?

MancunianSam: Thanks :smile: I missed out a major part of this ticket, sorry. I've marked the PR as draft and I'll add it in.

app/views/transferConfirmation.scala.html
```diff
@@ -1,12 +1,24 @@
...
+                    @progressIndicator(Messages("transferConfirmation.progress"))
+                    <p>Export complete placeholder page</p>
+                </div>
             </div>
-        </div>
+        } else {
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-two-thirds">
+                    <span id="general-error" class="govuk-error-message">
+                        The export failed, contact us
```
suzannehamilton, line 16: Is there a card for returning to this page and handling both conditions?

app/services/ConsignmentExportService.scala
```diff
@@ -0,0 +1,33 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class ConsignmentExportService @Inject()(val ws: WSClient, val configuration: Configuration, graphQLConfiguration: GraphQLConfiguration)
+                                        (implicit val executionContext: ExecutionContext) {
+
+  def updateTransferInititated(consignmentId: UUID, token: BearerAccessToken): Future[Boolean] = {
+    val client = graphQLConfiguration.getClient[Data, Variables]()
+    sendApiRequest(client, document, token, Variables(consignmentId))
+      .map(d => d.updateTransferInitiated.isDefined)
+      .recover(_ => false)
```
suzannehamilton, line 22: Can you make sure that any exceptions from failed futures and GraphQL errors are logged?

app/services/ConsignmentExportService.scala
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  def triggerExport(consignmentId: UUID, token: String): Future[Boolean] = {
+    val url = s"${configuration.get[String]("export.baseUrl")}/export/$consignmentId"
+    ws.url(url)
+      .addHttpHeaders(("Authorization", token), ("Content-Type", "application/json"))
+      .post("{}")
+      .map(r => r.status == 200)
+      .recover(_ => false)
```
suzannehamilton, line 31: Ditto here for any non-200 statues or failed futures.

app/views/transferConfirmation.scala.html
MancunianSam: I couldn't see one so I've added in TDR-672

app/services/ConsignmentExportService.scala
MancunianSam: The sendApiRequest method checks for errors from the GraphQL response and throws them as exceptions if there are any so any errors at all will be contained within the failed future. I've added a logging line into the recover block to log the error and this should log graphql errors and any network errors.

app/services/ConsignmentExportService.scala
MancunianSam: I've added in a log line for non 200 statuses which logs the status code and the response body as API gateway seems to give you nice errors in the response body. 
For the recover block, I've logged any exception that we get in there.

suzannehamilton:
Looks like you've got a test failure.

MancunianSam:
I updated the view to use the messages file and didn't update the test that was checking for the text. It should be ok now :crossed_fingers:

app/services/ConsignmentExportService.scala
```diff
@@ -0,0 +1,46 @@
...
+  def triggerExport(consignmentId: UUID, token: String): Future[Boolean] = {
+    val url = s"${configuration.get[String]("export.baseUrl")}/export/$consignmentId"
+    ws.url(url)
+      .addHttpHeaders(("Authorization", token), ("Content-Type", "application/json"))
+      .post("{}")
+      .map(r => {
+        r.status match {
+          case 200 => true
+          case _ =>
+            logger.error(s"Export api response ${r.status} ${r.body}")
```
suzannehamilton, line 37: Logging the status code would also be handy, in case there's no body but it's a 503 response or something.

MancunianSam: That is the status code in the log. You can see we're matching on it a few lines up.

suzannehamilton: I'm clearly not paying attention this morning. 🤦‍♀️

suzannehamilton (approved):

PR #287, opened 2020-11-25
Query api to get consignment summary

techncl:


suzannehamilton:
@Lholbourn Would you like to review this one?

Let me know if you'd like an extra pair of eyes. 👀

Lholbourn (commented):
Looks like the Jenkins build is failing. I think it's because of the `IFileWithPath` in some of the TypeScript files :thinking:

techncl:
@Lholbourn ah, yes funnily enough, I thought replacing all instances of TdrFile with IFileWithPath would prevent the errors, as I was receiving errors on my local computer but it seems as though the one before passed, probably because the IFileWithPath changes have not been merged yet.

techncl:
> @Lholbourn ah, yes funnily enough, I thought replacing all instances of TdrFile with IFileWithPath would prevent the errors, as I was receiving errors on my local computer but it seems as though the one before passed, probably because the IFileWithPath changes have not been merged yet.

I rolled it back.

Lholbourn:
> > @Lholbourn ah, yes funnily enough, I thought replacing all instances of TdrFile with IFileWithPath would prevent the errors, as I was receiving errors on my local computer but it seems as though the one before passed, probably because the IFileWithPath changes have not been merged yet.
> 
> I rolled it back.

Does that mean it'll break once the IFileWithPath changes are merged?

techncl:
> > > @Lholbourn ah, yes funnily enough, I thought replacing all instances of TdrFile with IFileWithPath would prevent the errors, as I was receiving errors on my local computer but it seems as though the one before passed, probably because the IFileWithPath changes have not been merged yet.
> > 
> > 
> > I rolled it back.
> 
> Does that mean it'll break once the IFileWithPath changes are merged?

Yes. I'll just merge my changes after the IFileWithPath changes are merged. all the other files that use `TdrFile[]` currently would have to be changed to IFileWithPath as well so there will probably be a bunch of failed tests no matter what.

app/services/ConsignmentService.scala
```diff
@@ -53,4 +55,11 @@ class ConsignmentService @Inject()(val graphqlConfiguration: GraphQLConfiguratio
     sendApiRequest(getConsignmentFolderDetailsClient, getConsignmentFolderDetails.document, token, variables)
       .map(data => data.getConsignment.get)
   }
+
+  def getConsignmentTransferSummary(consignmentId: UUID, token: BearerAccessToken): Future[getConsignmentSummary.GetConsignment] = {
+    val variables: getConsignmentSummary.Variables = new getConsignmentSummary.Variables(consignmentId)
+
+    sendApiRequest(getConsignmentSummaryClient, getConsignmentSummary.document, token, variables)
+      .map(_.getConsignment.get)
```
Lholbourn, line 63: Does the `_.getConsignment.get` function similarly to `data => data.getConsignment.get` in the above methods? Just wondering if theres a difference in functionality with that syntax?

app/controllers/TransferSummaryController.scala
```diff
@@ -17,44 +18,65 @@ import scala.concurrent.{ExecutionContext, Future}
...
   )
 
-  def transferSummary(consignmentId: UUID): Action[AnyContent] = consignmentExists(consignmentId) { implicit request: Request[AnyContent] =>
-    Ok(views.html.transferSummary(consignmentId, transferSummaryForm))
+  private def getConsignmentSummary(request: Request[AnyContent], consignmentId: UUID)
+                                   (implicit requestHeader: RequestHeader): Future[ConsignmentSummaryData] = {
+    consignmentService.getConsignmentTransferSummary(consignmentId, request.token.bearerAccessToken)
+      .map{summary =>
+        ConsignmentSummaryData(summary.series.get.code,
+          summary.transferringBody.get.name,
```
Lholbourn, line 41: Do you know what type of error this method would throw if you try to do this for a consignment with no consignment summary info yet?

app/controllers/TransferSummaryController.scala
```diff
@@ -17,44 +18,65 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    }
+  def transferSummary(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    getConsignmentSummary(request, consignmentId)
+      .map{ consignmentSummary =>
+        Ok(views.html.transferSummary(consignmentId, consignmentSummary, transferConfirmationForm))
+      }
+  }
 
-    val successFunction: TransferSummaryData => Future[Result] = { _: TransferSummaryData =>
+  def transferSummarySubmit(consignmentId: UUID): Action[AnyContent] =
```
Lholbourn, line 53: This is really cool, took me a second to understand but looks logical and a good way of doing the error handling for the form! :+1:

app/services/ConsignmentService.scala
techncl: Ah, yes, you're right!, I'll make the "data" one look like the underscore one. Thanks

app/views/transferSummary.scala.html
```diff
@@ -15,24 +15,24 @@ <h1 class="govuk-heading-xl">@Messages("transferSummary.title")</h1>
...
                     </dd>
                 </div>
                 <div class="govuk-summary-list__row">
                     <dt class="govuk-summary-list__key">
                         @Messages("transferSummary.filesUploadedForTransfer")
                     </dt>
-                    <dd class="govuk-summary-list__value summary-list__right-column">
-
+                    <dd class="govuk-summary-list__value">
+                        @summary.totalFiles @if(summary.totalFiles == 1) {file} else {files} uploaded
```
Lholbourn, line 35: This is cool! I didn't know you could do the messages with arguments like that! :1st_place_medal:

app/views/transferSummary.scala.html
```diff
@@ -15,24 +15,24 @@ <h1 class="govuk-heading-xl">@Messages("transferSummary.title")</h1>
...
                     </dd>
                 </div>
```
Lholbourn, line 38: This empty line could probably be removed :+1:

app/views/transferSummary.scala.html
```diff
@@ -45,19 +45,19 @@ <h1 class="govuk-heading-xl">@Messages("transferSummary.title")</h1>
                 <fieldset class="govuk-fieldset" aria-describedby="transfer-summary">
                     <!-- Open records & transfer legal ownership -->
                     @inputSingleCheckbox(
-                    transferSummaryForm("openRecords"),
+                    transferConfirmationForm("openRecords"),
                     Symbol("_label") -> Messages("transferSummary.openRecords"),
```
Lholbourn, line 49: I know the phrasing of this message isn't part of this PR but I don't know if it might be a good idea for us to flag it. The phrasing doesn't quite make sense to me, don't know if you have any feelings about it too?

app/controllers/TransferSummaryController.scala
techncl: Yes. This is something I was speaking to Suzanne about last week. Since you need a series code and transferring body to get a consignment, that part would be fine, the only part is the file upload. Currently, if you access that page before the upload, it just says "0 files" which is ok, but we probably wouldn't want them to continue on and click "transfer your records". Accessing pages before you are supposed to is something that Pia is looking into.

app/controllers/TransferSummaryController.scala
techncl: Thanks, but I got the idea from the transfer agreement controller. It's a pretty nifty feature

test/controllers/TransferSummaryControllerSpec.scala
```diff
@@ -48,23 +49,30 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
...
 
   def exportService(configuration: Configuration): ConsignmentExportService = {
     val wsClient = new InternalWSClient("http", 9007)
     new ConsignmentExportService(wsClient, configuration, new GraphQLConfiguration(configuration))
   }
 
-  "TransferSnummaryController GET" should {
+  "TransferSummaryController GET" should {
 
     "render the transfer summary page with an authenticated user" in {
```
Lholbourn, line 62: I think you could remove this empty line :+1:

app/views/transferSummary.scala.html
techncl: Ah yeah, thanks, got there through quite a bit of trial and error. I just thought, "since the HTML content is a string, you might be able to put these together and have them concatenate"

test/controllers/TransferSummaryControllerSpec.scala
```diff
@@ -75,49 +83,43 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
       playStatus(transferSummaryPage) mustBe OK
       contentType(transferSummaryPage) mustBe Some("text/html")
       contentAsString(transferSummaryPage) must include("transferSummary.header")
+
       contentAsString(transferSummaryPage) must include("transferSummary.seriesReference")
+      contentAsString(transferSummaryPage) must include(seriesCode.get.code.get)
```
Lholbourn, line 88: Is there a reason this can't just be something like `seriesCode.code` or `seriesCode.code.value` rather than `seriesCode.get.code.get`? More just wondering than anything else, same for the others below.

test/controllers/TransferSummaryControllerSpec.scala
```diff
@@ -156,11 +174,13 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
     }
 
     "redirects correctly when the call to the export api fails" in {
```
Lholbourn, line 176: Could this potentially be re-worded, at first glance `redirects correctly` doesn't tell you where the redirect should be to?

app/views/transferSummary.scala.html
techncl: Yeah, maybe "I confirm that in this version of TDR, all records that I **am transferring** are open" and/or maybe change "in this version of TDR" because it sounds ambiguous. We should probably let Kirsten know.

Lholbourn (changes requested):
All looks good! Just a few questions/queries really :+1:

app/views/transferSummary.scala.html
Lholbourn: Yeah the 'in this version of TDR' is the biggest thing throwing me off, but 'that I transfer' also doesn't sound completely smooth. I agree we should mention it to Kirsten :+1:

test/controllers/TransferSummaryControllerSpec.scala
techncl: These export API ones are ones that Sam added, so I'll ask him.

test/controllers/TransferSummaryControllerSpec.scala
techncl: Unfortunately, it has to be this way because the getConsignmentSummary expects an option (the series) and the series code within it is an option

Lholbourn: Ah I see! :+1:

Lholbourn (approved):
All looks good to me! :rocket:

PR #300, opened 2020-12-04
Add error summary to series details page

MancunianSam:
When using a screen reader with the series details page, if you don't select a series and try to continue, the screen reader won't currently read the error that's attached to the form. The advice from gov uk is to do the following:

1. show them the page again, with the form fields as the user filled them in
2. add ‘Error: ’ to the beginning of the page <title> so screen readers read it out as soon as possible
3. show an error summary at the top of the page, and move keyboard focus to it
4. show error messages next to fields with errors

We're already doing 1 and 4 but we need to add in 2 and 3 so that the screen reader can read out the error. I've added in 2 and 3 for the series page. The error summary is a generic template which can be added to other pages and the title change just needs a boolean hasError parameter passing to it so we can easily implement this for other pages if we need to.


TomJKing (approved):
Looks good

PR #309, opened 2020-12-09
Return the correct action to allow stored form posts to be re-run

MancunianSam:
The problem causing [TDR-244](https://national-archives.atlassian.net/browse/TDR-244) happens if the refresh token expires while you're still logged into the session and then try to submit a form post. These are the code snippets from pac4j of roughly what happens.

The refresh token is expired which causes this to be run 
```
if (startAuthentication(context, currentClients)) {
  LOGGER.debug("Starting authentication");
  saveRequestedUrl(context, currentClients, config.getClients().getAjaxRequestResolver());
  action = redirectToIdentityProvider(context, currentClients);
}
```
`saveRequestedUrl` eventually does this.
```
final String requestedUrl = getRequestedUrl(context);
if (ContextHelper.isPost(context)) {
  LOGGER.debug("requestedUrl with data: {}", requestedUrl);
  final String formPost = RedirectionActionHelper.buildFormPostContent(context);
  context.getSessionStore().set(context, Pac4jConstants.REQUESTED_URL, new OkAction(formPost));
} 
```
You can see that it's saving an `OkAction` with the form post data, i.e. the form html

Next, `action = redirectToIdentityProvider(context, currentClients);` is run which authenticates successfully because your play/keycloak session is still active. To authenticate, it redirects to keycloak and then back again. Once keycloak has redirected back to pac4j's CallbackController, the original request is retrieved from the session.

```
if (requestedAction instanceof FoundAction) {
  return RedirectionActionHelper.buildRedirectUrlAction(context, ((FoundAction) requestedAction).getLocation());
} else {
  return RedirectionActionHelper.buildFormPostContentAction(context, ((OkAction) requestedAction).getContent());
}
```

As you can see from above, the `requestedAction` is an `OkAction` so it hits the else branch. As the method name suggests, this rebuilds the form post html and the default behaviour for pac4j is just to render this in a `<pre>` block on the page. Which is a little odd but there you are. That's why we're seeing html.

Now the only time you get these `OkAction` classes as far as I can see is in this instance where we save and restore a post request. So I've changed the `FrontendHttpActionAdaptor` class to check for them:

```
override def adapt(action: HttpAction, context: PlayWebContext): Result = {
    action match {
      case a: OkAction =>
        Results.ok(Html(Option(a.getContent)))
      case _ => val code = action.getCode
      // ... the existing code in this class

```
Now the call to `Results.ok(Html(Option(a.getContent)))` resubmits the original post request using the html stored in `getContent` and play then treats this as it would any other post request.

I've also taken the opportunity to upgrade pac4j. It hasn't helped at all but it's good to keep these things at the latest versions.

### DIFF
--- a/test/util/FrontEndTestHelper.scala
+++ b/test/util/FrontEndTestHelper.scala
@@ -11,9 +11,11 @@ import org.keycloak.representations.AccessToken
 import org.mockito.ArgumentMatchers.any
 import org.mockito.Mockito._
 import org.pac4j.core.authorization.authorizer.Authorizer
-import org.pac4j.core.client.Clients
+import org.pac4j.core.authorization.checker.AuthorizationChecker
+import org.pac4j.core.client.{Client, Clients}
 import org.pac4j.core.config.Config
 import org.pac4j.core.context.WebContext
+import org.pac4j.core.credentials.Credentials
 import org.pac4j.core.engine.DefaultSecurityLogic
 import org.pac4j.core.http.ajax.AjaxRequestResolver
 import org.pac4j.core.profile.UserProfile
@@ -43,6 +45,7 @@ import viewsapi.FrontEndInfo
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
+import scala.language.existentials
 
 
 trait FrontEndTestHelper extends PlaySpec with MockitoSugar with Injecting with GuiceOneAppPerTest with BeforeAndAfterEach {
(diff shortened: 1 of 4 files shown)


test/util/FrontEndTestHelper.scala
```diff
@@ -123,7 +125,7 @@ trait FrontEndTestHelper extends PlaySpec with MockitoSugar with Injecting with
 
     //Return true on the isAuthorized method
     val logic = DefaultSecurityLogic.INSTANCE
-    logic.setAuthorizationChecker((_: WebContext, _: util.List[UserProfile], _: String, _: util.Map[String, Authorizer[_ <: UserProfile]]) => true)
+    logic.setAuthorizationChecker((_: WebContext, _: util.List[UserProfile], _: String, _: util.Map[String, Authorizer[_ <: UserProfile]], _: util.List[Client[_$1]] forSome {type _$1 <: Credentials}) => true)
     testConfig.setSecurityLogic(logic)
```
MancunianSam, line 129: I don't what or how this type works. I copied it off the signature for the method I'm overriding

TomJKing (approved):

PR #320, opened 2020-12-14
Add 'accept' parameter to file input to fix Safari bug

suzannehamilton:
Specify an arbitrary type in the 'accept' attribute to work around a bug in Safari 14.0.1, which does not let the user browse for files if the 'accept' attribute is missing. The actual value of the attribute is ignored because 'webkitdirectory' is specified. It just needs to be present to fix the Safari bug.

Related bug reports:

- https://stackoverflow.com/q/64843459/679924
- https://discussions.apple.com/thread/252051818


suzannehamilton:
I posted about the general issue in cross-government Slack, and it looks like it's a problem in Mojave but not Catalina. So this workaround is still necessary because it will affect some users, but it's not quite as widespread as thought.

TomJKing (approved):

PR #321, opened 2020-12-14
Add trustedProxies parameter

MancunianSam:
https://www.playframework.com/documentation/2.8.x/HTTPServer#Configuring-trusted-proxies

There are two IPs that the load balancer traffic can come from. These
are set in terraform inside these environment variables.

They're not needed for local development because it's only a way to make
sure that play treats https requests from the alb properly.


suzannehamilton:
I'll review this once we've finished the discussion on https://github.com/nationalarchives/tdr-terraform-environments/pull/107.

conf/application.staging.conf
```diff
@@ -30,6 +30,8 @@ akka.actor.warn-about-java-serializer-usage="off"
 
 play.filters.headers.frameOptions = "SAMEORIGIN"
 
+play.http.forwarded.trustedProxies=[${ALB_IP_A}, ${ALB_IP_B}]
```
suzannehamilton, line 33: This made me think: do you think it's worth moving most of the settings to a base config file like we do in the API? https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/resources/application.intg.conf#L1

suzannehamilton (approved):

PR #325, opened 2020-12-15
Update drag and drop so that it detects multiple folders

techncl:


suzannehamilton:
@techncl Looks like there are some failing Jest tests on Jenkins.

@MancunianSam Once the Jenkins build is passing, would you like to do a review of this? Let me know if you'd like me to review anything too.

techncl:
> @techncl Looks like there are some failing Jest tests on Jenkins.
> 
> @MancunianSam Once the Jenkins build is passing, would you like to do a review of this? Let me know if you'd like me to review anything too.

@suzannehamilton It seems to be an `IFileWithPath` issue. I think because `node_modules/@nationalarchives/file-information` still has `TdrFile` present instead of `IFileWithPath`. If we change it, we'd have to replace all of the other places that use `TdrFile` too. Should I do that on this branch?

suzannehamilton:
Yeah, it probably makes sense to do that on this branch. (The alternative is to create another branch off master and upgrade the library there, but given how many changes you've had to make to integrate the library with the drag-and-drop that would probably be quite tricky!)

npm/test/folderinforetrieval.test.ts
```diff
@@ -57,60 +80,299 @@ const mockFormHTML = `
...
+    ".drag-and-drop__failure"
+  )
+  folderNameElement: HTMLElement | null = document.querySelector("#folder-name")
+  folderSizeElement: HTMLElement | null = document.querySelector("#folder-size")
+
+  form = new UploadForm(this.uploadForm!, this.folderRetriever!, this.dropzone!)
+  fileUploader = this.setUpFileUploader()
+
+  setUpFileUploader(): FileUploader {
+    const client = new GraphqlClient("https://test.im", mockKeycloakInstance)
```
suzannehamilton, line 196: If this is an arbitrary URL, you normally want to use example.com because it was set up be IETF for that purpose: https://en.wikipedia.org/wiki/Example.com

npm/test/folderinforetrieval.test.ts
```diff
@@ -57,60 +80,299 @@ const mockFormHTML = `
...
+
+  expect(mockDom.folderRetrievalSuccessMessage!).not.toHaveClass("hide")
+  expect(mockDom.folderRetrievalfailureMessage!).toHaveClass("hide")
+  expect(mockDom.folderNameElement!.textContent).toStrictEqual("Mock Folder")
+  expect(mockDom.folderSizeElement!.textContent).toStrictEqual("2 files")
+})
+
+test("dropzone updates the page with an error if there are no files in folder", async () => {
+  const mockDom = new MockDom()
+  mockDom.entries.pop() // remove entries from previous test
```
suzannehamilton, line 274: Is it possible to reset MockDom with each test, or make it not have mutable state? This line implies that you can only run the tests in a specific order. It's much easier to update tests that are completely independent.

npm/src/upload/upload-form.ts
```diff
@@ -63,83 +135,117 @@ export class UploadForm {
     })
   }
 
-  addFolderListener() {
-    this.dropzone.addEventListener("drop", (e) => {
-      const droppedObject: File | null = e.dataTransfer!.files[0] // Since we do not allow multiple objects to be dropped, there is always only one file
+  handleDropppedItems: (ev: DragEvent) => any = async (ev) => {
+    const droppedItems: FileList | null = ev.dataTransfer!.files // THIS HAS TO BE A FILELIST
```
suzannehamilton, line 139: What does this comment mean? It sounds IMPORTANT but I'm not sure what I'd do with it if I was updating this code.

npm/src/upload/upload-form.ts
```diff
@@ -63,83 +135,117 @@ export class UploadForm {
...
+    this.checkIfFolderHasFiles(files)
+
+    return [...files].map((file) => ({
+      file,
+      path: (file as FileWithRelativePath).webkitRelativePath
+    }))
+  }
+
+  private rejectUserItemSelection() {
+    this.dropzone.classList.remove("drag-and-drop__dropzone--dragover") // Extract this into its own method to do with file selection
```
suzannehamilton, line 215: This comment sounds like a to-do?

npm/src/upload/upload-form.ts
```diff
@@ -63,83 +135,117 @@ export class UploadForm {
...
+    const successMessage: HTMLElement | null = document.querySelector(
+      ".drag-and-drop__success"
+    )
+    successMessage?.classList.add("hide")
+    const warningMessage: HTMLElement | null = document.querySelector(
+      ".drag-and-drop__failure"
+    )
+    warningMessage?.classList.remove("hide")
+    throw new Error("No files selected")
```
suzannehamilton, line 224: Do we need to throw an exception here? If we display the warning message like we do just above here, do we also need to throw an error?

suzannehamilton (commented):
A few comments which are hopefully quite small. Thanks to both of you (Sam and Nigel) for figuring out all the moving parts here.

I would really like to revisit the tests at some point to see if we can make them simpler. But I appreciate that working with the file selection APIs is complicated, and I'm sorry that I don't know Jest well enough to be able to make concrete suggestions in a code review.

npm/test/folderinforetrieval.test.ts
MancunianSam: It's arbitrary, I've updated to use `example.com`

npm/test/folderinforetrieval.test.ts
MancunianSam: It's not so much that it's not resetting, it's that the mocked files are hardcoded into the MockDom class, so to simulate zero files, they have to be removed for this test. I've moved it out to the class constructor and passed in a numberOfFiles argument. It's not really removing the logic but it's keeping it within the class.

I agree with your point on the tests but the functions for these File API methods is a mess of callbacks, symbol arrays and all sorts of fun things so as you say, they're ok for now and we can revisit them later.

npm/src/upload/upload-form.ts
MancunianSam: I'll let @techncl have the last word on this one but I don't think we need the comment because the type shows that it's a `FileList`

npm/src/upload/upload-form.ts
MancunianSam: Again, I'll let @techncl have final say but this is repeated a couple of times in the class with the same specific string for the class so it makes sense to have it as it's own function. I've moved it out but I can always put it back.

npm/src/upload/upload-form.ts
techncl: I think (but could be wrong) it's because @suzannehamilton either you were wondering why we were both accessing files via `ev.dataTransfer!.files` and `ev.dataTransfer?.items!` or it's because in an earlier iteration of this, where we were going to create the classes for the DroppedFile and BrowsedFiles, I think you might have suggested a ev.dataTransfer?.items like in the prototype.  I apologise, I should have been clearer; it probably made sense at the time :).

npm/src/upload/upload-form.ts
techncl: Yes, you are both correct, I forgot to do it. I thought writing comments in my code would remind me to add the feature, but I guess I need a reminder to look at my reminders :smile:

npm/src/upload/upload-form.ts
techncl: Hmm, it was there from before, it did serve me well when debugging (if you have multiple red warning messages in a row e.g. you upload multiple folders one time and then you upload a file), you might not realise that there were two errors) but it's probably not necessary now.

MancunianSam: There's few places in the code this is used. It is possible to refactor it with if else statements but I just tried and it gets messy quite quickly. It's just a convenient way of stopping the execution of the code that follows.

npm/test/folderinforetrieval.test.ts
techncl: Thanks @MancunianSam! Yes, I added the files directly to the mockDom to save having to add them each time; my comment could have been a little clearer though, I think they did share state initially, before I put them inside the Class

npm/test/folderinforetrieval.test.ts
suzannehamilton: It looks like it's still test.im. Is there a commit that hasn't been pushed?

npm/test/folderinforetrieval.test.ts
suzannehamilton: Parameterising it is good, thanks. 👍

npm/src/upload/upload-form.ts
suzannehamilton: Right, so is the comment meant to explain why we use `ev.dataTransfer!.files` in the first half of the method and `ev.dataTransfer?.items!` in the latter half? I agree that that's a bit confusing, so I think it's worth expanding the comment to explain the difference if it's still necessary to use `files` in one place and `items` in the other.

npm/test/folderinforetrieval.test.ts
MancunianSam: Not sure what happened there. I've updated it again.

npm/src/upload/upload-form.ts
techncl: Now that I think about it more, I think the reason for it might be due to the latter half of my answer (the implementation I was working on required it). Looking at the documentation, I think that we could actually replace `ev.dataTransfer!.files` with `ev.dataTransfer?.items!` as it also has a `length` property and the items within it have a `type` method. What do you think @MancunianSam?

MancunianSam: I've managed to get this to work now using only `ev.dataTransfer?.items!`

suzannehamilton (approved):
Thanks for both of your hard work on this. 🎉

PR #326, opened 2020-12-15
Pass None instead of allDigital form value

TomJKing:
Requires: https://github.com/nationalarchives/tdr-consignment-api/pull/101

To safely remve allDigital need to pass in None value first

This will then allow the allDigital parameter to be removed from the API, whilst not breaking the frontend before removal of the parameter from the front end


suzannehamilton (approved):
Just to check - are you making sure these get deployed to staging in the right order, too?

PR #367, opened 2021-01-06
Extend the connection timeout.

MancunianSam:
It looks like the connection timeout is causing the problems with large
file uploads. At least, when I make it very short, I can reliably make
the problem happen and Tom is getting errors after 2 minutes which is
the default timeout. I've upped this to 20 minutes to see if this will
help. It's a bit trial and error at the moment so we'll see.

### DIFF
--- a/npm/src/s3upload/index.ts
+++ b/npm/src/s3upload/index.ts
@@ -17,7 +17,9 @@ export class S3Upload {
   s3: TdrS3
   identityId: string
   constructor(identityId: string) {
-    this.s3 = new S3()
+    const timeout = 20 * 60 * 1000
+    const connectTimeout = 20 * 60 * 1000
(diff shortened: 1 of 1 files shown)


TomJKing (approved):

suzannehamilton (changes requested):
Before we merge this, what is the default and do you know why it's shorter than this? I'd like to understand the consequences of this change first!

suzannehamilton:
Ack. Got there too late. I'd still like to know the answers to those questions. 🙂

MancunianSam:
The default is two minutes. I think they assume that's a reasonable time to upload the default 5Mb chunk.

MancunianSam:
We can always put it back if this doesn't fix the problem. As I said, this is going to be a lot of trial and error.

PR #389, opened 2021-01-15
Accessibility changes

MancunianSam:
* The success and error message are not focused on when displayed after choosing a file so if you're using a screen reader, you can't easily read the message. I've changed it so after the files are selected, the focus is set on the error/success message in the javascript.
* The success and error message are missing the aria label which makes them an alert/dialog. I've added and aria label to both so to help the screen reader.
* Text for the success and error messages wasn't being read by the screen reader even if the alert div is focused. I've added an aria-describedby attribute and put the text inside a <p> tag so this is read out once the focus changes.
* Without css, the page looks very different as seen in the screenshot on the [ticket](https://national-archives.atlassian.net/browse/TDR-522) All of the error messages were being hidden with css. Without it, all the error messages are shown which looks quite confusing. Also, the svgs for the icons are enormous. I've used hidden to hide elements and hidden is set and removed as appropriate in the javascript. This involves making `hidden` !important in the css because hidden is overridden by any changes to the display property which we don't want. Now the behaviour is the same with css but the hidden fields don't show when it's switched off. I've also manually set a height and width on the svg to stop them appearing so large. I don't know if this is the best solution but it works.
The disallowed file extension detail text can't be tabbed to. I've added tabindex 0 so it can be tabbed to when it's visible.


app/views/greenTickMark.scala.html
```diff
@@ -2,6 +2,6 @@
 * This is the SVG for the green tick mark that goes before a message to indicate a successful event has occurred.
 *@
 
-<svg class="green-tick-mark" role="presentation" focusable="false" viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg">
+<svg class="green-tick-mark" role="presentation" focusable="false" viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg" width="25px" height="25px">
```
TomJKing, line 5: Not sure styling should be directly added to the HTML, ie the dimensions?

TomJKing (commented):
It all looks OK to me, got one comment. Though I'm not an expert on CSS either.

app/views/greenTickMark.scala.html
MancunianSam: Usually I think you're right but the problem with it is that if it's set in the css and css is turned off, the svg icons are massive and take up most of the page so this keeps them small without changing their size on the page with css on. I'm not sure if there's a better way.

TomJKing: Not sure, what is the correct way to go with this either, good practice v. good accessibility, which has greater weight. Probably needs some SVG magic to get to work properly. One for later I think.

TomJKing (approved):

PR #394, opened 2021-01-19
Set the upload error to show on error.

MancunianSam:
At the moment, when we get into the handleUploadError function, we're hiding the error message instead of showing it which is less than helpful.


TomJKing (approved):
Assume the e2e tests will still work with this change.

MancunianSam:
I found it because the e2e tests were breaking because they couldn't see the error messages. This should fix it.

PR #400, opened 2021-01-20
Tdr 769 remove metadata step

TomJKing:


app/views/fileChecksResults.scala.html
```diff
@@ -21,7 +21,7 @@ <h1 class="govuk-panel__title">
         <button class="govuk-button govuk-button--secondary" data-module="govuk-button">
             @Messages("fileChecksResults.dashboard")
         </button>
-        <a href="/consignment/@consignmentId/custom-metadata" role="button" draggable="false" class="govuk-button" data-module="govuk-button">
+        <a href="/consignment/@consignmentId/transfer-summary" role="button" draggable="false" class="govuk-button" data-module="govuk-button">
             @Messages("fileChecksResults.continueLink")
```
MancunianSam, line 25: It's a small thing and I know the original is in a different format but all of the other button links have the href in the form `@routes.TransferSummaryController.transferSummary(consignmentId)` Could we change this one as well just to keep them consistent?

TomJKing: Good point.

MancunianSam (approved):
LGTM

PR #401, opened 2021-01-20
Create transfer complete page.

MancunianSam:
We already had a placeholder but this is implementing the design for the page.
We're not displaying a specific error if triggering the export fails so I've removed the parameter and the test for it.
I've also renamed everything from transfer confirmation to transfer complete to match with the title.
I've also changed the export graphql and export api calls to fail if there are errors rather than recovering from them and returning false as they're doing at the moment.


test/controllers/TransferCompleteControllerSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+import java.util.UUID
+
+import play.api.test.CSRFTokenHelper.CSRFRequest
+import play.api.test.FakeRequest
+import play.api.test.Helpers.{GET, contentAsString, defaultAwaitTimeout}
+import util.FrontEndTestHelper
+
+class TransferCompleteControllerSpec extends FrontEndTestHelper {
+
+  "TransferConfirmationController GET" should {
```
TomJKing, line 12: Should this read: "Transfer**Complete**Controller ..."

test/controllers/TransferSummaryControllerSpec.scala
```diff
@@ -155,6 +155,7 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
     "redirects to the transfer confirmation page when a valid form is submitted" in {
```
TomJKing, line 155: Should this read: "... to the transfer **complete** page .."?

TomJKing (commented):
Looks good. Couple of renaming comments.

test/controllers/TransferCompleteControllerSpec.scala
MancunianSam: It should yeah, cheers. I've changed it.

test/controllers/TransferSummaryControllerSpec.scala
MancunianSam: I've changed this one as well.

TomJKing (approved):

PR #407, opened 2021-01-27
Remove iconify JavaScript library

suzannehamilton:
This icon library is not being used. We think it was copied from the prototype repo accidentally.


MancunianSam (approved):
I seem to remember adding this to the prototype so I think you're right :+1:

PR #411, opened 2021-01-28
Updated step counter

plembke-tna:
Updated step counter to a total of 5, removed step counter from transferComplete page, added step counter on fileChecksResults page, and updated them on the other pages


app/views/transferComplete.scala.html
```diff
@@ -4,7 +4,7 @@
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
         <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>
-        @progressIndicator(Messages("transferComplete.progress"))
```
suzannehamilton, line 7: Just to check - are you and Kirsten both happy with the fact that this page won't have a Step header on it at all? I'd have thought users would want to see what step they're on.

suzannehamilton: Ah, sorry, this is the final page, isn't it? So I guess it doesn't make sense to have a Step label here.

suzannehamilton (approved):

app/views/transferComplete.scala.html
plembke-tna: Yes it is the final page, I will double-check with Kirsten though! Thank you!

PR #423, opened 2021-02-04
Added beta banner

plembke-tna:
Added beta banner on every page - without a link to feedback survey for now.


conf/messages
```diff
@@ -1,6 +1,8 @@
 # https://www.playframework.com/documentation/latest/ScalaI18N
 
 main.header=Transfer Digital Records
+main.banner=beta
+main.banner.info=This is a new service – your feedback will help us to improve it.
```
suzannehamilton, line 5: I think you normally want to include a feedback link, but maybe that's OK for private beta the users will already be in direct contact with us? Is Kirsten happy with using this wording without a link?

app/views/fileChecksResults.scala.html
```diff
@@ -6,6 +6,8 @@
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
         <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>
+        @progressIndicator(Messages("fileChecksResults.progress"))
```
suzannehamilton, line 9: You've got some progress indicator changes here. Were they mixed up with the beta banner commit by mistake?

conf/messages
plembke-tna: I think it should be okay not have a link and add it later if I make a note. I will check with Kirsten though!

app/views/fileChecksResults.scala.html
plembke-tna: Because I hadn't merged the update-step-counter branch, I updated the step counter again which wasn't necessary. Thanks for spotting, I'm fixing it!

suzannehamilton (approved):

PR #424, opened 2021-02-04
Add links to footer

rsulxo:
Updated AC according to chat with KA on 25/2:
FAQ, help guide and accessibility statement still require work. So for now the only change to the footer will be the contact link for the TDR mailbox. The rest of the items can be updated once the content has been finalised.


app/views/main.scala.html
```diff
@@ -59,6 +59,32 @@
...
+                          </a>
+                        </li>
+                        <li class="govuk-footer__inline-list-item">
+                          <a class="govuk-footer__link" href="@routes.AccessibilityController.accessibilityStatement()">
+                            Accessibility statement
+                          </a>
+                        </li>
+                        <li class="govuk-footer__inline-list-item">
+                          <a class="govuk-footer__link" href="">
+                            Cookies
```
suzannehamilton, line 80: I think we came to the conclusion that we don't need a cookies page for now because we're only using essential cookies for auth - we don't have any analytics cookies that would need a cookie page. Is that right? Can we remove this link?

app/controllers/HelpGuideController.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import auth.OidcSecurity
+import javax.inject.{Inject, Singleton}
+import org.pac4j.play.scala.SecurityComponents
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, Request}
+
+@Singleton
+class HelpGuideController @Inject()(val controllerComponents: SecurityComponents) extends OidcSecurity with I18nSupport  {
+  def helpGuide(): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
```
suzannehamilton, line 11: Given how similar the pages are, and the fact that the controllers don't do anything complicated, I'd be inclined to merge the `helpGuide` and `accessibilityStatement` into a single controller. Maybe call it `HelpController`?

app/views/main.scala.html
```diff
@@ -59,6 +59,32 @@
...
             <div class="govuk-footer__meta">
                 <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+                    <ul class="govuk-footer__inline-list">
+                        <li class="govuk-footer__inline-list-item">
+                          <a class="govuk-footer__link" href="@routes.HelpGuideController.helpGuide()">
+                            Help
+                          </a>
+                        </li>
+                        <li class="govuk-footer__inline-list-item">
+                          <a class="govuk-footer__link" href="#">
```
suzannehamilton, line 69: Like we said in sprint planning a couple of weeks ago, let's skip any links for now if we don't have the pages yet. It's not hard to add them in later, and it saves confusion by not having a broken link in the meantime.

app/views/accessibilityStatement.scala.html
```diff
@@ -0,0 +1,58 @@
+@()(implicit messages: Messages)
+
+@main(Messages("accessibility.header")) {
+@defining(play.core.PlayVersion.current) { version =>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">@Messages("accessibility.title")</h1>
```
suzannehamilton, line 7: Wow, the messages file makes it hard to work with text-heavy pages like this! I'm going to raise this with the whole dev team, because I was never sure if the messages file was all that useful when we're fairly sure TDR will never be multilingual.

No need to do anything differently for this PR, though. If we remove the messages then it will happen later.

conf/messages
```diff
@@ -101,6 +101,127 @@ upload.progress.error.general=Something has gone wrong, click <a href="/series">
...
+accessibility.header6=Contacting us or visiting in person
+accessibility.body12=You can contact us by email. We provide information about visiting in person. People with health conditions or impairments that may impact their visit can find out more.
+accessibility.header7=Technical information about this website’s accessibility
+accessibility.body13=The National Archives is committed to making its website accessible, in accordance with the Public Sector Bodies (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018.
+accessibility.header8=Compliance status
+accessibility.body14=This website is partially compliant with the Web Content Accessibility Guidelines version 2.1 AA standard, due to the non-compliances and exemptions listed below:
+accessibility.subheader1=Non-accessible content
+accessibility.body15=Based on sample testing of pages selected by our user experience and metrics teams, we know that the content listed below is non-compliant with the accessibility regulations for the following reasons:
+accessibility.subheader2=Non-compliance with the accessibility regulations
+accessibility.item7=[list issues here]
```
suzannehamilton, line 215: I'm going to see if I can catch Kirsten for a quick chat about these pages. I want to know how she feels about deploying them while they're still in progress, or if it would be better to wait until we've written the full statement and asked Ashley to review.

app/views/main.scala.html
rsulxo: Yes definitely!

app/views/main.scala.html
rsulxo: No worries, I'll remove the link for 'contact' for now then :)

conf/messages
rsulxo: Yess, it's a bit tricky because even though the pages exist with the drafts in both, it's an ongoing process at the moment. Ashley is working on finishing the accessibility statement, so that might be something which can be deployed (help guide is still a WIP as we don't have the guidance video etc yet)

app/views/accessibilityStatement.scala.html
rsulxo: I know right! 😅  I didn't think it would be so difficult to come up with various naming conventions for all the text, but we'll probably need to think of a better solution in the long run.

app/controllers/HelpGuideController.scala
rsulxo: Sounds good, I'll have a look into changing that, thank you!

conf/messages
suzannehamilton: Just unresolving this comment because when I talked to Kirsten last week, she said that she only wants to deploy these changes once we've cleared up all of the to-do comments. Did she talk to you about it as well?

app/views/main.scala.html
```diff
@@ -59,6 +59,17 @@
         <div class="govuk-width-container ">
             <div class="govuk-footer__meta">
                 <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+                    <ul class="govuk-footer__inline-list">
+                        <li class="govuk-footer__inline-list-item">
+                          <a class="govuk-footer__link" href="mailto:tdr@@nationalarchives.gov.uk">
```
suzannehamilton, line 64: Is the double `@` sign a copy-paste error? Or is it some Play templating syntax?

rsulxo: the double `@` is for mailto to work with the play framework https://stackoverflow.com/questions/42714889/mailto-link-not-working-using-play-framework-in-scala-html

app/views/main.scala.html
```diff
@@ -69,7 +80,7 @@
                     </span>
                 </div>
                 <div class="govuk-footer__meta-item">
-                    <a class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
+                    <a class="govuk-footer__link">© Crown copyright</a>
```
suzannehamilton, line 83: I'm going to ask Kirsten about the logo issue, but what's the reason for removing the link as well? I would have thought we'd want to keep that to explain what Crown Copyright means?

rsulxo: Okay no worries, we can revisit the logo issue after you've spoken with her. And oops I didn't realise I had removed it 🙈  I will add it back

conf/messages
rsulxo: (removed the discussed items from this PR)

PR #444, opened 2021-02-11
Removed back links from UI

plembke-tna:
Commented the back links for private beta. We will use them later which is why I didn't delete the code.

### DIFF
--- a/app/views/seriesDetails.scala.html
+++ b/app/views/seriesDetails.scala.html
@@ -10,7 +10,6 @@
 
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-@*        <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>*@
         @progressIndicator(Messages("seriesDetails.progress"))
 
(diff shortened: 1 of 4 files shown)


app/views/seriesDetails.scala.html
```diff
@@ -10,7 +10,7 @@
 
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-        <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>
+@*        <a href="@routes.DashboardController.dashboard()" class="govuk-back-link">Back</a>*@
```
suzannehamilton, line 13: Could you delete these rather than commenting out? Commented-out code is confusing to work with, because if you aren't aware of the context then you don't know if it was supposed to be a temporary comment that's been committed by accident or if it was deliberate.

They'll be in the git history when we want to bring them back. 🙂

plembke-tna: Yes of course! Sorry about that.

suzannehamilton (approved):

PR #450, opened 2021-02-15
Handle final transfer confirmation form submission

techncl:


app/controllers/TransferSummaryController.scala
```diff
@@ -45,38 +51,46 @@ class TransferSummaryController @Inject()(val controllerComponents: SecurityComp
...
 
-    val formValidationResult: Form[TransferConfirmationData] = transferConfirmationForm.bindFromRequest()
-    formValidationResult.fold(
-      errorFunction,
-      successFunction
-    )
-  }
+        val variables: AddFinalTransferConfirmation.Variables = AddFinalTransferConfirmation.Variables(addFinalTransferConfirmationInput)
+
+        sendApiRequest(addFinalTransferConfirmationClient, AddFinalTransferConfirmation.document, request.token.bearerAccessToken, variables)
```
MancunianSam, line 74: At the moment, if there is a failure in sending the API request, the code will still go ahead and update the transfer initiated date and trigger the export which we don't want. It is also possible with it like this for the request to fail because the play request finishes first. The easiest way to get it to wait is to put it into the for comprehension like this:

```
        for {
          _ <- sendApiRequest(addFinalTransferConfirmationClient, AddFinalTransferConfirmation.document, request.token.bearerAccessToken, variables)
          _ <- consignmentExportService.updateTransferInititated(consignmentId, request.token.bearerAccessToken)
         ...
```

Unfortunately, this breaks your tests because the wiremock stubs aren't set up for this api call. Because this is graphql and both calls go to the same endpoint, you need to differentiate between them by the request body. I've hacked this together so it'll need tidying up but this is the general idea

```
    val client = new GraphQLConfiguration(app.configuration).getClient[aftc.Data, aftc.Variables]()
    val data: client.GraphqlData = client.GraphqlData(Option(aftc.Data(AddFinalTransferConfirmation(consignmentId, false, false))), List())
    val dataString: String = data.asJson.printWith(Printer(dropNullValues = false, ""))
    val query = "{\"query\":\"mutation AddFinalTransferConfirmation($input:AddFinalTransferConfirmationInput!)" +
      "{addFinalTransferConfirmation(addFinalTransferConfirmationInput:$input)" +
      "{consignmentId finalOpenRecordsConfirmed legalOwnershipTransferConfirmed}}\"" +
      ",\"variables\":{\"input\":{\"consignmentId\":\"" + consignmentId.toString + "\",\"finalOpenRecordsConfirmed\":true," +
      "\"legalOwnershipTransferConfirmed\":true}}}"
      .withRequestBody(equalToJson(query))
      .willReturn(okJson(dataString)))

    val utClient = new GraphQLConfiguration(app.configuration).getClient[ut.Data, ut.Variables]()
    val utData: utClient.GraphqlData = utClient.GraphqlData(Some(ut.Data(Option(1))), List())
    val utDataString: String = utData.asJson.printWith(Printer(dropNullValues = false, ""))
    val utQuery = "{\"query\":\"mutation updateTransferInitiated($consignmentId:UUID!){updateTransferInitiated(consignmentid:$consignmentId)}\"" +
      ",\"variables\":{\"consignmentId\":\"" + consignmentId.toString + "\"}}"
      .withRequestBody(equalToJson(utQuery))
      .willReturn(okJson(utDataString)))
```
Hope I've explained this properly :)

MancunianSam:
It looks good mate. If you move that statement into the for comprehension and then sort out wiremock so the tests pass then this should be good.

app/controllers/TransferSummaryController.scala
techncl: Thanks Sam! :)

MancunianSam (approved):
That's neatly done, thanks for sorting that out.

PR #458, opened 2021-02-18
Get consignment reference

Lholbourn:
These changes get the consignment reference and render on the transfer summary page. Existing methods within the `TransferSummaryController` have been edited to retreive the consignment reference and pass it to the view so it can be rendered for the user. 

I have also updated the relevant tests to include the consignment reference and updated the generated graphql version.

### DIFF
--- a/test/controllers/TransferSummaryControllerSpec.scala
+++ b/test/controllers/TransferSummaryControllerSpec.scala
@@ -68,7 +68,7 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
       val seriesCode = Some(gcs.GetConsignment.Series(Some("Mock Series")))
       val transferringBodyName = Some(gcs.GetConsignment.TransferringBody(Some("MockBody")))
       val totalFiles: Int = 3
-      val consignmentReference = Option("TEST-TDR-2021-MTB")
+      val consignmentReference = Some("TEST-TDR-2021-MTB")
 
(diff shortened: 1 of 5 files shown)


test/controllers/TransferSummaryControllerSpec.scala
```diff
@@ -68,8 +68,9 @@ class TransferSummaryControllerSpec extends FrontEndTestHelper {
       val seriesCode = Some(gcs.GetConsignment.Series(Some("Mock Series")))
       val transferringBodyName = Some(gcs.GetConsignment.TransferringBody(Some("MockBody")))
       val totalFiles: Int = 3
+      val consignmentReference = Option("TEST-TDR-2021-MTB")
```
TomJKing, line 71: Probably should use `Some` constructor instead of `Option` constructor, as sure the value is not a `null`

TomJKing (commented):
Looks good. One minor comment. Also will need to update the end to end tests for the Transfer Summary page. There is another PR which will affect this code: https://github.com/nationalarchives/tdr-transfer-frontend/pull/460. So will need to deal with any conflicts depending which PR gets merged

test/controllers/TransferSummaryControllerSpec.scala
Lholbourn: Cool, I'll change it to `Some` 👍

TomJKing (approved):

TomJKing (dismissed):

TomJKing (dismissed):

TomJKing (approved):

PR #459, opened 2021-02-18
Replaced radio buttons with checkboxes

plembke-tna:
This change is for private beta/MVP. It looks fine on the front-end and works with the error messages. Is this being captured correctly in the database?


TomJKing (approved):
Looks good. In asnwer to your question, yes the form information is saved to the database. However these changes will cause the e2e tests to fail. You will need to create a new e2e test branch with the fix. Then deploy both changes.

PR #504, opened 2021-03-12
Updated heading levels and spacing

plembke-tna:
Changed some paragraphs with class of a heading into headings and updated all headings to be the same level (making sure they have the correct hierarchy) matching the GDS guidelines and templates.
Also added the missing divs on the dashboard page and changed 'Transfer Complete' to 'Transfer complete' to be consistent with our wording.


app/views/dashboard.scala.html
```diff
@@ -1,11 +1,15 @@
 @()(implicit messages: Messages)
 
 @main(Messages("dashboard.header")) {
-<h1 class="govuk-heading-xl">@Messages("dashboard.title")</h1>
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
```
suzannehamilton, line 5: What's the purpose of the two-thirds column on this page?

plembke-tna: The purpose of it is to keep it consistent to the other pages and to avoid the content to be wider than we want (in this case only the heading, but it might affect more in future depending on what we add to the page).

suzannehamilton: Ah, I didn't realise it was the same on other pages. Do you think it's necessary? There are a few things in our basic page templates that were copied from elsewhere without really thinking about it.

plembke-tna: Yes it is actually really important because users don't read when the text is getting to long. There are numerous studies on this and GDS included it here https://design-system.service.gov.uk/styles/layout/#common-layouts and it says "For most types of page, we recommend using either a ‘two-thirds’ or a ‘two-thirds / one third layout’. That stops lines of text getting so long that the page becomes difficult to read on desktop devices." :-)

suzannehamilton: Great, thanks - just wanted to make sure it was an intentional change, not something we'd copied from somewhere else without considering!

suzannehamilton (approved):
I'm approving this, but you might want to check if it's going to break any end-to-end tests and update them at the same time as merging this PR.

plembke-tna:
It actually did break a couple of end-to-end tests. I have just fixed them and created a PR and will merge both PRs once that is approved. Thank you Suzanne for helping me.

PR #528, opened 2021-03-24
Disable upload page buttons after initiated upload

techncl:


npm/src/upload/upload-form.ts
```diff
@@ -175,6 +175,8 @@ export class UploadForm {
   ) {
     this.formElement.addEventListener("submit", (ev) => {
       ev.preventDefault()
+
+      this.disableButtonsAndDropzone()
```
suzannehamilton, line 179: Let's have a chat about this in our 1-1 today. I think this change will help in most circumstances, but there are some subtle edge-cases that this might not catch.

suzannehamilton: It would also be good to have a test to make sure the form only gets submitted once, but let's chat about the implementation first.

suzannehamilton:
Also temporarily closing this one until Nigel can get back to it to reduce the list of open PRs that need to be checked.

suzannehamilton (dismissed):
🚀

suzannehamilton (approved):

PR #541, opened 2021-03-29
Update test file to reflect current upload page

techncl:


suzannehamilton:
What's the purpose of this PR? There are some commits which add and then remove tests, so it's quite hard to follow what's going on and whether it's OK to merge it as-is.

techncl:
Ah yes, sorry about that; the purpose was just clean up; in the end, these were the changes made

1. I noticed that the HTML in the test didn't match the one that was on the frontend and so I changed it. 
2. I made a variable that wasn't camel case, camel case
3. Removed a test that wasn't actually doing anything
4. Removed two methods that were actually duplicated at the top of he page (they were moved out of the class but never deleted within the class)
5. Renamed the file from folderinforetriever to "upload-form.test" to match the ts file that it is testing

Initially I was going to add a basic test for the submit button since I noticed that we didn't have one, but since it was complicated and required more though, I thought it might be easier to keep it simple and decided that it was probably better to do in another PR.
  
It might be better to just look at the end [result](https://github.com/nationalarchives/tdr-transfer-frontend/pull/541/files)

suzannehamilton:
Just going to temporarily close this until Nigel has time to pick it up again, so we don't have to check it each morning in the open PR list.

npm/test/upload-form.test.ts
```diff
@@ -40,53 +38,57 @@ class MockDom {
...
+                                  <button class="govuk-button" type="submit" data-module="govuk-button" role="button">
+                                      Start upload
+                                  </button>
+                              </div>
+                          </div>
+                      </div>
+                  </div>
+              </form>
+                      <!--        Form to redirect user once upload has completed. It sends consignmentId to record processing placeholder page -->
+              @form(routes.FileChecksController.recordProcessingPage(consignmentId), Symbol("id") -> "upload-data-form") { }
```
suzannehamilton, line 88: Is this `@form` part of the Scala templates? Is it necessary for the test? If so, could you replace with HTML? If not, we could skip it altogether.

techncl: No, I don't think it'll be necessary; I'll remove it. Thanks.

suzannehamilton (approved):

techncl:
Thank you! 🙂

PR #554, opened 2021-04-06
Move govuk-frontend initialization into main TypeScript module

suzannehamilton:
Removed the inline script which used to initialize govuk-frontend.

The two methods of initialization are equivalent: https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript

The advantage of doing the initialization in the TypeScript module instead is that it removes inline JavaScript, which means we can prevent all inline JavaScript using a Content Security Policy. This reduces the risk of cross-site scripting attacks. The CSP will be added in a later commit.


TomJKing (approved):
Looks good

PR #556, opened 2021-04-07
Commonise shared config in base config file

suzannehamilton:
Create application.base.conf and move all the config values into it which are the same in all of the dev and environment-specific config files.

This will make it easier to update these values, and to make sure that config is added to all environments at once.

The reason for doing this now is to make it easier to add the [Content Security Policy header config](https://www.playframework.com/documentation/2.8.x/CspFilter), which will be the same in all environments.

It looks possible to commonise even more of the config, but we can revisit that if it looks like it would help us with a future task.


MancunianSam (approved):
LGTM

PR #561, opened 2021-04-09
Add Content Security Policy

suzannehamilton:
Add CSP header to every page. Use the default Play CSP as a basis: https://www.playframework.com/documentation/2.8.x/CspFilter

Override some of the defaults to allow data to be sent to the TDR API and AWS APIs, and to allow the Keycloak silent login iframe.

This produces a CSP header which looks like this in the integration environment, with something similar in other environments:

```
base-uri 'none'; script-src 'self' 'unsafe-eval'; child-src 'self' https://auth.tdr-integration.nationalarchives.gov.uk; default-src 'self'; object-src 'none'; connect-src 'self' https://auth.tdr-integration.nationalarchives.gov.uk https://cognito-identity.eu-west-2.amazonaws.com https://sts.amazonaws.com https://api.tdr-integration.nationalarchives.gov.uk https://tdr-upload-files-dirty-intg.s3.amazonaws.com https://tdr-upload-files-dirty-intg.s3.eu-west-2.amazonaws.com
```


suzannehamilton:
Marking this as draft for now because I'm checking whether it works with the fully local dev environment.

suzannehamilton:
Marking as ready for review because I've tested it with the full local setup. I can't test the upload page because of the `No configuration setting found for key 'cognito.roleArn'` error, but the steps leading up to it work, and the CSP isn't blocking anything.

MancunianSam (approved):
LGTM

PR #566, opened 2021-04-12
Show partially complete error

MancunianSam:
- Make consignment reference not optional
- Bump to latest version of generated graphql
- Add upload in progress page and messages
- Add new endpoint and route
- Add consigment status service and tests
- Add in consignment status check to action
- Add new tests for the upload endpoint


test/services/ConsignmentStatusServiceSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+  private val consignmentId = UUID.fromString("e1ca3948-ee41-4e80-85e6-2123040c135d")
+  when(graphQLConfig.getClient[gcs.Data, gcs.Variables]()).thenReturn(getConsignmentStatusClient)
+
+  private val consignmentStatusService = new ConsignmentStatusService(graphQLConfig)
+
+  override def afterEach(): Unit = {
+    Mockito.reset(getConsignmentStatusClient)
+  }
+
+  "consignmentStatus" should {
```
TomJKing, line 33: Maybe add the word "function" to make it clear what is being tested: `"consignmentStatus function" should ...`

TomJKing (commented):
Looks good. A few minor comments.

I think these changes will break the e2e tests, so these will need updating as well.

Also when trying to run locally against intg I was getting the following error. Does the DB need updating?

`Unexpected response from GraphQL API: Response(Left({"data":null,"errors":[{"message":"Cannot query field 'currentStatus' on type 'Consignment'. (line 1, column 95):\nquery getConsignmentStatus($consignmentId:UUID!){getConsignment(consignmentid:$consignmentId){currentStatus{upload}}}\n                                                                                              ^","locations":[{"column":95,"line":1}]}]}),400,Bad Request ... `

app/validation/ValidatedActions.scala
```diff
@@ -25,13 +25,24 @@ abstract class ValidatedActions() extends TokenSecurity with I18nSupport {
     }
   }
 
-  def transferAgreementExistsAction(consignmentId: UUID)(f: Request[AnyContent] => Result): Action[AnyContent]
-  = secureAction.async { implicit request: Request[AnyContent] =>
+  def uploadPermitted(consignmentId: UUID)(f: Request[AnyContent] => Result): Action[AnyContent] = secureAction.async {
```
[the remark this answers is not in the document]
MancunianSam: I didn't yet because the tests in `UploadControllerSpec` cover the different paths through this method so I didn't think we needed tests for this method specifically. I'm happy to put some in but that's why I didn't originally. What do you think?

conf/messages
```diff
@@ -60,6 +60,9 @@ upload.chooseFolderLink=Choose folder
 upload.continueLink=Start upload
 uploadProgress.header=Uploading records
 uploadProgress.title=Do not close your browser window while your files are being uploaded. This could take a few minutes.
+upload.error.inProgress=Your upload was interrupted and could not be completed.
+upload.error.restart=Please return to the start to begin a new transfer
```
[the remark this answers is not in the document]
MancunianSam: Yep, I think so, I've added one in.

MancunianSam:
The API is failing because there's a PR still open for the changes to add that field. Once that's merged, this should work but I can't merge this until then.

app/validation/ValidatedActions.scala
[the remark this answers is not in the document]
TomJKing: I'd rather see separate unit tests, as don't really like relying on tests for another class to check logic from another class, can lead to issues when things get more complicated in the future.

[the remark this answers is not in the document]
MancunianSam: Yeah good point. I've added in some tests for this method and I added in a couple for the existing consignmentExists method as well. I think we're going to start using this as some kind of central validation point so the tests will be useful.

MancunianSam:
The API changes are in now so this can me merged when it's ready.

MancunianSam:
You're right about the e2e tests as well, there was one that was trying to upload twice which you can't do with this change. There's a PR here. https://github.com/nationalarchives/tdr-e2e-tests/pull/101/files

app/validation/ValidatedActions.scala
```diff
@@ -17,21 +16,32 @@ abstract class ValidatedActions() extends TokenSecurity with I18nSupport {
...
+    } yield {
+      val uploadInProgress: Boolean = status.flatMap(_.upload.map(_ == "InProgress")).getOrElse(false)
 
-    transferAgreementService.transferAgreementExists(consignmentId, request.token.bearerAccessToken) map {
-      case true => f(request)
-      case false => Redirect(routes.TransferAgreementController.transferAgreement(consignmentId))
+      if(!transferAgreementExists) {
+        Redirect(routes.TransferAgreementController.transferAgreement(consignmentId))
+      } else if(uploadInProgress) {
+        Redirect(routes.UploadController.uploadInProgress(consignmentId))
```
suzannehamilton, line 41: This is interesting - I expected us to show the same page with a different message rather than doing it as a redirect.

What's the reason for doing it this way? I haven't thought about the trade-offs in any detail, so maybe this is the best option. My initial thought is that it's a bit odd to have a separate page because do you then have to redirect _back_ if someone happens to go there after the upload has finished?

MancunianSam: The reason I did it as a separate page is because the upload page is getting quite complicated already with different bits being shown or hidden depending on the state. This make accessibility more difficult as you have to make sure the page isn't dependant on css to show or hide things and we've already got bugs with things not showing when they should so I didn't want to make it worse. 
You're right about redirecting back when the upload is complete but we can add this logic into the controllers and it should be a bit clearer what's happening rather than digging through the javascript to see which bits are shown/hidden.

suzannehamilton: Wait, wouldn't this logic be in the controller or HTML templates rather than the JavaScript and CSS?

This controller won't be any more complex if you render a different template on line 41 rather than do a redirect, and the user journeys will be easier to think about too because we don't need to add redirects on both controller actions.

MancunianSam: You're right about the html templates, ignore what I said about the css. I think I see what you mean. I thought you were talking about adding more if/else logic to the existing `upload.scala.html` which is what I was trying to avoid but do you mean rendering the `uploadInProgress.scala.html` template inside the validated actions? Like this 
```scala
if(!transferAgreementExists) {
        Redirect(routes.TransferAgreementController.transferAgreement(consignmentId))
      } else if(uploadInProgress) {
        Ok(views.html.uploadInProgress(consignmentId))
      } else {
        f(request)
      }
``` 
I've changed it to this as it makes sense I think

suzannehamilton: Great, thanks!

suzannehamilton:
I haven't fully reviewed this - I just commented on the controller because I noticed the redirect in the end-to-end test. @TomJKing are you happy with this PR? I'll let you approve it if so.

TomJKing (approved):

PR #575, opened 2021-04-16
Add sign out button

MancunianSam:
This includes the changes to get the logout to actually work as well as the changes to show the sign out button. The sign out button is subject to change but the rest of it is ready for review. It will just be a case of fixing the html/css up. I'll mark this as ready for review then but won't merge until we're happy with the sign out button changes.


npm/css-src/sass/_header.scss
```diff
@@ -15,7 +15,10 @@ $govuk-header-link: govuk-colour('white');
 .govuk-header__tna-content {
   margin: 0 auto;
   position: relative;
-  width: 290px;
+}
+
+.tna-header {
+  display: flex;
```
suzannehamilton, line 21: I'll try to have a look at this whole PR later today, but have you seen how other services like [Notify](https://www.notifications.service.gov.uk/) and [Pay](https://www.payments.service.gov.uk/) add links to the right-hand-side of the navigation using the design system? It looks like they're using `.govuk-header__navigation-item`.

MancunianSam: Yeah, I did try it but because the other elements in that header are using our own custom classes `govuk-header__tna-content` and `govuk-header__tna-logo` so I couldn't get the navigation-item element looking right. I think it's fine if everything else is design system standard but here, they're not.

app/auth/UnauthenticatedController.scala
```diff
@@ -0,0 +1,19 @@
+package auth
+
+import org.pac4j.core.profile.{CommonProfile, ProfileManager}
+import org.pac4j.play.PlayWebContext
+import org.pac4j.play.scala.{Security, SecurityComponents}
+import play.api.mvc.{AnyContent, Request}
+
+import javax.inject.Inject
+
+class UnauthenticatedController @Inject ()(val controllerComponents: SecurityComponents) extends Security[CommonProfile] {
```
suzannehamilton, line 10: I'm wondering about the name for this, because I started interpreting it as a controller for 401 errors or something. But it looks like it's for pages which don't need to authenticate the user - is that right? It sounds odd to call it "unauthenticated" when the user might actually be authenticated - it just doesn't matter whether they are or not.

I'm trying to think of alternatives. `PublicPageController`? `UnprotectedPageController`? The "page" bit isn't very nice, but it sounds even stranger without it. I wonder if anyone has any better ideas on Slack?

conf/application.intg.conf
```diff
@@ -15,6 +15,8 @@ consignmentapi.domain="https://api.tdr-integration.nationalarchives.gov.uk"
 
 environment=${ENVIRONMENT}
 
+logout.url="https://tdr-integration.nationalarchives.gov.uk/sign-out"
```
suzannehamilton, line 18: Is the base URL already an environment variable or a config value so we don't need to set this in each config file?

app/views/main.scala.html
```diff
@@ -38,6 +38,15 @@
                     @Messages("main.header")
                 </a>
             </div>
+            @if(isLoggedIn) {
+                <ul class="govuk-header__navigation govuk-header__navigation--end">
+                    <li class="govuk-header__navigation-item">
+                        <a href="/logout" class="govuk-header__link">
```
suzannehamilton, line 44: How does the /logout route work? Is it a default Pac4J thing?

I'm just wondering if we can think of some clearer routes than `/logout` and `/sign-out` because they're so similar that it's hard to work out what's going on.

npm/css-src/sass/_header.scss
suzannehamilton: I've added a comment to the card for adding the user's name so that we revisit all the CSS for the header: https://national-archives.atlassian.net/browse/TDR-796?focusedCommentId=14620

I think there's probably a simpler way to do this, but it probably makes sense to wait until we're adding the final element.

MancunianSam: Ok cool. Flexbox is pretty good. I did try to put the user name element in as well to see what it looked like and it was pretty easy to add with this css but as you say, we can wait until we need to do it.

app/auth/UnauthenticatedController.scala
MancunianSam: I see your point. `UnprotectedPageController` sounds right, I've changed it to that.

conf/application.intg.conf
MancunianSam: No, we've got specific variables in each file for auth.callback as well. There's nothing I can see in the base config I can see that we can re-use here. There aren't any environment variables in the ECS task for the base url either. We could do it but that's a bigger change with terraform changes which would be better on another ticket if you're going to do it.

app/views/main.scala.html
MancunianSam: You can change them. `/logout` is a Pac4J think but it's defined in the routes file and you can change it. What should we change it to though?

suzannehamilton: Maybe `/sign-out` as the Pac4J route, since that's the one that does the sign out action? Is the action in the SignOutController just used to render the page? If so, maybe call it `/signed-out`? I think you could work out the difference between `/sign-out` and `/signed-out`, though you could check with someone like Tom to see if they make sense to him out of context.

MancunianSam: They make sense to me. I've changed them

suzannehamilton: Cheers. Probably makes sense to rename the view too, for consistency.

MancunianSam: Yep, make sense. I've updated the entries in the messages file as well.

suzannehamilton (approved):

PR #579, opened 2021-04-20
Add cryptographic nonce

TomJKing:
keycloak-js depends on js-sha256, which currently imports two libraries using a call to eval

unsafe-eval should not be used because eval is a vector for attackers to run arbitrary code in a real user’s browser.

Using cryptographic nonce makes using unsafe-eval slightly less risky


suzannehamilton (approved):
These changes look pretty much how I expected, thanks!

PR #586, opened 2021-04-22
Don't show leave page confirmation on upload error

MancunianSam:
I've moved the pageUnloadFunction out of the class and exported it so it can be used in the error handling file and then I remove the event listener when the handleUploadError function is called.

### DIFF
--- a/npm/src/errorhandling/index.ts
+++ b/npm/src/errorhandling/index.ts
@@ -1,3 +1,5 @@
+import { pageUnloadAction } from "../upload"
+
 export class LoggedOutError extends Error {
   loginUrl: string
   constructor(loginUrl: string, message?: string) {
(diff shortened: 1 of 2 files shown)


suzannehamilton:
@techncl Would you like to review this one, since you know the most about the upload JavaScript?

techncl:
> @techncl Would you like to review this one, since you know the most about the upload JavaScript?

Sure.

techncl:
It seems as though the solution is not working, I've left a comment with images on [Jira](https://national-archives.atlassian.net/jira/software/projects/TDR/boards/6?assignee=5e5e19e83011ed0c8f8c7cc6&selectedIssue=TDR-861)

MancunianSam:
That's odd, it's working for me. Can you check the newer javascript iis loaded in your browser. If you look in the `errorhandling/index.ts` file like in the screenshot on the ticket, you should see the code to remove the event listener in there, if not, you'll have to rebuild the project/delete the cache in the browser. If it is there then we can have a  look.

techncl (approved):
I went through these steps and everything is fine now 👍 . Not sure what went wrong in the first place. This all works as expected: the prompt is removed when trying to navigate away from the page upon encountering an error and the prompt is still there when trying to navigate away from the page if you don't receive an error

MancunianSam:
Perfect :+1: thanks. It was probably just Chrome trying to be helpful last week.

PR #601, opened 2021-04-28
Show the error message instead of hiding it

MancunianSam:
This is for bug https://national-archives.atlassian.net/browse/TDR-957
The loggedOutError element is the error we're supposed to be showing to the user but it's being hidden instead of shown. I've changed this.


TomJKing (approved):
Might want to double check this doesn't break the e2e tests.

MancunianSam:
I think it's ok. There doesn't seem to be any tests checking for it.

PR #615, opened 2021-05-04
File status checks failed page

MancunianSam:
I've changed this now so there's only one records result page and it renders success or error depending on the response from the API. This has made it a lot simpler.


test/controllers/FileChecksResultsControllerSpec.scala
```diff
@@ -1,15 +1,20 @@
 package controllers
 
 import java.util.UUID
-
 import com.github.tomakehurst.wiremock.WireMockServer
-import com.github.tomakehurst.wiremock.client.WireMock.{okJson, post, urlEqualTo}
+import com.github.tomakehurst.wiremock.client.WireMock.{equalToJson, okJson, post, urlEqualTo}
```
TomJKing, line 5: Is the `equalToJson` import needed?

TomJKing (commented):
Looks good. One very minor comment. Also did you check this change doesn't break the e2e tests?

test/controllers/FileChecksResultsControllerSpec.scala
MancunianSam: No, I've removed it.

MancunianSam:
There are changes to the e2e tests on this https://github.com/nationalarchives/tdr-e2e-tests/pull/104 PR which you've already approved. I ran them against this branch and they were ok.

TomJKing (approved):

PR #616, opened 2021-05-04
Display message if submission without folder selection

techncl:


npm/test/upload-form.test.ts
```diff
@@ -297,6 +309,43 @@ class MockDom {
...
+  await mockDom.form.handleFormSubmission(submitEvent)
+
+  mockDom.submitAndLabelButtons.forEach((button) =>
+    expect(button).not.toHaveAttribute("disabled", "true")
+  )
+
+  expect(mockDom.hiddenInputButton).not.toHaveAttribute("disabled", "true")
+})
+
+test("Clicking the Submit button, without selecting a folder, displays an warning message to the user", async () => {
```
suzannehamilton, line 325: Typo: "an warning"

app/views/upload.scala.html
```diff
@@ -43,15 +43,20 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
...
                             </dd>
-                            <dd id="drag-and-drop-failure" class="govuk-summary-list__value drag-and-drop__failure" hidden tabindex="-1" role="alert" aria-describedby="failureMessageText">
-                                @redWarningSign()
-                              <p id="failureMessageText">@Messages("upload.dragAndDropErrorMessage")</p>
+                            <dd id="drag-and-drop-failure" class="govuk-summary-list__value drag-and-drop__failure" hidden
+                                tabindex="-1" role="alert" aria-describedby="failure-message-text">
+                                <div id="failure-message-text">
+                                    @redWarningSign()
+                                    <p id="non-folder-selected-message-text" hidden>@Messages("upload.dragAndDropErrorMessage")</p>
+                                    <p id="submission-without-a-folder-message-text" hidden>@Messages("upload.submissionWithoutAFolderMessageText")</p>
```
suzannehamilton, line 58: This is a little bit tricky to follow, because the JavaScript has to show and hide the failure container and the individual messages. It's also slightly confusing because the `submission-without-a-folder-message-text` element isn't solely related to the drag-and-drop, but it's nested in an element that sounds drag-and-drop specific.

What about having two separate `<dd>` elements so that you can just show and hide those individually? They'll be very similar, but you could extract a sub-view to commonise the code.

suzannehamilton (commented):
This look pretty sensible. Just a suggestion for a refactoring which I think will make the code a little easier to follow.

app/views/upload.scala.html
techncl: This is a really good idea! Thanks.

suzannehamilton (commented):
The new sub-view looks good. 👍 Just one question about the tests.

npm/test/upload-form.test.ts
```diff
@@ -82,22 +82,37 @@ class MockDom {
...
+                                          <path d="M25,6.2L8.7,23.2L0,14.1l4-4.2l4.7,4.9L21,2L25,6.2z"></path>
+                                       </svg>
+                                       <p id="success-message-text">The folder "<span id="folder-name"></span>" (containing <span id="folder-size"></span>) has been selected</p>
+                                    </div>
+                                 </dd>
+                                 <dd id="folder-selection-failure" class="govuk-summary-list__value drag-and-drop__failure" hidden=""
+                                    tabindex="-1" role="alert" aria-describedby="non-folder-selected-message-text">
+                                    <div>
+                                       <span class="drag-and-drop__error">
                                           <svg class="green-tick-mark" role="presentation" focusable="false" viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg" width="25px" height="25px">
```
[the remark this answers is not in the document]
techncl: Ah yes, it's the real view but it's due to a mistake in the view; I'll make the class name more generic.

suzannehamilton (approved):

PR #626, opened 2021-05-07
Remove unused div with duplicate id from file check failure page

MancunianSam:
Lighthouse in Chrome reported that there were two divs with identical ids which there were. But the div isn't needed, I've removed it and the behaviour doesn't change and the lighthouse report now shows 100% for
accessibility.


TomJKing (approved):

techncl:
I noticed that there was still a reference to the div in the tests.

PR #628, opened 2021-05-07
Add attribute to button in order to ignore clicks made within 1 sec

techncl:


app/views/transferSummary.scala.html
```diff
@@ -76,7 +76,8 @@ <h1 class="govuk-heading-l">@Messages("transferSummary.title")</h1>
                 </fieldset>
                 <div>
                     <!-- Transfer -->
-                    <button class="govuk-button" type="submit" data-module="govuk-button" role="button">
+                    @* GOV.UK's data-prevent-double-click attribute added to ignore clicks made within 1 second *@
```
suzannehamilton, line 79: I think this comment might not be necessary, since "data-prevent-double-click" is pretty self-explanatory. And if you google it, the first hit is for the right page in the docs.

suzannehamilton (approved):

PR #633, opened 2021-05-10
Clear selected files on error.

MancunianSam:
I've put the code to clear the files in the rejectUserItemSelection function as this is only called when there's an error.


techncl (approved):
Looks good!

PR #642, opened 2021-05-13
Updated Sign out page

plembke-tna:
Updated title and added paragraph and turned back link into link


conf/messages
```diff
@@ -144,5 +144,6 @@ transferSummary.progress=5
 progress.totalSteps=5
 
 signedOut.header=Signed out
-signedOut.title=You have signed out
+signedOut.title=You have successfully signed out
+signedOut.paragraph=Thanks for using the Transfer Digital Records service.
```
suzannehamilton, line 148: It would be good to give this a more descriptive name than "paragraph" so its purpose is clearer in the template. What about `signedOut.thankYouMessage`?

plembke-tna: Sounds good!

suzannehamilton (approved):

PR #645, opened 2021-05-14
Check folder has files before saving it

techncl:


suzannehamilton:
This PR looks like it fixes a bug, but I'm not clear from the commit message what the bug is. Would you mind adding an explanation to the commit message? And are they two different bugs, or should the commits be squashed?

Give me a shout if you'd like a hand with git rebase!

techncl:
> This PR looks like it fixes a bug, but I'm not clear from the commit message what the bug is. Would you mind adding an explanation to the commit message? And are they two different bugs, or should the commits be squashed?
> 
> Give me a shout if you'd like a hand with git rebase!

Hmm, they _are_ different and very minor that's why I put them on the same branch. With the first one, at line 159, a folder could be full or empty. It was: first, assigning the (full or empty) folder, checking whether the folder was full or empty and then assigning an empty folder if it isn't, so if the folder was empty, it would be assigning the empty folder twice (which is redundant). The second one was added because I noticed the dragover shadow got removed too late, it's not that noticeable but thought since I was there, I could just move it up slightly. Should I squash these, change the branch name to something more generic or both?

suzannehamilton:
If they're separate bugs, it's worth having separate commits. 👍 

It would be good to add a bit more detail to the commit messages to make that clear, though. You can do it with interactive rebase - give me a shout if you'd like a hand.

suzannehamilton (approved):
Thanks! 🙌

PR #648, opened 2021-05-17
Update versioning strategy to always bump package.json

MancunianSam:


TomJKing (approved):

.github/dependabot.yml
```diff
@@ -4,6 +4,7 @@ updates:
     directory: "/npm"
     schedule:
       interval: "daily"
+    versioning-strategy: increase
```
suzannehamilton, line 7: Do we need to do this in any of the other projects?

MancunianSam: Yep. I've already done the other ones

suzannehamilton: 👍

PR #650, opened 2021-05-18
Replace buttons with links

MancunianSam:
The "Return to start" buttons on these pages aren't part of a form and don't do anything when clicked. They should be links.


suzannehamilton (commented):
I know she's not in today, but it would be good if @plembke-tna could take a look at this one? She's been looking at the other buttons.

plembke-tna:
Yep, Sam is right.

plembke-tna (approved):
Looks good

PR #661, opened 2021-05-21
Relocated messages into views

plembke-tna:
Also updated the wording from the newest UI review as noted in TDR-1135.


suzannehamilton:
Thanks for doing this! 🙌 

Given that this is such a fundamental change, could you add a bit more detail to the commit message explaining the reason for it? You can use `git commit --amend` to edit the most recent commit message - give me a shout if you'd like a hand.

Also, it looks like a few tests are failing because they're expecting the old message keys rather than the actual text.

app/views/notFoundError.scala.html
```diff
@@ -1,14 +1,14 @@
...
     <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-l">@Messages("notFoundError.header")</h1>
-        <p class="govuk-body">@Messages("notFoundError.body1")</p>
-        <p class="govuk-body">@Messages("notFoundError.body2")</p>
-        <p class="govuk-body">@Html(Messages("notFoundError.body3"))</p>
-        <p class="govuk-body">@Html(Messages("notFoundError.body4"))</p>
+        <h1 class="govuk-heading-l">Page not found</h1>
+        <p class="govuk-body">If you typed a web address, check it is correct.</p>
+        <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+        <p class="govuk-body">@Html("If the web address is correct, please <a href= \"/contact\">contact<a/> us.")</p>
```
suzannehamilton, line 10: I don't think we need the `@Html` wrapper now that the text is in the view rather than the messages file.

app/views/unauthenticatedError.scala.html
```diff
@@ -1,11 +1,11 @@
...
 
-@main(Messages("401.error.header")) {
+@main("Sign-in required") {
 @defining(play.core.PlayVersion.current) { version =>
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-l">@Messages("unauthenticatedError.header")</h1>
-        <p class="govuk-body">@Html(Messages("unauthenticatedError.body"))</p>
+        <h1 class="govuk-heading-l">You must be signed in to see this page</h1>
+        <p class="govuk-body">@Html("<a href= \"#\">Sign in</a> to continue.")</p>
```
suzannehamilton, line 8: Another `@Html`.

app/views/upload.scala.html
```diff
@@ -65,7 +65,7 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
                                 tabindex="-1" role="alert" aria-describedby="success-message-text">
                                 <div>
                                     @greenTickMark()
-                                    <p id="success-message-text" >@Html(Messages("upload.dragAndDropSuccessMessage"))</p>
+                                    <p id="success-message-text" >@Html("The folder \"<span id=\"folder-name\"></span>\" (containing <span id=\"folder-size\"></span>) has been selected")</p>
```
suzannehamilton, line 68: Another `@Html`.

app/views/upload.scala.html
```diff
@@ -112,19 +112,19 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
...
-                    @Messages("upload.error.title")
+                    There is a problem
                 </h2>
                 <div class="govuk-error-summary__body">
-                    <p>@Messages("upload.error.failure")</p>
-                    <p class="upload-progress-error-timeout__message" hidden>@Html(Messages("upload.progress.error.timeout"))</p>
-                    <p class="upload-progress-error-authentication__message" hidden>@Html(Messages("upload.progress.error.authentication"))</p>
-                    <p class="upload-progress-error-general__message" hidden>@Html(Messages("upload.progress.error.general"))</p>
+                    <p>Some or all of your files failed to upload.</p>
+                    <p class="upload-progress-error-timeout__message" hidden>@Html("Your upload has timed out, click <a href=\"/series\">series</a> to start again.")</p>
```
suzannehamilton, line 125: Another `@Html`.

app/views/notFoundError.scala.html
suzannehamilton, line 10: Now that these links aren't in the messages file, we can generate them in a more reliable way using `@routes`. Here's an example: https://github.com/nationalarchives/tdr-transfer-frontend/blob/c8a0c418f4060a010bc77deaadabb593b6ab8b78/app/views/fileChecksResults.scala.html#L25

This means that if we change the URL of any of the pages then the links will be updated automatically.

app/views/upload.scala.html
suzannehamilton, line 125: Another set of links that could use `@routes`.

suzannehamilton (commented):
Looks good! Just a couple of suggestions to make better use of the Play view templates.

app/views/upload.scala.html
```diff
@@ -65,7 +65,7 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
                                 tabindex="-1" role="alert" aria-describedby="success-message-text">
                                 <div>
                                     @greenTickMark()
-                                    <p id="success-message-text" >@Html(Messages("upload.dragAndDropSuccessMessage"))</p>
+                                    <p id="success-message-text">The folder "<span id="folder-name"></span>" (containing <span id="folder-size"></span> has been selected</p>
```
suzannehamilton, line 68: Looks like there's a missing closing bracket `)` after the closing `</span>`.

suzannehamilton (commented):
Looks good! Just one typo to fix.

suzannehamilton:
Just for completeness, here's the PR for adding an Architecture Decision Record about this change: https://github.com/nationalarchives/tdr-dev-documentation/pull/116

app/views/upload.scala.html
plembke-tna: Done!

suzannehamilton (approved):

PR #680, opened 2021-06-03
Accessibility testing upload page

rsulxo:


TomJKing:
Not sure if this is meant to be ready for review, but these changes have affected the unit tests, so these will need updating.

npm/src/clientfileprocessing/index.ts
```diff
@@ -25,11 +25,26 @@ export class ClientFileProcessing {
...
+    const progressLabelElement: HTMLDivElement | null =
+      document.querySelector(".progress-label")
+
+    if (fileUpload && progressBar) {
+      fileUpload.setAttribute("hidden", "true")
+      progressBar.removeAttribute("hidden")
+    }
 
-    if (progressBarElement) {
+    if (progressBarElement && progressLabelElement) {
```
TomJKing, line 43: This is the same code as at line 60. It might be worth abstracting this out into a separate function:

```
renderWeightedPercent(weightedPercent: number) {
    const progressBarElement: HTMLDivElement | null =
        document.querySelector(".progress-display")
    const progressLabelElement: HTMLDivElement | null =
        document.querySelector(".progress-label")

    if (progressBarElement && progressLabelElement) {
      var stringWeightedPercentage = weightedPercent.toString()
      progressBarElement.setAttribute("value", weightedPercent.toString())
      progressLabelElement.innerText =
          "Uploading records " + stringWeightedPercentage + "%"
    }    
  }
```
Then you can calculate the weighted percent value in the calling function, and pass it into the new function

npm/test/clientfileprocessing.test.ts
```diff
@@ -526,4 +527,4 @@ function checkNoPageStateChangeExpected() {
   expect(fileUpload && fileUpload.classList.toString()).toEqual(
     "govuk-grid-row"
   )
-}
+}
```
TomJKing, line 530: Missing line break at end of file

TomJKing (commented):
Looks good. Just a couple of comments.

PR #681, opened 2021-06-03
Added consignment reference on confirmation page

plembke-tna:
- added span with reference in view
- added request in consignmentService
- updated controller
- updated test


app/controllers/TransferCompleteController.scala
```diff
@@ -6,13 +6,29 @@ import configuration.KeycloakConfiguration
...
+                                           consignmentService: ConsignmentService)
+                                          (implicit val ec: ExecutionContext) extends TokenSecurity with I18nSupport {
+
+  private def getConsignmentReference(request: Request[AnyContent], consignmentId: UUID)
+                                   (implicit requestHeader: RequestHeader): Future[ConsignmentReferenceData] = {
+    consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
+      .map(r => ConsignmentReferenceData(r.consignmentReference))
+  }
 
   //noinspection ScalaUnusedSymbol
```
TomJKing, line 25: Is this noinspection comment needed?

app/controllers/TransferCompleteController.scala
```diff
@@ -6,13 +6,29 @@ import configuration.KeycloakConfiguration
 import javax.inject.Inject
```
TomJKing, line 6: Needs a line space between `package controllers` and `import java.util.UUID` statements at the beginning of the file

app/controllers/TransferCompleteController.scala
```diff
@@ -6,13 +6,29 @@ import configuration.KeycloakConfiguration
...
-    Ok(views.html.transferComplete())
+  def transferComplete(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    getConsignmentReference(request, consignmentId)
+      .map { consignmentReference =>
+        Ok(views.html.transferComplete(consignmentReference))
+      }
   }
 }
+
+case class ConsignmentReferenceData(consignmentReference: String)
```
TomJKing, line 34: Is there is a reason for using a case class, rather than just returning the consignment reference directly in the `getConsignmentReference` function? For example:

```
private def getConsignmentReference(request: Request[AnyContent], consignmentId: UUID)
                                   (implicit requestHeader: RequestHeader): Future[String] = {
    consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
      .map(r => r.consignmentReference)
  }
```

app/services/ConsignmentService.scala
```diff
@@ -62,4 +64,11 @@ class ConsignmentService @Inject()(val graphqlConfiguration: GraphQLConfiguratio
     sendApiRequest(getConsignmentSummaryClient, getConsignmentSummary.document, token, variables)
       .map(data => data.getConsignment.get)
   }
+
+  def getConsignmentRef(consignmentId: UUID, token: BearerAccessToken): Future[getConsignmentReference.GetConsignment] = {
```
TomJKing, line 68: Is this function needed? It looks like the existing `getConsignmentTransferSummary` function will return the consignment reference value as part of its response. That would save creating the new function and the new client: `getConsignmentReferenceClient`

app/views/transferComplete.scala.html
```diff
@@ -1,4 +1,4 @@
-@()(implicit messages: Messages, request: RequestHeader)
+@(consignmentReferenceData: ConsignmentReferenceData)(implicit messages: Messages, request: RequestHeader)
```
TomJKing, line 1: Related to my comment about, whether this needs to a case class, instead of just passing a string directly.

app/views/transferComplete.scala.html
```diff
@@ -8,8 +8,11 @@
             <h1 class="govuk-panel__title">
                 @Messages("transferComplete.title")
             </h1>
-            <div class="govuk-panel__body">
-                @Messages("transferComplete.body")
+            <div class="govuk-panel__body govuk-!-font-size-27">
+                <span>@Messages("transferComplete.body")</span><br/><br/>
```
TomJKing, line 12: Is it good practice to use the `<br/>` HTML element to create line spaces? Not sure

TomJKing (commented):
Just a few questions. Wasn't sure whether a new function is needed to retrieve the consignment reference when there is an existing function that already gets it. It might be that the existing function cannot be used at this point in the workflow.

app/controllers/TransferCompleteController.scala
plembke-tna: I wasn't sure so I left it.. Shall I remove it?

app/views/transferComplete.scala.html
plembke-tna: Thanks for reminding me. It's not good practice. The reason behind this was that using the paragraph would mean having to create CSS just for this which is something I wanted to avoid. But I am just working out a different solution.

plembke-tna: Worked out a better way and will change it :)

app/controllers/TransferCompleteController.scala
plembke-tna: @MancunianSam can this be removed?

app/services/ConsignmentService.scala
plembke-tna: Good point. We discussed this with Suzanne and doing it this way will make it slightly quicker.

TomJKing: OK. Fair enough.

TomJKing (approved):

PR #699, opened 2021-06-14
Change the file checks continue button from link to a button.

MancunianSam:
The button needs to be disabled when the file checks are in progress and you can't disable a link without doing some hacky things like removing the href so I've converted this to a form with a button and added the disabled attribute. I've updated the js to remove the disabled attribute and changed the tests to check.


suzannehamilton:
If you don't mind waiting a day, it might be useful for @plembke-tna to take a look at this one since she's put some thought into how we're using links and buttons elsewhere.

plembke-tna:
I've tested and researched a bit and I agree with Sam that it gets very hacky when it comes to disabling a link. Although we don't actually have a form here, I think it is probably be the best solution to my knowledge now. I can however put some more time into researching for another solution but I think we should go ahead with this for now.

suzannehamilton:
Was there any more discussion in the design phase about whether we should even show the disabled button at all? I know someone mentioned that there are accessibility problems with disabled buttons because you're using colour to indicate the state change. Did you decide that you still preferred it to the other option of not showing the button at all until the file checks are complete?

rsulxo:
@suzannehamilton From what i recall, when you disable a button using the design system it doesn't change the colour per say, it simply makes the button look faded (but still green) if that makes sense. So essentially it shouldn't effect any accessibility needs. Also, from user research and testing it was a better choice to disable the button rather than hiding/showing it. Hope this clears it up a bit better.

plembke-tna (approved):

PR #703, opened 2021-06-16
Added paragraph with link to FAQ on file checks page

plembke-tna:
-added paragraph with link to FAQ
-removed spacing underneath the list that was needed without the paragraph
-added govuk-body class to paragraphs in success summary


app/views/fileChecksProgress.scala.html
```diff
@@ -13,11 +13,15 @@ <h1 class="govuk-heading-l">Checking your records</h1>
...
             <p class="govuk-body">Please wait while your records are being checked. This may take a few minutes.</p>
             <p class="govuk-body">The following checks are now being performed:</p>
-            <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-7">
+            <ul class="govuk-list govuk-list--bullet">
                 <li>Anti-virus scanning</li>
                 <li>Identifying file formats</li>
                 <li>Validating data integrity</li>
             </ul>
+            <p class="govuk-body govuk-!-margin-bottom-7">For more information on these checks, please see our
+                <a href="@routes.FaqController.faq()" class="govuk-link">FAQ</a> to this service.
```
suzannehamilton, line 22: Should this be "for this service" rather than "to this service"?

plembke-tna: You probably know that better than I do. :) I have used Ashley's copy but I'm happy to change it. Just let me know.

suzannehamilton: Sarra and Tom agree, so I think we should change it: https://tna-digital.slack.com/archives/CJYJYG54L/p1623999024060600

suzannehamilton (approved):

PR #718, opened 2021-06-24
Added link to survey on transfer complete page

plembke-tna:
- added a paragraph with a link to the satisfaction survey (linking to SmartSurvey) that opens in a new tab


app/views/transferComplete.scala.html
```diff
@@ -23,6 +23,12 @@ <h2 class="govuk-heading-m govuk-!-margin-top-2">What happens next</h2>
 
                     <p class="govuk-body">We will contact you via email within 90 days. If you do not receive an email, please contact us.</p>
 
+                    <p class="govuk-body govuk-!-margin-bottom-6">
+                        <a href="https://www.smartsurvey.co.uk/s/tdr-feedback/" class="govuk-link" rel="noreferrer noopener" target="_blank" title="What did you think of this service? (opens in new tab)">
```
suzannehamilton, line 27: What's the purpose of this `title` attribute? Particularly when it's identical to the link text, which means it doesn't add anything extra.

https://www.w3.org/TR/WCAG20-TECHS/H33.html suggests using `title` when the link text itself doesn't give enough context, and it also says "authors should use caution in applying this technique".

plembke-tna: I was talking to Chris about this and he said it needs to be added. But now that you pointed out that the purpose is more context, I think he meant adding it to say it opens up in a new tab. But I put that in the link itself so it is redundant and I will remove it.

suzannehamilton (approved):

PR #726, opened 2021-06-28
Add code to check if upload status is 'Completed' & page in the case …

techncl:
…it is


app/validation/ValidatedActions.scala
```diff
@@ -26,22 +26,24 @@ abstract class ValidatedActions() extends TokenSecurity with I18nSupport {
...
-      } else if(uploadInProgress) {
-        Ok(views.html.uploadInProgress(consignmentId))
-      } else {
-        f(request)
+        if (!transferAgreementExists) {
+          Redirect(routes.TransferAgreementController.transferAgreement(consignmentId))
+        } else if (uploadStatus == "InProgress") {
+          Ok(views.html.uploadInProgress(consignmentId))
+        } else if (uploadStatus == "Completed") {
+          Ok(views.html.uploadHasCompleted(consignmentId))
```
suzannehamilton, line 43: Could you add a controller test for this case?

app/views/uploadHasCompleted.scala.html
```diff
@@ -0,0 +1,19 @@
...
+@(consignmentId: UUID)(implicit request: RequestHeader, messages: Messages)
+
+@main("Upload your records") {
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      @progressIndicator(Messages("upload.progress"))
+      <h1 class="govuk-heading-l">Uploading records</h1>
+
+      <div class="govuk-grid-column-two-thirds">
+        <p>Your upload is complete and has been saved. You cannot make amendments to your upload or add additional files</p>
```
suzannehamilton, line 11: Should there be a full stop at the end of the second sentence?

app/views/uploadHasCompleted.scala.html
```diff
@@ -0,0 +1,19 @@
+@import java.util.UUID
+@(consignmentId: UUID)(implicit request: RequestHeader, messages: Messages)
+
+@main("Upload your records") {
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      @progressIndicator(Messages("upload.progress"))
```
suzannehamilton, line 7: Should the progress bar be here? I wouldn't have expected to see it here, but I'm not sure of the design.

suzannehamilton (commented):
Just a few comments, thanks!

app/views/uploadHasCompleted.scala.html
techncl: Oh no, this is calling the [progress indicator ](https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/views/progressIndicator.scala.html) file, the one that shows which step you are on out of N steps

suzannehamilton: You're quite right. Ignore me. 🙂

suzannehamilton (approved):
🚀

PR #727, opened 2021-06-28
FAQ links open in new tab and on the line required

rsulxo:
This PR is for the FAQ links to open in a new tab so they don't disrupt the user journey. They also jump to the line required by the user.


app/views/faq.scala.html
```diff
@@ -70,7 +70,7 @@ <h3 class="govuk-heading-s">How do I know if my records are Crown copyright?</h3
                 Works created by officers or servants of the crown during the course of their duties are Crown copyright. This includes material created by ministers, civil servants, government departments and agencies. The National Archives will accept Public Records that are not Crown copyright, however during this initial private beta stage, you cannot currently use TDR to transfer them to us.
             </p>
 
-            <h3 class="govuk-heading-s">The records I wish to transfer are in another language (for example, Welsh)</h3>
+            <h3 id="other-transfers" class="govuk-heading-s">The records I wish to transfer are in another language (for example, Welsh)</h3>
```
suzannehamilton, line 73: Should this be something like `other-languages`? It looks like this section is just about language rather than all other transfers.

suzannehamilton (approved):

PR #735, opened 2021-06-30
Make upload controller spec tests more robust

techncl:


test/controllers/UploadControllerSpec.scala
```diff
@@ -115,9 +119,15 @@ class UploadControllerSpec extends FrontEndTestHelper {
...
-      "\"variables\":{\"consignmentId\":\"c2efd3e6-6664-4582-8c28-dcf891f60e68\"}}"
+    val body = """{"query":"query isTransferAgreementComplete($consignmentId:UUID!){
+                 |                getTransferAgreement(consignmentid:$consignmentId){
+                 |                  isAgreementComplete
+                 |                }
+                 |          }",
+                 |         "variables":{
+                 |            "consignmentId":"c2efd3e6-6664-4582-8c28-dcf891f60e68"
+                 |          }
+                 |}""".stripMargin.replaceAll("\n\\s*", "")
```
suzannehamilton, line 130: What happens if you don't do the `replaceAll`? Does the request not match because it's checking for the exact whitespace?

It might be nice to write our own little custom request matcher that does all the string wrangling, so you don't have to repeat it every test: http://wiremock.org/docs/extending-wiremock/#custom-request-matchers

Then you could change the Wiremock lines below to something like:

```
  .withRequestBody(equalToGraphQlQuery(body))
```

techncl: I had a look into this but I didn't quite understand how to implement it (the custom request matcher). There also doesn't seem to be a lot of documentation outside this page to further explain how to do it. 

I also just noticed that the `.stripmargin` is not even necessary (and therefore, the vertical bars too), so it only needs the part that replaces all new lines and spaces. Should I just write a method to do that? If so, should I put this method in some sort of helper file or have one per spec test (similar to the stub response methods)?

suzannehamilton: Yeah, writing a method that does the formatting sounds good. You can give it a name to show what it does so people won't have to figure out the logic from scratch.

suzannehamilton:
Looks like the Jenkins build has failed.

suzannehamilton (approved):

PR #743, opened 2021-07-06
Refactor file checks code

techncl:


npm/src/filechecks/verify-checks-completed-and-display-banner.ts
```diff
@@ -0,0 +1,35 @@
+import { IFileCheckProgress } from "./get-file-check-progress"
```
suzannehamilton, line 1: The name of this file is a combination of the two methods that are in it. What if we add a third method? 🙂 

What about calling it `completed-checks.ts` as a summary? Or something similar - something that's easy to interpret when you see it in the file list.

npm/src/filechecks/get-file-check-progress.ts
```diff
@@ -34,33 +33,31 @@ export const getFileChecksInfo: (
...
-        })
+        return {
+          antivirusProcessed: fileChecks.antivirusProgress.filesProcessed,
+          checksumProcessed: fileChecks.checksumProgress.filesProcessed,
+          ffidProcessed: fileChecks.ffidProgress.filesProcessed,
+          totalFiles: getConsignment.totalFiles
+        }
       } else {
         console.log(
           `No progress metadata found for consignment ${consignmentId}`
```
suzannehamilton, line 55: Does this case ever actually happen? i.e. there are no errors in the GraphQL response, but nothing in the `data` response either. I wonder if it would be better to throw an error instead? Then you could get simplify the `IFileCheckProgress | null` return type.

npm/src/filechecks/verify-checks-completed-and-display-banner.ts
suzannehamilton: Actually, thinking about this more, would it be better to split this up into two separate files?

It's useful for modules to be at the same level of abstraction - it makes their purpose easier to follow when you read them. Whereas this knows about file check result logic _and_ about updating the DOM.

npm/test/getfilecheckprogress.test.ts
```diff
@@ -59,56 +56,71 @@ class GraphqlClientFailure {
   }
 }
 
-test("getFileChecksInfo returns the correct consignment data with a successful api call", (done) => {
+class GraphqlClientErrors {
+  mutation: (
+    query: DocumentNode,
+    variables: GetFileCheckProgressQueryVariables
+  ) => Promise<ExecutionResult> = async (_, __) => {
```
suzannehamilton, line 63: Ooh, I've never seen this double underscore in TypeScript before. What does it do?

npm/test/getfilecheckprogress.test.ts
```diff
@@ -59,56 +56,71 @@ class GraphqlClientFailure {
...
-      return false
-    }
-  getFileChecksInfo(client, callback)
+  const fileChecksProgress: IFileCheckProgress | null =
+    await getFileChecksProgress(client)
+
+  expect(fileChecksProgress).toBeNull()
+})
+
+test("getFileChecksProgress returns null with errors from a successful api call", async () => {
```
suzannehamilton, line 113: What case does this test cover - if the API call was a success, why should it return null? And what does "null with errors" mean?

npm/test/updatefilecheckprogress.test.ts
```diff
@@ -0,0 +1,148 @@
+const mockFileCheckProcessing = {
```
suzannehamilton, line 1: It looks like you're copying what's already there, but should we be formatting the test names in the same way as the main TypeScript file names? The file names use kebab-case, whereas in the tests we're puttingeverythingintooneword.

npm/test/verifycheckscompletedanddisplaybanner.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    expect(notificationBanner.getAttribute("hidden")).toBeNull()
+    expect(
+      continueButton.classList.contains("govuk-button--disabled")
+    ).toBeFalsy()
+    expect(continueButton.hasAttribute("disabled")).toBeFalsy()
+  }
+
+})
+
+test("displayChecksCompletedBanner displays banner but can't unhide the button", () => {
```
suzannehamilton, line 63: The phrasing "but can't unhide the button" makes it sound like the test is checking that something isn't possible. What about "even if button is missing"?

npm/test/verifycheckscompletedanddisplaybanner.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  displayChecksCompletedBanner()
+
+  const notificationBanner = document.querySelector(
+    "#file-checks-completed-banner"
+  )
+  const continueButton = document.querySelector("#file-checks-continue")
+  expect(continueButton).toBeNull()
+  expect(notificationBanner).not.toBeNull()
+  //Keep typescript happy
+  if (notificationBanner) {
```
suzannehamilton, line 74: It's good to avoid `if` statements in tests. If someone breaks the HTML fixture on line 64, this test would still pass because the `if` evaluates to `true`.

Better to do something that will result in an a failed test or an error, so that we spot the problem. I not familiar enough with TypeScript to know what the idiomatic thing to do is, but can you use the exclamation mark syntax instead?

```
expect(notificationBanner!.getAttribute("hidden")).toBeNull()
```

npm/test/verifycheckscompletedanddisplaybanner.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    //Hidden attribute evaluates to empty string in the tests if not removed.
+    expect(notificationBanner.getAttribute("hidden")).toBeNull()
+  }
+})
+
+test("displayChecksCompletedBanner displays banner but can't unhide the button", () => {
```
suzannehamilton, line 80: Same name as the test above?

suzannehamilton (commented):
Thanks for doing this! A few suggestions for the tests and file names.

npm/src/filechecks/verify-checks-completed-and-display-banner.ts
techncl: Ok, sure. Yes, I was thinking that the name was too long but I wanted to name it in a way that you instantly knew what it was/did without having to open the file. but yes, splitting it up into smaller files is an even better solution.

npm/src/filechecks/get-file-check-progress.ts
techncl: You're referring to Line 58 (in the old version, Line 55 in the new)? I'm not sure, Sam added this. @MancunianSam do you know of a case?

@suzannehamilton but even if we don't return null here, the `catch` further down does (in case the response fails). Also, I don't know how hard it would make it to test for exceptions; I tried but I couldn't get it to work, but I could have been doing something wrong.

npm/test/getfilecheckprogress.test.ts
techncl: Ah, it's used here the same way as a single underscore i.e. unused/insignificant parameter, it's just that you can not do `(_, _) => foo` like in Scala because the underscores here are just regular parameter names (which have to be unique). I.e. because Javascript allows you to have underscores for variable/parameter names, we just used them to signify these parameters will not be used since we are mocking the response.

npm/test/getfilecheckprogress.test.ts
techncl: Because I realised that we don't have a test [for this](https://github.com/nationalarchives/tdr-transfer-frontend/blob/c83056108f273cc982e6d82df7b47e4f1c94d606/npm/src/filechecks/get-file-check-progress.ts#L37), which I thought was a large omission since it would test the case in which the GraphQL call was successful but it returned errors such as "the user does not have access to this consignment". The only tests we had were

1. GraphQL response success, response has data - the code in first `else`
2. GraphQL response failure - the code in `catch`
3. GraphQL response success, response has no data - the code in nested `else`; the one that you asked about
but no "GraphQL response success, response has errors" - the code in `if` statement

> And what does "null with errors" mean

I just copied the same format as the previous test (maybe we can reword it?). This basically means the function returns null after the successful graphql response had errors in it

npm/test/updatefilecheckprogress.test.ts
techncl: Yes, I was just following convention but I can add give it the same/similar name. should I do it just for the file checks files and then create another PR where I rename the others?

suzannehamilton:
Looks like there are test failures on Jenkins.

npm/test/updatefilecheckprogress.test.ts
suzannehamilton: If you could rename any that you're working with directly, that would be great. We can update the other ones as we go if it's too much effort to rename them all!

npm/src/filechecks/get-file-check-progress.ts
suzannehamilton: Do you want to have a call about the exception handling to see if we can get something to work?

techncl: Sure!

npm/test/get-file-check-progress.test.ts
```diff
@@ -43,10 +40,7 @@ class GraphqlClientEmptyData {
     query: DocumentNode,
     variables: GetFileCheckProgressQueryVariables
   ) => Promise<FetchResult<GetFileCheckProgressQuery>> = async (_, __) => {
-    const data: GetFileCheckProgressQuery = {
-      getConsignment: null
-    }
-    return { data }
+    throw new Error("No progress metadata found for consignment")
```
suzannehamilton, line 43: Is this the right test setup? The class is called `GraphqlClientEmptyData`.

npm/test/get-file-check-progress.test.ts
```diff
@@ -55,60 +49,66 @@ class GraphqlClientFailure {
     query: DocumentNode,
     variables: GetFileCheckProgressQueryVariables
   ) => Promise<FetchResult<GetFileCheckProgressQuery>> = async (_, __) => {
-    return Promise.reject("error")
+    throw new Error("error")
```
suzannehamilton, line 52: Should this one still be a Promise, so that it simulates an async API call? Or an awaited Promise? I forget what combinations we tried yesterday, sorry. 🙂

suzannehamilton (commented):
Just a couple of small comments. 👍

npm/test/get-file-check-progress.test.ts
techncl: This is testing the case where [getConsignment](https://github.com/nationalarchives/tdr-transfer-frontend/blob/9273ee451f6652309835b6994757d5408d8a0ada/npm/src/filechecks/get-file-check-progress.ts#L55) is empty. Should I maybe rename it to `GraphqlClientEmptyGetConsignment`?

npm/test/get-file-check-progress.test.ts
techncl: Hmm, yes, I think it should still be a Promise so that is simulates an async API call and so that it's consistent with the other client mocks; the result of the call is being awaited in the test

npm/test/get-file-check-progress.test.ts
suzannehamilton: Ah, I see. This case is now pretty much the same as `GraphqlClientFailure` then, because they both throw and `Error`. So I don't think you need two separate test cases any more.

techncl: But they generate different errors though; in the case of a graphql client failure (this was the one where we removed the `catch` block), now it just throws an exception and in the case of empty data, we are throwing our own error with this message `No progress metadata found for consignment ${consignmentId}`. Should I still remove the test for the `GraphqlClientFailure`?

techncl: After thinking about it and I came to the conclusion that it probably does make sense to have them separate, as this `GraphqlClientEmptyGetConsignment` test, checks that the user is seeing the message we expect them to see. If someone accidentally changed that message, the test would fail. Or if they purposely removed it, thinking that it's not necessary, the test failure might cause them to reconsider.

Also, if we merged the two failures and someone removed the `else` statement, we wouldn't know it as the test would still pass (since it would be just checking for an exception being thrown).

suzannehamilton: Does the message get shown to the user regardless of what it says, or is there some logic that's depending on the wording of the message?

If it's the former, then that's what we should be testing - you wouldn't need multiple tests, just a test that whatever message is in the Error is show on screen.

npm/test/get-file-check-progress.test.ts
suzannehamilton: Does this still need to be updated to be a Promise?

techncl: Hmm, I'm not sure, I don't think so as the actual code doesn't return a promise with an error in it, it either returns a promise with the `GetFileCheckProgressQuery` result or an error. I would have suggested that maybe I could change the return type of the  function to an `Error`, but I don't think you can do that in Typescript.

npm/test/get-file-check-progress.test.ts
techncl: For the case of the graphql failure, it would show a different failure message than the empty data 'failure' but in the case of the empty data failure, the message would be the same each time except for the consignmentId, so I did a fuzzy match for just `No progress metadata found for consignment`.

npm/test/get-file-check-progress.test.ts
suzannehamilton: Ah, I think it actually returns a promise even though you don't declare one because it's an `async` function. So the code throws an error, but the actual return value is a rejected promise containing that error.

suzannehamilton (approved):
I'd still like to dig into the different error types to understand them, but I'm approving this to stop holding it up. This is definitely better than it was before, so thanks for doing the refactoring! 🎉

PR #751, opened 2021-07-09
Throw error if setting upload status as completed fails

techncl:


npm/src/updateconsignmentstatus/index.ts
```diff
@@ -18,23 +19,28 @@ export class UpdateConsignmentStatus {
...
+      if (!result.data || !result.data.markUploadAsCompleted || result.errors) {
+        const errorMessage: string = result.errors
+          ? result.errors.toString()
+          : "no data"
+        throw Error(errorMessage)
+      } else {
+        return result.data.markUploadAsCompleted
+      }
+    } catch (e) {
+      handleUploadError(
```
suzannehamilton, line 40: This function is about calling the API, so making it call a method which changes the UI mixes up the levels of abstraction, which I think will make it harder to find this code. Where is the `markConsignmentStatusAsCompleted` method called, and does it deal with code at the page level? Would moving the `catch` and the call to `handleUploadError` fit better there?

suzannehamilton, line 40: Can this try/catch be deleted now that upload/index.ts calls `handleUploadError`?

techncl: 🤦  ah, I thought I deleted it. Thanks

suzannehamilton (approved):
🎉

PR #753, opened 2021-07-12
Progress bar shows integers, tests updated

rsulxo:


suzannehamilton:
@techncl Would you like to review this one, since you've done a lot of work on the TypeScript code recently?

techncl (commented):
This is good! Just one thing, I'm being really really nitpicky 😄  but could you change

`Math.floor(
      50 + progressInformation.percentageProcessed / 2
    )`

`50 + Math.floor(
      progressInformation.percentageProcessed / 2
    )`

just so when someone reads it, they know immediately that the floor is being applied to the s3 upload percentage and the `50` is inconsequential when it comes to the `Math.floor`

rsulxo:
No problem 😄  I've changed that line and linted it too so hopefully should be OK now!

techncl (approved):

PR #780, opened 2021-07-23
Fix nan percentage on progress bar

techncl:


npm/test/clientfileprocessing.test.ts
```diff
@@ -12,6 +12,7 @@ import { S3Upload, ITdrFile } from "../src/s3upload"
 import { ManagedUpload } from "aws-sdk/clients/s3"
 import { mockKeycloakInstance } from "./utils"
 import { FileUploadInfo } from "../src/upload/upload-form"
+import S3 = require("aws-sdk/clients/s3")
```
MancunianSam, line 15: Do you need this import? It's only being used for `S3.ManagedUpload` and you've already got `ManagedUpload` imported further up. Can you remove this and just put `sendData: ManagedUpload.SendData[]` ?
Or is the `require` import the better way of doing it and you just need to remove the `import { ManagedUpload }` line?

MancunianSam (commented):
Looks good. Just one comment on the imports. I guess all the other changes are the prettier formatting and some variable name changes.

npm/test/clientfileprocessing.test.ts
techncl: Ah yes, what you suggested works; Thanks 👍 . For some reason IntelliJ imported S3 like this `import { S3, ManagedUpload } from "aws-sdk/clients/s3"` and then it was showing me errors so I imported it the way I did.

techncl:
> Looks good. Just one comment on the imports. I guess all the other changes are the prettier formatting and some variable name changes.

👍 , yeah, the 1st commit and maybe the 4th and 6th are the most important ones

MancunianSam (approved):
Looks good, thanks

PR #790, opened 2021-07-26
Change for-in loop to for-of loop to avoid potential bugs

techncl:
The previous code was fine as it was, but using a for-in could lead to bugs in the future, especially if
someone builds off that code with the wrong impression (e.g. batch order is retained) or Object/Array.prototype is altered
https://stackoverflow.com/questions/500504/why-is-using-for-in-for-array-iteration-a-bad-idea
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#difference_between_for...of_and_for...in


TomJKing (approved):
Looks good. Thanks for spotting this.

PR #794, opened 2021-07-27
Fix eslint error before upgrading eslint version

TomJKing:
Latest eslint version finding error, fix before upgrading


techncl (approved):
Looks good 🙂  .

PR #797, opened 2021-07-28
Refactor s3 upload tests

techncl:


npm/test/s3upload.test.ts
```diff
@@ -59,13 +59,50 @@ const checkCallbackCalls: (
   }
 }
 
+const createFiles = (
```
TomJKing, line 62: For unit tests it is better to keep the setup with the actual test. It makes it easier for someone else to see what is being tested. Also helps keep each unit test independent from any other tests and code as much as possible.

I've fallen into this trap of abstracting out test setup many times, only to have it cause problems later ;)

TomJKing (commented):
Thanks for looking at this. Just one comment.

npm/test/s3upload.test.ts
techncl: Ah, I see. This function creates a set of new files each time rather than sharing them with each test though. Should I delete this commit?

TomJKing: I'd retain the change to the array name, and just revert the changes related to the creation of files function. Thanks

TomJKing (approved):

PR #798, opened 2021-07-28
Add cache-control header with 'no-store' directive

techncl:
This in order to prevent using the cached file checks page that doesn't display the blue banner immediately


TomJKing (commented):
Looks good. 
I was wondering whether its possible to test this change. I had a quick look at the `play.api.test.Helpers.scala` where the `contentType` function at line 63 is used in the `FileChecksControllerSpec.scala`. There also looks like there is a `header` function which might let you test this in the `Helpers.scala`.

TomJKing (approved):

PR #800, opened 2021-07-29
Rewording error messages upload page

plembke-tna:
- wording: click 'return to start' instead of 'series'
- added button instead of having a link for consistency and good practice
- updated wording with single quotes where referred to a continue button (upload already complete)

Please test this again on your machine to make sure the button is not behaving weirdly (it displayed as disabled on my machine for a while).


techncl:
Hi Pia, This looks good, just had some comments.

1. As you know, the button is disabled. I noticed the reason why the button is disabled is because in the `upload-form.ts` file, on line 187, there is this disableButtonsAndDropzone() method [call](https://github.com/nationalarchives/tdr-transfer-frontend/blob/8bb0c49c0dec09c84288b8ef39b6c358a409a1c4/npm/src/upload/upload-form.ts#L187). I added this to prevent someone from (in the case that the page takes too long to move to the next page) clicking any buttons on the page. so I don’t know if you want to maybe add some code to that method to enable the button again. I was thinking of asking Suzanne if I should maybe remove this code as it was added when it took awhile for the page to transition to the next and the user could still click something on the page.

2. I think you should hide this new button you added (by default) until the error message shows, as it currently displays the button throughout on the `Uploading records` half.

techncl (dismissed):
Hi Pia, This looks good, just had some comments.

1. As you know, the button is disabled. I noticed the reason why the button is disabled is because in the `upload-form.ts` file, on line 187, there is this disableButtonsAndDropzone() method [call](https://github.com/nationalarchives/tdr-transfer-frontend/blob/8bb0c49c0dec09c84288b8ef39b6c358a409a1c4/npm/src/upload/upload-form.ts#L187). I added this to prevent someone from (in the case that the page takes too long to move to the next page) clicking any buttons on the page. so I don’t know if you want to maybe add some code to that method to enable the button again. I was thinking of asking Suzanne if I should maybe remove this code as it was added when it took awhile for the page to transition to the next and the user could still click something on the page.

2. I think you should hide this new button you added (by default) until the error message shows, as it currently displays the button throughout on the `Uploading records` half.

plembke-tna:
> Hi Pia, This looks good, just had some comments.
> 
>     1. As you know, the button is disabled. I noticed the reason why the button is disabled is because in the `upload-form.ts` file, on line 187, there is this disableButtonsAndDropzone() method [call](https://github.com/nationalarchives/tdr-transfer-frontend/blob/8bb0c49c0dec09c84288b8ef39b6c358a409a1c4/npm/src/upload/upload-form.ts#L187). I added this to prevent someone from (in the case that the page takes too long to move to the next page) clicking any buttons on the page. so I don’t know if you want to maybe add some code to that method to enable the button again. I was thinking of asking Suzanne if I should maybe remove this code as it was added when it took awhile for the page to transition to the next and the user could still click something on the page.
> 
>     2. I think you should hide this new button you added (by default) until the error message shows, as it currently displays the button throughout on the `Uploading records` half.

Thank you, Nigel. I think I need a little bit of help with this. Unfortunately, I am running out of time because I am on A/L from tomorrow. Would you be able to pick this up for me? I can't figure out why the button is already visible and does not appear together with the error message. There is some magic behind the scenes! Also I was wondering if the progress bar is supposed to stay on the page underneath the error summary? 
Regarding 1. - have you talked to Suzanne about this yet? I guess if you do remove it, the button won't be disabled which solves the problem. If not, however, would you be able to fix it? I would definitely need your help with this IF this can wait until I am back. 
Thanks!!

techncl:
> > Hi Pia, This looks good, just had some comments.
> > ```
> > 1. As you know, the button is disabled. I noticed the reason why the button is disabled is because in the `upload-form.ts` file, on line 187, there is this disableButtonsAndDropzone() method [call](https://github.com/nationalarchives/tdr-transfer-frontend/blob/8bb0c49c0dec09c84288b8ef39b6c358a409a1c4/npm/src/upload/upload-form.ts#L187). I added this to prevent someone from (in the case that the page takes too long to move to the next page) clicking any buttons on the page. so I don’t know if you want to maybe add some code to that method to enable the button again. I was thinking of asking Suzanne if I should maybe remove this code as it was added when it took awhile for the page to transition to the next and the user could still click something on the page.
> > 
> > 2. I think you should hide this new button you added (by default) until the error message shows, as it currently displays the button throughout on the `Uploading records` half.
> > ```
> 
> Thank you, Nigel. I think I need a little bit of help with this. Unfortunately, I am running out of time because I am on A/L from tomorrow. Would you be able to pick this up for me? I can't figure out why the button is already visible and does not appear together with the error message. There is some magic behind the scenes! Also I was wondering if the progress bar is supposed to stay on the page underneath the error summary?
> Regarding 1. - have you talked to Suzanne about this yet? I guess if you do remove it, the button won't be disabled which solves the problem. If not, however, would you be able to fix it? I would definitely need your help with this IF this can wait until I am back.
> Thanks!!

Hi Pia, I forgot to ask but I will. Sure, I can take over. Yes, I was also wondering about the progress bar remaining too; I think I might have put that on a card, but I'll check.

app/views/upload.scala.html
```diff
@@ -113,24 +113,29 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
...
                 </div>
             </div>
+            <a class="govuk-button" href="@routes.DashboardController.dashboard()" role="button" draggable="false" data-module="govuk-button">
+            Return to start
+            </a>
         </div>
-        <div>
-            <span id="upload-status-screen-reader"><label for="upload-records-progress-bar" class="govuk-label progress-label"></label></span>
-            <progress class="progress-display" value="0" max="100"></progress>
+        <div id="progress-bar-and-message">
```
suzannehamilton, line 133: Looking at this in the context of the existing elements, it's a bit counter-intuitive for the `<div>` on line 112 to be called `progress-bar` when this one is called `progress-bar-and-message`. They sound like they're nested in the wrong order.

Maybe the outer one should be called something more generic like `upload-progress`?

It would also be good to find a name for this one that doesn't list the elements, because it will be out-of-date if we change what's inside, but I'm struggling to think of a good name to use instead. 🙂

suzannehamilton (commented):
Just a comment on the HTML element IDs, but otherwise this looks good.

suzannehamilton (approved):

PR #801, opened 2021-07-29
Remove '/' from path if present

TomJKing:
Ensure file path entries in database are consistent

Files uploaded with 'drag and drop' have a '/' prepended to the file path, those uploaded with 'browse' don't

Make the change at the metadata upload point as this is just to ensure consistent entries in the database, and not for any other purpose


techncl (approved):
I've tested this locally and compared it with Integration and it looks good 👍

PR #802, opened 2021-07-29
Add check for cache-control header in the UploadContollerSpec

techncl:


TomJKing (approved):
Looks good. Thanks for adding this additional test.

PR #810, opened 2021-08-02
Refactor client metadata upload code

techncl:


npm/src/clientfilemetadataupload/index.ts
```diff
@@ -50,29 +45,37 @@ export class ClientFileMetadataUpload {
...
-      (cf) => cf.metadataInput
-    )
+    const {
+      metadataInputs,
+      matchFileMap
+    }: {
+      metadataInputs: ClientSideMetadataInput[]
+      matchFileMap: Map<number, File>
+    } = allFileMetadata.reduce(
+      (metadataInputsAndFileMapper, metadata: IFileMetadata, matchId) => {
```
suzannehamilton: Minor point, but the name "mapper" confused me the first time I read it, because it makes it sound like a function rather than a value.

You could just call it something like `result`, since it's only used in the context of this `reduce`.

npm/test/clientfilemetataupload.test.ts
```diff
@@ -195,6 +195,30 @@ test("createMetadataInputBatches generates batches of the defined size", () => {
   expect(result[1]).toEqual([input4, input5])
 })
 
+test("createMetadataInputsAndFileMap returns metadata inputs and file map", () => {
+  const metadata: IFileMetadata[] = [mockMetadata1, mockMetadata2]
+  const client = new GraphqlClient("https://test.im", mockKeycloakInstance)
```
suzannehamilton, line 206: I think this is copied from elsewhere, but we should use example.com when we need a realistic but fake URL. It's a reserved domain which is designed to be used in examples, whereas test.im is some random person's real website. 🙂

npm/test/clientfilemetataupload.test.ts
```diff
@@ -195,6 +195,30 @@ test("createMetadataInputBatches generates batches of the defined size", () => {
...
+  const uploadMetadata = new ClientFileMetadataUpload(client)
+
+  const { metadataInputs, matchFileMap } =
+    uploadMetadata.createMetadataInputsAndFileMap(metadata)
+
+  expect(metadataInputs).toHaveLength(metadata.length)
```
suzannehamilton, line 212: Minor thing, but can you make this `.toHaveLength(2)`? It makes it more obvious that we have a strong expectation about what the value should be, and it makes it easier to avoid bugs where we accidentally use empty arrays so none of the assertions below get run.

npm/test/clientfilemetataupload.test.ts
```diff
@@ -195,6 +195,30 @@ test("createMetadataInputBatches generates batches of the defined size", () => {
...
+  for (let i = 0; i < metadataInputs.length; i += 1) {
+    expect(metadataInputs[i].checksum).toEqual(metadata[i].checksum)
+    expect(metadataInputs[i].fileSize).toEqual(metadata[i].size)
+    expect(metadataInputs[i].originalPath).toEqual(metadata[i].path)
+    expect(metadataInputs[i].lastModified).toEqual(
+      metadata[i].lastModified.getTime()
+    )
+  }
+
+  expect(matchFileMap.size).toEqual(metadata.length)
```
suzannehamilton, line 222: Same for this one.

npm/test/clientfilemetataupload.test.ts
```diff
@@ -219,6 +219,30 @@ test("createMetadataInputsAndFileMap returns metadata inputs and file map", () =
...
 })
 
+test("createMetadataInputsAndFileMap replaces double slash in metadata inputs file path with single slash", async () => {
+  const mockMetadata1WithDoubleSlash = <IFileMetadata>{
+    ...mockMetadata1,
+    path: "/path/to/file1"
+  }
+  const mockMetadata2WithDoubleSlash = <IFileMetadata>{
+    ...mockMetadata2,
+    path: "path2/to/file2"
```
suzannehamilton, line 235: Neither of these examples look like they have a double slash in them, so it looks like test setup doesn't match the name of the test. Am I missing something?

suzannehamilton (commented):
Looks good, thanks. Just some minor comments.

npm/src/clientfilemetadataupload/index.ts
techncl: Ah, ok. I was was trying to come up with something descriptive because I thought it might be confusing for someone reading it

npm/test/clientfilemetataupload.test.ts
techncl: Ah, this was a mistake; I'll fix the naming. Thanks.

npm/src/clientfilemetadataupload/index.ts
suzannehamilton: Yeah, it's a tough one to name because it combines two outputs. And `reduce` functions are often hard to read if you're not fluent in functional programming, so I think calling it `result` makes it clearer what role it's playing in the `reduce` function.

You have to call `result.metadataInputs` and `result.matchFileMap` to interact with it, which makes it a bit easier to follow what's going on anyway.

npm/test/clientfilemetataupload.test.ts
suzannehamilton: Cheers for fixing the existing ones. 👍

npm/test/clientfilemetataupload.test.ts
```diff
@@ -195,6 +201,50 @@ test("createMetadataInputBatches generates batches of the defined size", () => {
...
+    )
+  }
+
+  expect(matchFileMap.size).toEqual(2)
+  for (let i = 0; i < metadataInputs.length; i += 1) {
+    expect(matchFileMap.get(i)).toEqual(metadata[i].file)
+  }
+})
+
+test("createMetadataInputsAndFileMap removes slash at beginning of metadata input's file path", async () => {
```
suzannehamilton, line 228: I think it's worth adding a short comment about the context for this, because I don't think I'd understand the need for it if I was reading this for the first time. Something like this? `Files added by drag-and-drop include a leading slash but files added using the Browse button do not. Check that any leading slashes are be removed for consistency.`

techncl: Ah ok. There is a comment in the [clientfilemetadata code](https://github.com/nationalarchives/tdr-transfer-frontend/blob/b0e904909616bf4d304e7d7a520c158b45917919/npm/src/clientfilemetadataupload/index.ts#L117) about it. Should I still add this comment for clarity though?

suzannehamilton: Yeah, I think so. Comments are helpful whenever something is surprising, and I think this test is surprising.

suzannehamilton (approved):

PR #840, opened 2021-08-13
Reduce redundant calls to the DOM to update the progress bar

techncl:


suzannehamilton:
Was this causing a problem with the screen reader, or a performance problem or something? Just trying to understand what improvement this gives.

techncl:
In some of the solutions I was trying out, it was causing an issue with the screen reader; I was meaning to add it but then your time-based solution came along. I was also thinking it would benefit users on less performant computers but it's just a theory; I can remove it until we've gotten some solid evidence that it could cause an issue?

suzannehamilton:
I think if it could cause an issue for screen readers, depending on what we do with the ARIA roles, that's a good reason to merge it.

suzannehamilton (approved):

techncl:
Having a little issue here 😅 . I'm basically trying to make all commits "Verified" because I think the fact that the first commit is not "Verified" is going to mean that I won't be able to merge. I'm using rebase but I'm running into merge conflicts

suzannehamilton (approved):

techncl:
👍

PR #879, opened 2021-08-31
Upgrade jest to latest version

TomJKing:
Dependabot Jest upgrade PR cause all tests to fail

Manually do upgrade with fix for Jest tests:
- 'ts-jest' version required upgrade: https://github.com/kulshekhar/ts-jest/issues/2612#issuecomment-848587164
- Add 'testEnvironment' to Jest config: https://stackoverflow.com/questions/67373953/jest-fails-with-self-is-not-defined-when-importing-react-diagrams


suzannehamilton (approved):
This looks good to me. I normally want to run either the JS or Scala tests, but not both, so running them together got in the way more than it helped.

PR #880, opened 2021-08-31
Upgrade to typescript 4.2.4

suzannehamilton:
Fix version number because current version (4.4) introduces breaking changes to the DOM types which Jest doesn't support yet.


suzannehamilton:
I've added a card about upgrading to 4.4, because it's a bit more complicated than this upgrade: [TDR-1356](https://national-archives.atlassian.net/browse/TDR-1356)

TomJKing (approved):

PR #885, opened 2021-09-02
Remove 'isComplete' from input of AddFilesAndMetadata mutation variables

techncl:


npm/src/clientfilemetadataupload/index.ts
```diff
@@ -56,11 +56,9 @@ export class ClientFileMetadataUpload {
     const allFiles: ITdrFile[] = []
 
     for (const [index, metadataInput] of metadataBatches.entries()) {
```
suzannehamilton, line 58: Can we also get rid of this index, or is it being used elsewhere?

suzannehamilton (approved):

PR #886, opened 2021-09-02
Get screen readers to read uploading records page

techncl:


techncl:
Just to note, this PR doesn't have to wait until we've come to a decision about how to update the progress bar for screen reader users as it's only about adding the ARIA roles (so that the page could be read out), so they need to be added to this page regardless.

suzannehamilton (approved):

PR #892, opened 2021-09-06
Tdr 1326 authentication error large upload

TomJKing:


npm/src/upload/index.ts
```diff
@@ -43,6 +50,17 @@ export class FileUploader {
...
     window.addEventListener("beforeunload", pageUnloadAction)
 
+    const intervalId = setInterval(() => {
+      if (idleSessionTimeoutAboutToExpire(this.keycloak)) {
+        //Allow for refreshing the token before the refresh token expires
+        //Keycloak documentation: https://www.keycloak.org/docs/latest/server_admin/#user-session-management
+        /*
+         SSO Session Idle: ... The idle timeout is reset by a client requesting authentication or by a refresh token request. ...
+        */
+        refreshOrReturnToken(this.keycloak)
```
suzannehamilton, line 60: Just to check - does this exchange the refresh token for a new set of tokens? As opposed to launching an iframe, I mean.

npm/src/upload/index.ts
```diff
@@ -43,6 +50,17 @@ export class FileUploader {
   ) => {
     window.addEventListener("beforeunload", pageUnloadAction)
 
+    const intervalId = setInterval(() => {
```
suzannehamilton, line 53: Since we know the expiry time of the token, is it possible to use `setTimeout` rather than `setInterval`? Or is there some reason you need to keep checking rather than relying on the original token expiry time?

TomJKing: From what I've seen when the access token is refreshed, the refresh token exp is reset. So it needs to be continually checked, I think using `setTimeout` means the reset only occurs once, which won't work where the idle session is more than an hour. I'm not sure what you mean by "... relying on the original token expiry time". The function `idleSessionTimeoutAboutToExpire` is relying on the current refresh token expiry to calculate whether the idle session timeout is going to expire.

suzannehamilton: So is the access token refreshed automatically - we can't add something that runs after every the token refresh?

npm/src/upload/index.ts
TomJKing: As far as I can tell debugging through, and looking at the response from the request, it provides a new set of tokens with different expiry times. The iframe doesn't appear to be touched.

npm/src/upload/index.ts
TomJKing: The access token is not refreshed automatically. Can do something like this though:

```
checkIdleSessionTimeOut: () => void = () => {
    //Refresh token will expire after 30mins, therefore check 1min before this, ie in 29 mins
    const timeoutInMs = 29 * 60 * 1000
    setTimeout(() => {
      console.log("Refreshing idle session")
      refreshOrReturnToken(this.keycloak).then(() => this.idleSession())
    }, timeoutInMs)
  }
...
uploadFiles: (
    files: IFileWithPath[],
    uploadFilesInfo: FileUploadInfo
  ) => Promise<void> = async (
    files: IFileWithPath[],
    uploadFilesInfo: FileUploadInfo
  ) => {
    window.addEventListener("beforeunload", pageUnloadAction)
    this.idleSession()
    ...
}
```
Can also calculate the timeout using "now" and the refresh token expiry value, instead of having a fixed value.
Assuming you're wanting avoid a continuous call to check the state of the refresh token?

TomJKing: The access token is not refreshed automatically when the session is idle, ie when an upload is occurring.

Is the problem that there is a continual check on the refresh token on page? Could have a function something like this, that would only attempt the refresh just before the refresh token is due to expire:

```
idleSession: () => void = () => {
    const nowInSecs = Math.round(new Date().getTime() / 1000)
    const refreshToken = this.keycloak.refreshTokenParsed
    if (refreshToken != undefined && refreshToken.exp != undefined) {
      const expInSecs = refreshToken.exp
      const timeoutInMs = (expInSecs - (nowInSecs + 60)) * 1000
      console.log("Timeout in Ms: " + timeoutInMs)
      setTimeout(() => {
        console.log("Refreshing idle session")
        refreshOrReturnToken(this.keycloak).then(() => this.idleSession())
      }, timeoutInMs)
    }
  }
```

npm/src/auth/index.ts
```diff
@@ -33,6 +33,26 @@ const isRefreshTokenExpired: (
   return token != undefined && token.exp != undefined && token.exp < now
 }
 
+export const refreshIdleSessionTimeout: (
```
suzannehamilton, line 36: Just thinking about the name for this function, because it sounds like it refreshes the token immediately. It would be good to make it more obvious that it starts a loop that repeatedly keeps the token refreshed.

What about something like `scheduleTokenRefresh` or `keepTokenUpdated`?

npm/src/auth/index.ts
```diff
@@ -33,6 +33,26 @@ const isRefreshTokenExpired: (
...
+  if (refreshToken != undefined && refreshToken.exp != undefined) {
+    const nowInSecs = Math.round(new Date().getTime() / 1000)
+    const expInSecs = refreshToken.exp
+    //Expiry is a future time, add min validity to the 'now' to check if expiry is about to expire
+    const timeoutInMs =
+      (expInSecs - (nowInSecs + idleSessionMinValiditySecs)) * 1000
+
+    setTimeout(() => {
+      refreshOrReturnToken(keycloak).then(() =>
+        refreshIdleSessionTimeout(keycloak)
```
suzannehamilton, line 50: When we come to implement the session refresh modal, we'll need to have a discussion about whether this should run indefinitely, or if there should be some point at which even this process gets stopped.

suzannehamilton (commented):
Still reviewing the tests for this, but I'll just post the comments I've added so far.

npm/src/upload/index.ts
suzannehamilton: Maybe we could have a quick call about this? It sounds like I don't understand the intended workflow. 🙂

npm/test/auth.test.ts
```diff
@@ -312,3 +312,126 @@ test("Throws an error if the access token and refresh token have expired", async
     new LoggedOutError("", "User is logged out")
   )
 })
+
+test("'scheduleTokenRefresh' should refresh tokens if refresh token will expire within the given timeframe", async () => {
+  const updateToken = jest.fn()
+  const isTokenExpired = jest.fn().mockImplementation(() => true)
+  const mockKeycloak: KeycloakInstance = {
```
suzannehamilton, line 319: As discussed on the call, it would be good to commonise these mocks if possible.

npm/test/utils.ts
```diff
@@ -1,21 +1,34 @@
...
-  hasResourceRole: jest.fn(),
-  loadUserInfo: jest.fn(),
-  loadUserProfile: jest.fn(),
-  token: "fake-auth-token"
+export const createMockKeycloakInstance: (
+    updateToken?: Mock,
+    isTokenExpired?: Mock,
+    refreshTokenParsed?: KeycloakTokenParsed
+) => KeycloakInstance = (updateToken = jest.fn(),
+                         isTokenExpired = jest.fn(),
```
suzannehamilton, line 9: All the places that use this parameter need to pass a function that returns a boolean. The type system doesn't help too much here because the type is just `Mock`.

Maybe the parameter could just be a simple boolean instead, and line 23 could be a function that returns the boolean value?

suzannehamilton: Similarly for `refreshTokenParsed`. It looks like the main thing that needs to be passed as a parameter is the expiry time.

TomJKing: Not sure changing the `refreshTokenParsed` simply to a date will work, as the test at line 230 requires no refresh token at all. So just passing a number won't work for that test.

npm/test/auth.test.ts
```diff
@@ -230,85 +232,83 @@ test("Throws an error if getting credentials from STS fails", async () => {
...
+
+test("'scheduleTokenRefresh' should not refresh tokens if the access token has not expired", async () => {
+  const mockUpdateToken = jest.fn()
+  const isTokenExpired = false
+  const refreshTokenParsed: KeycloakTokenParsed = { exp: Math.round(new Date().getTime() / 1000) + 60 }
+  const mockKeycloak: KeycloakInstance = createMockKeycloakInstance(mockUpdateToken, isTokenExpired, refreshTokenParsed)
+
+  jest.useFakeTimers()
+  scheduleTokenRefresh(mockKeycloak)
+  jest.runAllTimers()
```
suzannehamilton, line 285: I did not know Jest had something like this. Nice find!

suzannehamilton (approved):

PR #924, opened 2021-09-16
Add check for ta status flag

techncl:


test/validation/ValidatedActionsSpec.scala
```diff
@@ -103,7 +83,7 @@ class ValidatedActionsSpec extends FrontEndTestHelper {
     "return a redirect to the transfer agreement page if the transfer agreement is not complete" in {
       val consignmentId = UUID.randomUUID()
 
-      val graphqlConfigurationMock = mockUploadPermittedGraphqlResponse(itac.Data(Option(itac.GetTransferAgreement(false))), gcs.Data(Option(GetConsignment(CurrentStatus(Option.empty)))))
+      val graphqlConfigurationMock = mockUploadPermittedGraphqlResponse(itac.Data(Option(itac.GetTransferAgreement(false))), gcs.Data(Option(GetConsignment(CurrentStatus(None, Option.empty)))))
```
suzannehamilton, line 86: Is it significant that one of these parameters is `None` and one is `Option.empty`? If not, it would be less confusing for them to be the same.

test/services/ConsignmentStatusServiceSpec.scala
```diff
@@ -32,12 +32,12 @@ class ConsignmentStatusServiceSpec extends WordSpec with Matchers with MockitoSu
...
   "consignmentStatus function" should {
     "return the consignment status" in {
-      val data = Option(gcs.Data(Option(gcs.GetConsignment(CurrentStatus(Option("TestStatus"))))))
+      val data = Option(gcs.Data(Option(gcs.GetConsignment(CurrentStatus(Option("TestStatus"), None)))))
       val response = GraphQlResponse(data, Nil)
       when(getConsignmentStatusClient.getResult(token, gcs.document, Some(gcs.Variables(consignmentId))))
         .thenReturn(Future.successful(response))
       val status = consignmentStatusService.consignmentStatus(consignmentId, token).futureValue
-      status should equal(Option(CurrentStatus(Option("TestStatus"))))
+      status should equal(Option(CurrentStatus(Option("TestStatus"), None)))
```
suzannehamilton, line 40: Could this use a real value rather than `None`, to make sure that the real value has been passed through rather than a default?

app/controllers/TransferAgreementController.scala
```diff
@@ -49,6 +49,7 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
       val transferAgreementStatus: String = consignmentStatus.flatMap(_.transferAgreement).getOrElse("")
       if (transferAgreementStatus == "Completed") {
         Ok(views.html.transferAgreementAlreadyCompleted(consignmentId, transferAgreementForm, options))
+          .withHeaders("Cache-Control" -> "no-store, must-revalidate")
```
suzannehamilton, line 52: Could you add some more detail to the commit message for this change? Then someone who doesn't know why the cache-control header is needed will be able to work it out from the commit history.

app/controllers/TransferAgreementController.scala
```diff
@@ -43,9 +38,22 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
+  private val options: Seq[(String, String)] = Seq("Yes" -> "true", "No" -> "false")
+  implicit val language: Lang = langs.availables.head
+  private val addTransferAgreementClient = graphqlConfiguration.getClient[AddTransferAgreement.Data, AddTransferAgreement.Variables]()
+
+  def transferAgreement(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
 
-  def transferAgreement(consignmentId: UUID): Action[AnyContent] = consignmentExists(consignmentId) { implicit request: Request[AnyContent] =>
-    Ok(views.html.transferAgreement(consignmentId, transferAgreementForm, options))
+    for (consignmentStatus <- consignmentStatusService.consignmentStatus(consignmentId, request.token.bearerAccessToken)) yield {
```
suzannehamilton, line 48: I think a `map` would be easier to read than a `for` comprehension here, because there's only one statement in the comprehension, and the `yield` section has so much logic in it.

app/controllers/TransferAgreementController.scala
```diff
@@ -43,9 +38,22 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
+      val transferAgreementStatus: String = consignmentStatus.flatMap(_.transferAgreement).getOrElse("")
```
suzannehamilton, line 49: Rather than `getOrElse`, which implies that there's a good default value, you could leave this as an `Option[String]` and then do a case match rather than the `if` statement:

```
transferAgreementStatus match {
  case Some("Completed") => // Return the "already completed" version
  case _ => // Return the TA form
}
```

app/controllers/TransferAgreementController.scala
```diff
@@ -43,9 +38,22 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
+      if (transferAgreementStatus == "Completed") {
+        Ok(views.html.transferAgreementAlreadyCompleted(consignmentId, transferAgreementForm, options))
+          .withHeaders("Cache-Control" -> "no-store, must-revalidate")
+      } else {
+        Ok(views.html.transferAgreement(consignmentId, transferAgreementForm, options))
```
suzannehamilton, line 54: This version also needs the `Cache-Control` header. Otherwise the user's browser will cache this version, and it won't even bother to request the updated version when the user goes back to the page.

app/views/fileChecksResults.scala.html
```diff
@@ -8,7 +8,7 @@
         @progressIndicator(Messages("fileChecksResults.progress"))
         <h1 class="govuk-heading-l">Results of your checks</h1>
         <div id="upload-progress-success" class="success-summary" aria-labelledby="success-summary-title"
-        role="alert" data-module="success-summary" tabindex="-1">
+        role="status" data-module="success-summary" tabindex="-1">
```
suzannehamilton, line 11: What's the reason for this change? It's not obvious from the commit message.

app/views/inputSingleCheckbox.scala.html
```diff
@@ -14,13 +14,14 @@
...
-                    type="checkbox"
-                    value="@value"
-            />
+                @isSelected
+                class="govuk-checkboxes__input"
+                id="@elements.id"
+                name="@elements.id"
+                type="checkbox"
+                value="@value"
+                @elements.args.disabledInput(value.toString())
```
suzannehamilton, line 23: What are your thoughts on keeping these `FormFunctions` methods, after our discussion yesterday? Are you still planning to come back and improve them, or did you decide that they're OK as they are?

app/views/transferAgreementAlreadyCompleted.scala.html
```diff
@@ -0,0 +1,104 @@
...
+            <h2 class="success-summary__title">You have already confirmed all statements</h2>
+            <p class="govuk-body">Click 'Continue' to proceed with your transfer.</p>
+          </div>
+        </div>
+        @form(
+            routes.UploadController.uploadPage(consignmentId),
+            (Symbol("novalidate"), "")
+        ) {
+        <div class="govuk-form-group">
+            @inputSingleCheckbox(
```
suzannehamilton, line 34: Bit worried that we'll have to keep this page manually in sync with the existing transfer agreement form - both things like the headers, and the list of options. Is there a good way to commonise them so that they can pull in the shared text?

app/views/transferAgreementAlreadyCompleted.scala.html
```diff
@@ -0,0 +1,104 @@
...
+                Symbol("_value") -> "true",
+                Symbol("_smallCheckbox") -> true,
+                Symbol("_checkedOption") -> "true",
+                Symbol("_requiredOption") -> true,
+                Symbol("_disabledOption") -> "true"
+            )
+        </div>
+        <!-- Buttons -->
+        <div class="govuk-button-group">
+            <!-- Continue -->
```
suzannehamilton, line 92: I'd get rid of these comments - they code they explain is already clear enough. 🙂

test/controllers/TransferAgreementControllerSpec.scala
```diff
@@ -182,6 +169,35 @@ class TransferAgreementControllerSpec extends FrontEndTestHelper {
...
+      contentType(transferAgreementPage) mustBe Some("text/html")
+      headers(transferAgreementPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
+      transferAgreementPageAsString must include("Transfer agreement")
+      transferAgreementPageAsString must include("You have already confirmed all statements")
+      transferAgreementPageAsString must include("I confirm that the records are Public Records.")
+      transferAgreementPageAsString must include("I confirm that the records are all Crown Copyright.")
+      transferAgreementPageAsString must include("I confirm that the records are all in English.")
+      transferAgreementPageAsString must include("I confirm that the Departmental Records Officer (DRO) has signed off on the appraisal and selection")
+      transferAgreementPageAsString must include("I confirm that the Departmental Records Officer (DRO) has signed off on the sensitivity review.")
+      transferAgreementPageAsString must include("I confirm that all records are open and no Freedom of Information (FOI) exemptions apply to these records.")
```
suzannehamilton, line 199: Could you add some checks to make sure that it's not possible to submit the form again?

app/views/transferAgreementAlreadyCompleted.scala.html
```diff
@@ -0,0 +1,104 @@
...
+            <button data-prevent-double-click="true" class="govuk-button" type="submit" data-module="govuk-button" role="button">
+                Continue
```
suzannehamilton, line 94: Should this one be a link rather than a Submit button? What happens if you click it at the moment? Does it try to re-save the the form data?

suzannehamilton (commented):
Thanks for this! There are quite a lot of changes here, so there are quite a few comments. 🙂

test/validation/ValidatedActionsSpec.scala
techncl: 👍  I'm not entirely sure why the upload status is `Option.empty` but yes, I agree, it should be `None` for both

app/views/inputSingleCheckbox.scala.html
techncl: Yes, I was initially thinking of making the change now but then I'd have to change every place that's using it, so I thought, it's best to keep it as it is for now and then afterwards, create a new branch and start the refactoring there.

app/views/transferAgreementAlreadyCompleted.scala.html
techncl: This is a good concern, I could possibly pass in a variable to the view and add some logic that uses it to either hide or display the "already completed" message and the disabling and ticking of the boxes

app/views/transferAgreementAlreadyCompleted.scala.html
techncl: Yes, that would make more sense, although, if I keep just the one TA page, and use logic to display the banner and greyed out buttons, I'd have to keep it this way

app/views/fileChecksResults.scala.html
techncl: I thought I read somewhere that alert was for warnings but I can't find it anymore, so I'll revert it to the `alert` role

app/controllers/TransferAgreementController.scala
techncl: I've removed the new TA page (essentially merged it with the original) as you'll see in my newer commits; should I still update this commit?

suzannehamilton: This has made me think - there are a few pages which use this header, but I don't think the meaning is obvious if you're not familiar with cache controls.

What about having an extension method called something like `.uncached` which is used by every page that needs it? Then you could explain the purpose in a code comment on the method itself.

Give me a shout on Slack if that doesn't make sense, or if you want to talk about the idea.

test/controllers/TransferAgreementControllerSpec.scala
suzannehamilton: Is this the line that addresses this comment?

```
transferAgreementPageAsString must include(s"""<form action="/consignment/${consignmentId}/upload" method="GET" novalidate="">""")
```

I wonder if it would be better to test that there is no form submission button, since you're trying to demonstrate that there's no way to submit the form?

app/views/transferAgreement.scala.html
```diff
@@ -28,69 +39,73 @@ <h1 class="govuk-heading-l">Transfer agreement</h1>
...
+                Symbol("_checkedOption") -> shouldOptionBeSelected("openRecords"),
+                Symbol("_requiredOption") -> true,
+                Symbol("_disabledOption") -> isAlreadyCompleted.toString
             )
         </div>
-        <!-- Buttons -->
         <div class="govuk-button-group">
-            <!-- Continue -->
             <button data-prevent-double-click="true" class="govuk-button" type="submit" data-module="govuk-button" role="button">
                 Continue
```
suzannehamilton, line 106: Should this be a link rather than a form submission button when we're showing the "already completed" version?

techncl: The issue is, with these new set of commits, the TA page now one i.e. I'm no longer using a separate `TAAlreadyCompleted` page so do you suppose I should maybe add some logic to display the button (if the form has not been submitted) or the link (if it has)?

test/controllers/TransferAgreementControllerSpec.scala
techncl: I guess question is related to the question I answered below, should I add some logic do display either a submit button or a link button depending on if the user has submitted or not?

app/views/transferAgreement.scala.html
suzannehamilton: The other way to commonise the page is to have two separate templates, and then bring in the common components as partials. That's sometimes easier when there are a lot of differences between the two types of page, rather than lots of `if` statements. Would it be worth trying it here?

Give me a shout if that doesn't make sense.

app/views/transferAgreementHeader.scala.html
```diff
@@ -0,0 +1,10 @@
+@* helper package contains inputRadioGroup therefore exclude from imports to prevent ambiguous import with view of the same name *@
+@import helper.{inputRadioGroup => _, _}
+@import viewsapi.FormFunctions._
+@import java.util.UUID
```
suzannehamilton, line 4: Are all these imports necessary in this partial?

app/views/transferAgreementWarning.scala.html
```diff
@@ -0,0 +1,15 @@
+@* helper package contains inputRadioGroup therefore exclude from imports to prevent ambiguous import with view of the same name *@
+@import helper.{inputRadioGroup => _, _}
+@import viewsapi.FormFunctions._
+@import java.util.UUID
```
suzannehamilton, line 4: Same question about the imports.

app/controllers/TransferAgreementController.scala
```diff
@@ -43,14 +39,29 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
 
-  def transferAgreement(consignmentId: UUID): Action[AnyContent] = consignmentExists(consignmentId) { implicit request: Request[AnyContent] =>
-    Ok(views.html.transferAgreement(consignmentId, transferAgreementForm, options))
+    consignmentStatusService.consignmentStatus(consignmentId, request.token.bearerAccessToken).map {
+      consignmentStatus =>
+        val transferAgreementStatus: Option[String] = consignmentStatus.flatMap(_.transferAgreement)
+        transferAgreementStatus match {
+          case Some("Completed") => Ok(views.html.transferAgreementAlreadyConfirmed(consignmentId, transferAgreementForm, options, isAlreadyCompleted = true))
+                                      .uncache()
+          case _ =>  Ok(views.html.transferAgreement(consignmentId, transferAgreementForm, options, isAlreadyCompleted = false))
```
suzannehamilton, line 55: The `isAlreadyCompleted` parameter looks redundant now the views are separate.

app/views/transferAgreementHeader.scala.html
```diff
@@ -0,0 +1,10 @@
+@* helper package contains inputRadioGroup therefore exclude from imports to prevent ambiguous import with view of the same name *@
```
suzannehamilton, line 1: Does Play support you putting the partials into a subfolder? This views folder has got quite long, so it could do with some structure. We could start with this change.

app/controllers/TransferAgreementController.scala
techncl: Yes, I'm using passing it into the [formGroup](https://github.com/nationalarchives/tdr-transfer-frontend/pull/924/files#diff-51fb0672e4f7baf91c47a6201f1d28829ad17d7ac936969c7d9b4bcf7f7cc55b) view [here](https://github.com/nationalarchives/tdr-transfer-frontend/pull/924/files#diff-14c3b469abbfcd919a4f7bf521a6dafd9fdbe932289f3ca7bdd4ef23c665e7f9R26) and [here](https://github.com/nationalarchives/tdr-transfer-frontend/pull/924/files#diff-91030991d683ebd29198079140e082333656ea95b8c1a801be98331592819096R21)

The formGroup view uses it to determine whether to disable and select the checkboxes and whether to add the `CSRF.formField` to the form

suzannehamilton: Could the two main templates pass in `true` or `false` directly, rather than you having to pass a parameter from the controller?

techncl: Sure!

app/views/seriesDetails.scala.html
```diff
@@ -1,5 +1,6 @@
 @import helper._
 @import viewsapi.FormFunctions._
+@import views.html.partials.{errorMessage, errorSummary, inputDropdown, progressIndicator}
```
suzannehamilton, line 3: Are these all used on the series page? I'm surprised to see the progressIndicator here.

app/views/fileChecksResults.scala.html
suzannehamilton: This looks like it's gone back to "status". Which might be fine! I just couldn't figure out the reason for the change from the commit message.

app/views/seriesDetails.scala.html
suzannehamilton: Oh sorry, progressIndicator is the "Step X of Y" header, rather than the upload progress bar.

suzannehamilton (approved):

PR #941, opened 2021-09-21
Update auth utils library.

MancunianSam:
The library was changed a while ago on this commit https://github.com/nationalarchives/tdr-auth-utils/commit/25ce91da61663b68d751c5ea261564ff02a31204
This stopped us calling the keycloak server for every token verification
request as this was slowing down the API.
The front end hadn't been updated since then. This changes the classes
`token()` method to add the implicit.


TomJKing (approved):
Looks good. Cheers for sorting this out.

PR #949, opened 2021-09-23
Use cloudfront proxy for upload

MancunianSam:
- Add upload urls to the app config file.
- Add jest-fetch-mock library.
- Update the FrontEndInfo class on the scala side.
- Update javascript to use the new upload endpoint.
- Update the tests to work with the new changes.


conf/application.staging.conf
```diff
@@ -55,3 +49,5 @@ play.filters.hosts {
 }
 
 export.baseUrl = ${?EXPORT_API_URL}
+
+upload.url="https://upload.tdr-staging.nationalarchives.gov.uk"
```
TomJKing, line 53: Missing line break

npm/src/auth/index.ts
```diff
@@ -46,7 +44,14 @@ export const scheduleTokenRefresh: (
       (expInSecs - (nowInSecs + idleSessionMinValiditySecs)) * 1000
 
     setTimeout(() => {
-      refreshOrReturnToken(keycloak).then(() => scheduleTokenRefresh(keycloak))
+      refreshOrReturnToken(keycloak).then(() => {
```
TomJKing, line 47: Assuming these extra steps are now necessary to ensure that the upload to s3 continues by updating the signed cookie?

Will this have any impact should this function be used for anything other than the upload. For example keeping the user signed in on the transfer agreement page?

npm/src/auth/index.ts
```diff
@@ -67,59 +72,3 @@ export const refreshOrReturnToken: (
     throw "Token is expired"
   }
 }
-
-export const authenticateAndGetIdentityId: (
```
TomJKing, line 71: Satisfying ;)

npm/src/aws-config.ts
```diff
@@ -1,32 +1,32 @@
...
 const overrideS3Endpoint: (s3EndpointOverride: string) => void = (
   s3EndpointOverride: string
 ) => {
-  console.log(`Overriding S3 endpoint with '${s3EndpointOverride}'`)
+  const endpoint = new AWS.Endpoint(s3EndpointOverride)
+  AWS.config.update({
+    credentials: new Credentials({
+      // The AWS SDK requires credentials to generate a request, but we are not using them for authentication so set
+      // them to placeholder values
+      accessKeyId: "some-placeholder-id",
```
TomJKing, line 20: Maybe change to consistent naming by removing "some" from the value

npm/src/s3upload/index.ts
```diff
@@ -16,69 +16,76 @@ type TdrS3 = Pick<S3, "upload">
...
   s3: TdrS3
-  identityId: string
 
-  constructor(identityId: string, region: string) {
+  constructor() {
     const timeout = 20 * 60 * 1000
     const connectTimeout = 20 * 60 * 1000
-    this.s3 = new S3({ region, httpOptions: { timeout, connectTimeout } })
-    this.identityId = identityId
+    this.s3 = new S3({ httpOptions: { timeout, connectTimeout } })
```
TomJKing, line 23: How come the `region` is no longer needed?

TomJKing (commented):
Looks good. Just some minor comments and question. Also are we going to need to update the https://github.com/nationalarchives/tdr-local-aws repo, so it still works with the new cloudfront proxy?

MancunianSam:
I've had a look at the https://github.com/nationalarchives/tdr-local-aws repo and it's using s3 ninja which expects a bucket in the url like we're sending at the moment. We can make a new service to respond to the requests like Cloudfront does but it will take a bit of time. I guess it's up to you @TomJKing and @techncl if we want to do this now or put something in the backlog to do it later. Let me know what you think.

npm/src/s3upload/index.ts
MancunianSam: I think it's because we were sending the requests to https://tdr-upload-files-dirty-intg.s3.eu-west-2.amazonaws.com so it needed the region to construct the url but now we're overriding it, it doesn't bother so it doesn't need it.

npm/src/auth/index.ts
MancunianSam: I know. I'm going to delete the cognito terraform next week, I'm really looking forward to it.

npm/src/aws-config.ts
MancunianSam: Makes sense, I've removed it.

npm/src/auth/index.ts
MancunianSam: It will if we use this `setTimeout` on the other pages but I think we're going to have it so there's a popup saying `Do you want to stay signed in?` and then it runs some javascript when you click ok. We'd have to do a separate function then that calls `refreshOrReturnToken` but doesn't call the cookies API so this function would still just be for the upload page.

TomJKing: OK. Worth bearing mind. Thanks

npm/src/s3upload/index.ts
TomJKing: Fair enough.

TomJKing (approved):

TomJKing:
@MancunianSam @techncl I've actually not had the need to use the tdr-local-aws so happy just to add something to the backlog, and pick it up at convenient point.

TomJKing (approved):

PR #962, opened 2021-09-28
Move the get cookies url into the upload handler.

MancunianSam:
This will prevent a user being able to select and upload files before
the cookies endpoint has returned the signed cookies.

What does happen now is that there is a noticeable delay after you click
upload when the lambda has to start from cold but we should be able to
get around this if it causes us problems in the future.


npm/src/upload/index.ts
```diff
@@ -48,8 +48,12 @@ export class FileUploader {
     uploadFilesInfo: FileUploadInfo
   ) => {
     window.addEventListener("beforeunload", pageUnloadAction)
-    scheduleTokenRefresh(this.keycloak, `${this.uploadUrl}/cookies`)
-
+    const cookiesUrl = `${this.uploadUrl}/cookies`
+    scheduleTokenRefresh(this.keycloak, cookiesUrl)
+    await fetch(cookiesUrl, {
```
suzannehamilton, line 53: Does this first `fetch` still take a while if the Lambda is cold? And does that cause any problems if the user starts the upload too quickly?

I'm wondering whether any of the new users who might be uploading a few small files every day might get used to the workflow and end up clicking things really quickly.

npm/src/index.ts
```diff
@@ -50,26 +50,18 @@ export const renderModules = () => {
   if (uploadContainer) {
     uploadContainer.removeAttribute("hidden")
     const frontEndInfo = getFrontEndInfo()
-
     getKeycloakInstance().then((keycloak) => {
-      fetch(`${frontEndInfo.uploadUrl}/cookies`, {
```
suzannehamilton, line 55: Does this fetch still need to be run in a loop somehow, so that the upload can continue once the initial cookies have expired?

npm/src/upload/index.ts
MancunianSam: This happens after the start upload button is pressed. So they press submit and the call to the `/cookies` endpoint starts and then they see the upload progress bar. The fetch does take a while with a cold start so the user will see the progress bar sat on 0% for ~10s (worse case) and then it will start to upload. 
Once the fetch has returned, then the client side processing starts and the upload progress bar progresses.

npm/src/index.ts
MancunianSam: It is run in a loop, it's in here https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/npm/src/auth/index.ts#L48

suzannehamilton: Ah, I got confused because `scheduleTokenRefresh` sounded like it just refreshed the Keycloak token. Maybe it needs a name that makes it clear it refreshes all the credentials?

suzannehamilton (approved):

suzannehamilton:
I'm going to merge this to see if it fixes the end-to-end tests.

PR #984, opened 2021-10-04
Move validated actions code into upload controller

techncl:


app/controllers/UploadController.scala
```diff
@@ -17,9 +19,30 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
...
 
-  def uploadPage(consignmentId: UUID): Action[AnyContent] = uploadPermitted(consignmentId) { implicit request: Request[AnyContent] =>
-    Ok(views.html.upload(consignmentId, frontEndInfoConfiguration.frontEndInfo)).uncache()
+      transferAgreementStatus match {
+        case Some("Completed") =>
+          uploadStatus match {
+            case Some("InProgress") =>
+              Ok(views.html.uploadInProgress(consignmentId))
+            case Some("Completed") =>
+              Ok(views.html.uploadHasCompleted(consignmentId))
```
suzannehamilton, line 39: I missed this before, but all of the responses should have the cache-control header. Otherwise the user might return to the page and see the cached version (e.g. the InProgress one) rather than the latest one.

suzannehamilton (commented):
Looks good - just one comment on the cache headers.

app/controllers/UploadController.scala
techncl: 👍

suzannehamilton (approved):

PR #1040, opened 2021-10-19
Make the status 200 when navigating back to Transfer Agreement Page

techncl:
There was an issue where if you tried to submit an incomplete form, submitted a complete form (after the
page had reloaded with the error messages), clicked 'Continue' to go to the 'Upload' page and clicked the back
button of your browser, it would show the user the TA page with the error messages, instead of the 'already confirmed'
page. Making the status code, for POST requests that occur when the Transfer Agreement status has been set to 'Completed', 200,
remedies this and loads the 'already confirmed' page.


test/controllers/TransferAgreementControllerSpec.scala
```diff
@@ -164,6 +164,13 @@ class TransferAgreementControllerSpec extends FrontEndTestHelper {
...
         getValidKeycloakConfiguration, langs)
 
+      val client = new GraphQLConfiguration(app.configuration).getClient[gcs.Data, gcs.Variables]()
+      val consignmentResponse = gcs.Data(Option(GetConsignment(CurrentStatus(None, None))))
+      val data: client.GraphqlData = client.GraphqlData(Some(consignmentResponse))
+      val dataString: String = data.asJson.printWith(Printer(dropNullValues = false, ""))
+      wiremockServer.stubFor(post(urlEqualTo("/graphql"))
+        .willReturn(okJson(dataString)))
+
       val transferAgreementSubmit = controller.transferAgreementSubmit(consignmentId)
```
MancunianSam, line 174: It is worth having a test in here to make sure that we get the transfer agreement completed page when a request is submitted with errors?

techncl: Ah, yes; that's a good idea

PR #1065, opened 2021-10-26
Fix and refactor ta controller tests

techncl:


test/controllers/TransferAgreementControllerSpec.scala
```diff
@@ -279,20 +279,15 @@ class TransferAgreementControllerSpec extends FrontEndTestHelper {
...
-      ("droSensitivity", haveDroSensitivity.toString),
-      ("openRecords", recordsAreOpen.toString)
-    )
+      ("publicRecord", true.toString),
+      ("crownCopyright", true.toString),
+      ("english", true.toString),
+      ("droAppraisalSelection", true.toString),
+      ("droSensitivity", true.toString),
+      ("openRecords", true.toString)
+    ).dropRight(numberOfValuesToRemove)
```
MancunianSam, line 290: This is mostly personal preference but `true.toString` could be moved to `val value = "true"` and then use `value` in place. 
The other way is to make them boolean and map them at the end 
```
Seq(
      ("publicRecord", true),
      ("crownCopyright", true),
      ("english", true),
      ("droAppraisalSelection", true),
      ("droSensitivity", true),
      ("openRecords", true)
    ).dropRight(numberOfValuesToRemove).map(r => (r._1, r._2.toString))
```
As I say, my personal preference so feel free to ignore me.

techncl: Ah, yes. The first one is probably best/simpler; I'll do that.

MancunianSam (approved):
That's cleaner now, thanks.

PR #1085, opened 2021-11-03
Fix checked option bug and make tests more robust

techncl:


test/controllers/TransferAgreementControllerSpec.scala
```diff
@@ -159,13 +170,16 @@ class TransferAgreementControllerSpec extends FrontEndTestHelper {
       val dataString: String = data.asJson.printWith(Printer(dropNullValues = false, ""))
       wiremockServer.stubFor(post(urlEqualTo("/graphql"))
         .willReturn(okJson(dataString)))
+      val incompleteTransferAgreementForm: Seq[(String, String)] = getTransferAgreementForm(6)
```
MancunianSam, line 174: Would it be better here to just have `Seq()` instead of `getTransferAgreementForm(6)` ? This working is dependant on there being 6 fields on the form which could change (although you could just change the tests) and it's not immediately obvious that this is an empty form. What do you think?

test/controllers/TransferAgreementControllerSpec.scala
```diff
@@ -66,7 +77,7 @@ class TransferAgreementControllerSpec extends FrontEndTestHelper {
       contentType(transferAgreementPage) mustBe Some("text/html")
       headers(transferAgreementPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
       transferAgreementPageAsString must include(s"""<form action="/consignment/$consignmentId/transfer-agreement" method="POST" novalidate="">""")
-      checkHtmlContentForDefaultText(transferAgreementPageAsString)
+      checkHtmlOfFormOptions.checkForOptionAndItsAttributes(transferAgreementPageAsString, Map())
```
MancunianSam, line 80: The default is the empty Map so I don't think you need this argument here.

MancunianSam (commented):
There's a couple of minor suggestions on the tests. 
The other thing is that the logic for this is (almost) duplicated in the confirm transfer and transfer agreement pages so I wonder if it might be better moved out into the FormFunctions class. 

There's a couple of advantages I can see. We can re-use it for each different form, including the upcoming judgment ones. We can also add tests for it. 

I was playing around getting this working to see if it was worth it and I've put the changes on a new branch of yours. The comparison is here https://github.com/nationalarchives/tdr-transfer-frontend/compare/fixCheckedOptionBugAndMakeTestsMoreRobust...move-method-to-form-functions if you want to have a look.

If you think this makes things more difficult to read or maintain though then let me know.

test/controllers/TransferAgreementControllerSpec.scala
techncl: 😄  Ironically, I did this because I thought it would be clearer what was happening since it's consistent with the other tests. But yeah, I see what you mean, it is weird to pass in 6 only to have the code drop those values and return an empty `Seq`. Also, by making it a `Seq()`, you can see that it's empty immediately and the variable name gives additional info as to why it's an empty `Seq`

MancunianSam (approved):
Looks good, cheers.

PR #1110, opened 2021-11-11
Judgment upload page implementation

techncl:


test/controllers/UploadControllerSpec.scala
```diff
@@ -109,6 +109,87 @@ class UploadControllerSpec extends FrontEndTestHelper {
...
+
+      val uploadPage = controller.uploadPage(judgmentId)
+        .apply(FakeRequest(GET, s"/judgment/$judgmentId/upload").withCSRFToken)
+
+      status(uploadPage) mustBe OK
+      headers(uploadPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
+      contentAsString(uploadPage) must include("Upload a court judgment")
+      contentAsString(uploadPage) must include("You can upload your judgment by dragging and dropping it in the area below or by clicking 'Choose file'")
+    }
+
```
TomJKing, line 160: Unnecessary line break

test/controllers/UploadControllerSpec.scala
```diff
@@ -109,6 +109,87 @@ class UploadControllerSpec extends FrontEndTestHelper {
...
+      stubGetConsignmentStatusResponse()
+
+      val uploadPage = controller.uploadPage(judgmentId)
+        .apply(FakeRequest(GET, "/judgment/1/upload").withCSRFToken)
+      status(uploadPage) mustBe SEE_OTHER
+      redirectLocation(uploadPage).get must equal(s"/consignment/$judgmentId/transfer-agreement") // have to keep it as consignment until we create page
+    }
+
+    // This is unlikely but it's possible that they've bypassed the checks and partially agreed to things
```
TomJKing, line 129: I'd delete this comment.

test/controllers/UploadControllerSpec.scala
```diff
@@ -109,6 +109,87 @@ class UploadControllerSpec extends FrontEndTestHelper {
       contentAsString(uploadPage) must include("Uploading records")
       contentAsString(uploadPage) must include("Your upload is complete and has been saved")
     }
+
+    // Judgment Controller tests
```
TomJKing, line 113: Is this comment necessary?

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,565 @@
+import "@testing-library/jest-dom"
```
TomJKing, line 1: There appears to be quite a bit of shared test setup stuff between this set of tests and the `upload-form-folder.test.ts` Wondering if its possible to create `upload-form-utils` file and put common stuff in there, and import it into each of the different tests sets? Would reduce the amount of duplicate code, and make the tests themselves more visible. What do you think? Not sure that would even work with Jest tests.

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,565 @@
...
+test("clicking the submit button, without selecting a file, doesn't reveal the progress bar", async () => {
+  const mockDom = new MockDom()
+
+  const submitEvent = mockDom.createSubmitEvent()
+  await mockDom.form.handleFormSubmission(submitEvent)
+
+  expect(mockDom.uploadingRecordsSection).toHaveAttribute("hidden")
+})
+
+test("clicking the submit button, without selecting a file, doesn't disable the buttons on the page", async () => {
```
TomJKing, line 360: Could the "_clicking the submit button, without selecting a file ..._" tests just be combined into a single test, and just check all the different elements in the single test?

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,565 @@
...
+    "hidden",
+    "true"
+  )
+  expect(mockDom.fileRetrievalSuccessMessage).toHaveAttribute("hidden", "true")
+})
+
+// docx test and wrong number of files test
+// add file name to page
+
+test("input button updates the page with correct number of files if only 1 file has been selected", () => {
```
TomJKing, line 390: Can a user select more than one file using the browse? If so, should there be a test for this scenario?

TomJKing (commented):
Looks good. I've added some comments and suggestions. Some of which may turn out to be complete nonsense to feel free to ignore.

app/views/judgmentUpload.scala.html
```diff
@@ -0,0 +1,139 @@
+@import java.util.UUID
```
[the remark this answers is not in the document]
techncl: This is what I was thinking but it's a bit awkward (especially the form, which makes a call to a template). The "<noscript>" part at the top could be reused since it's exactly the same on both pages but the other parts have context-specific text; it would require a lot of if-else statements and possibly using the "messages" file.

Alternatively, I could pass a case class of strings specific to each page from the upload controller and populate the page; this would probably allow me to make the "upload in progress" section into a template and possibly the second half of the form. What do you think?

npm/src/upload/index.ts
```diff
@@ -74,6 +83,10 @@ export class FileUploader {
   }
 
   initialiseFormListeners(): void {
+    const checkIfJudgmentUserExists: boolean | undefined =
```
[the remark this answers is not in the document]
TomJKing: I'd have it as the ternary. It could return `judgment_user = false`. Its is an implementation detail in Keycloak that we don't set it to `false`, which may change in the future.

app/views/judgmentUpload.scala.html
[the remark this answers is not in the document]
TomJKing: How many strings do you think you would need to pass in via the case class. If its a lot I don't think its worth doing.

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,565 @@
...
+  ).not.toHaveAttribute("hidden", "true")
+
+  expect(mockDom.warningMessages.incorrectItemSelectedMessage).toHaveAttribute(
+    "hidden",
+    "true"
+  )
+  expect(mockDom.fileRetrievalSuccessMessage).toHaveAttribute("hidden", "true")
+})
+
+// docx test and wrong number of files test
```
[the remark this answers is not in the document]
techncl: No. 🤦

npm/test/upload-form-files.test.ts
techncl: Ah, yes, that's a good idea. I'll do the same for the upload page tests too

npm/src/upload/index.ts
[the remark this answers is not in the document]
techncl: Cool. thanks.

app/views/judgmentUpload.scala.html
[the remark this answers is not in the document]
techncl: 1. To move the `<noscript>`, 0
2. To Move the logged-out-error, 0
3. to move the title, instruction and upload-error, 3
4. To move the bottom half of the form (from the dropzone to the start upload buttons, about 6
5. In order to move the "upload in progress" section, should be about 4

Out of the ones that require strings to be passed in (3, 4 and 5), I think 5 would have the biggest significance, since it's large. What do you think?

Also, since 1 and 2 could be moved easily, should I just move them?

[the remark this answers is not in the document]
TomJKing: Go with 1 and 2. Leave the rest for now. We can come back to it, if the maintenance starts to become an issue in the future.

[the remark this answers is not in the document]
techncl: Cool. Thank you!

npm/test/upload-form-files.test.ts
techncl: They shouldn't be able to as, I've now removed the ability to do that yesterday. I'll change it to 

`input button updates the page with the file that has been selected`

..to reflect that.

test/controllers/UploadControllerSpec.scala
```diff
@@ -109,6 +109,83 @@ class UploadControllerSpec extends FrontEndTestHelper {
       contentAsString(uploadPage) must include("Uploading records")
       contentAsString(uploadPage) must include("Your upload is complete and has been saved")
     }
+
+    "redirect to the transfer agreement page if the transfer agreement for that judgment has not been signed" in {
```
TomJKing, line 113: Should this test be: "... for that judgment has not been **agreed**" to make it consistent with the other tests.

npm/test/upload-form-utils/mockUploadFormDom.ts
```diff
@@ -0,0 +1,249 @@
+import { IReader, IWebkitEntry, UploadForm } from "../../src/upload/upload-form"
```
TomJKing, line 1: Should rename this file using hypens instead of camel case to keep consistent with other file names: `mock-upload-form-dom.ts`

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,217 @@
...
+  dummyFolder
+} from "./upload-form-utils/mock-files-and-folders"
+import { MockUploadFormDom } from "./upload-form-utils/mockUploadFormDom"
+import { htmlForFileUploadForm } from "./upload-form-utils/html-for-file-upload-form"
+
+beforeEach(() => {
+  document.body.innerHTML = htmlForFileUploadForm
+})
+
+test("clicking the submit button, without selecting a file, doesn't reveal the progress bar & disables the buttons on the page", async () => {
```
TomJKing, line 15: Maybe remove the `&` and write out `and` in full instead?

npm/test/upload-form-files.test.ts
```diff
@@ -0,0 +1,217 @@
...
+  await mockDom.form.handleFormSubmission(submitEvent)
+
+  expect(mockDom.submitButton).toHaveAttribute("disabled", "true")
+  expect(mockDom.hiddenInputButton).toHaveAttribute("disabled", "true")
+  /*There is currently no way in Javascript to check if an event has been removed from an element,
+   therefore it is not possible to see if the submission code removed the drop event from the dropzone
+   */
+})
+
+test("clicking the submit button, after selecting a file, hides 'upload file' section & reveals progress bar", async () => {
```
TomJKing, line 201: Maybe remove the `&` and write out `and` in full instead?

npm/src/upload/index.ts
```diff
@@ -74,6 +82,9 @@ export class FileUploader {
   }
 
   initialiseFormListeners(): void {
+    const isJudgmentUser: boolean =
```
TomJKing, line 85: Does this still work if no `judgment_user` on the token?

TomJKing (commented):
Looks good. Just some minor additional comments.

app/views/partials/loggedOutErrorMessage.scala.html
```diff
@@ -0,0 +1,14 @@
+@()
+<div id="logged-out-error" class="govuk-error-summary logged-out-error" hidden aria-labelledby="logged-out-error-title"
+role="alert" tabindex="-1" data-module="govuk-error-summary">
+  <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-title">
+    You have been signed out due to inactivity
+  </h2>
+  <div class="govuk-error-summary__body logged-out-body">
+    <ul class="govuk-list govuk-error-summary__list">
+      <li>
+        <p><a class="logged-out-error-link">Sign in</a></p>
```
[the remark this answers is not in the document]
techncl: No. Because the sign in url changes each time, Sam added code to the [error handling js file](https://github.com/nationalarchives/tdr-transfer-frontend/blob/5a35e485c33d667b7e651575a9496ea8b55968ea/npm/src/errorhandling/index.ts#L110) that sets it

npm/src/upload/index.ts
techncl: Yes, it should work because if `judgment_user` is not there, it would return `undefined` and since that is not `true`, it will just make the value `false`

app/views/judgmentUpload.scala.html
```diff
@@ -0,0 +1,120 @@
...
+@import viewsapi.FrontEndInfo
+@import helper._
+@import views.html.partials.{frontEndInputs, greenTickMark, itemSelectionFailureMessage, progressIndicator}
```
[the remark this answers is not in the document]
TomJKing: Still seeing the individual imports here?

npm/test/upload-form-files.test.ts
TomJKing: Still seeing `&` in the test descriptions?

npm/test/upload-form-files.test.ts
TomJKing: Still seeing `&` in the test descriptions?

npm/src/upload/upload-form.ts
```diff
@@ -146,35 +148,78 @@ export class UploadForm {
...
-    }
-    const droppedItem: DataTransferItem | null = items[0]
-    const webkitEntry = droppedItem.webkitGetAsEntry()
-    if (webkitEntry!.isFile) {
-      this.rejectUserItemSelection()
-    }
-    this.removeDragover()
-    const files = await getAllFiles(webkitEntry, [])
-    this.checkIfFolderHasFiles(files)
+    if (this.isJudgmentUser) {
```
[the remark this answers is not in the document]
TomJKing: If this is too much of a pain just reply to this comment to that effect.

TomJKing (approved):

PR #1111, opened 2021-11-12
Tdr 1454 make series id optional

TomJKing:


test/services/ConsignmentServiceSpec.scala
```diff
@@ -52,7 +52,7 @@ class ConsignmentServiceSpec extends WordSpec with Matchers with MockitoSugar wi
 
   "consignmentExists" should {
     "return true when given a valid consignment id" in {
-      val response = GraphQlResponse(Some(gc.Data(Some(gc.GetConsignment(consignmentId, seriesId)))), Nil)
+      val response = GraphQlResponse(Some(gc.Data(Some(gc.GetConsignment(consignmentId, Some(seriesId))))), Nil)
```
MancunianSam, line 55: `seriesId` is set at the top  of the class. Could you change it to an `Option` there  to save from having `Some` each time?

MancunianSam (commented):
Looks good. One small suggestion on the tests.

MancunianSam (approved):

PR #1115, opened 2021-11-12
Add judgment dashboard page

MancunianSam:
The design of the dashboard page is the same for the standard user but instead of the button being a link, it now submits an empty form to the dashboard controller. This creates the consignment and redirects to a placeholder judgment transfer agreement page.


app/views/judgmentDashboard.scala.html
```diff
@@ -1,14 +1,23 @@
...
-                <h1 class="govuk-heading-l">Welcome to the Transfer Digital Records service</h1>
+@main("Welcome") {
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">Welcome to the Transfer Digital Records service</h1>
 
-                <h2 class="govuk-heading-m govuk-!-margin-top-9">Upload your judgment to start a new transfer</h2>
-                <p class="govuk-body">Placeholder page</p>
-            </div>
+      <h2 class="govuk-heading-m govuk-!-margin-top-9">Upload your records to start a new transfer</h2>
```
TomJKing, line 10: Not sure what the designs say, but think this should actually say: _Upload your **judgment** to start a new transfer_

test/controllers/DashboardControllerSpec.scala
```diff
@@ -39,23 +67,76 @@ class DashboardControllerSpec extends FrontEndTestHelper {
...
+      val dashboardPage = controller.dashboard().apply(FakeRequest(GET, "/dashboard").withCSRFToken)
       status(dashboardPage) mustBe OK
       contentType(dashboardPage) mustBe Some("text/html")
-      contentAsString(dashboardPage) must include ("Judgment Transfer Welcome")
+
+      contentAsString(dashboardPage) must include ("Welcome")
       contentAsString(dashboardPage) must include ("Welcome to the Transfer Digital Records service")
-      contentAsString(dashboardPage) must include ("Upload your judgment to start a new transfer")
-      contentAsString(dashboardPage) must include ("Placeholder page")
+      contentAsString(dashboardPage) must include ("Upload your records to start a new transfer")
```
TomJKing, line 78: Think this wording should actually be different for the judgments, so might be able to test for this to ensure the correct version of the page is being displayed.

test/controllers/DashboardControllerSpec.scala
```diff
@@ -39,23 +67,76 @@ class DashboardControllerSpec extends FrontEndTestHelper {
...
+        getValidKeycloakConfiguration,
+        consignmentService)
       val dashboardPage = controller.dashboard().apply(FakeRequest(GET, "/dashboard"))
       redirectLocation(dashboardPage).get must startWith("/auth/realms/tdr/protocol/openid-connect/auth")
       status(dashboardPage) mustBe FOUND
     }
   }
+
+  "DashboardController POST" should {
+    "create a new consignment" in {
```
TomJKing, line 95: Maybe reword the text of the test to make it more specific: _create a new consignment for a judgment user_

test/controllers/DashboardControllerSpec.scala
```diff
@@ -39,23 +67,76 @@ class DashboardControllerSpec extends FrontEndTestHelper {
...
+
+      wiremockServer.stubFor(post(urlEqualTo("/graphql"))
+        .willReturn(okJson(dataString)))
+
+      val redirect = controller.judgmentDashboardSubmit()
+        .apply(FakeRequest(POST, "/dashboard").withCSRFToken)
+
+      redirectLocation(redirect).get must equal(s"/judgment/$consignmentId/transfer-agreement")
+      wiremockServer.getAllServeEvents.size should equal(1)
+
```
TomJKing, line 115: Unnecessary line break

TomJKing (commented):

app/views/judgmentDashboard.scala.html
MancunianSam: The designs do say records but looking at the designs for all the other tickets, they all say judgment so it looks like it should be judgment. I've changed it. I also had to change the e2e tests back so can you re-check this one https://github.com/nationalarchives/tdr-e2e-tests/pull/139 when you get a second.

test/controllers/DashboardControllerSpec.scala
MancunianSam: I've changed it here. I've removed the check for the form element and just check for `Upload your judgment...`

TomJKing (approved):

PR #1129, opened 2021-11-16
Redirect to upload page if user submits the transfer agreement form twice

thanhz:
Redirects the user to the upload page if they try to submit the same consignment in another tab


app/controllers/TransferAgreementController.scala
```diff
@@ -65,8 +65,17 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
     }
 
     val successFunction: TransferAgreementData => Future[Result] = { formData: TransferAgreementData =>
-      transferAgreementService.addTransferAgreement(consignmentId, request.token.bearerAccessToken, formData)
-        .map(_ => Redirect(routes.UploadController.uploadPage(consignmentId)))
+      val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
```
TomJKing, line 68: Could use a `for` comprehension here, instead of the double nested flat map. Makes the code a bit easier to understand. Something like this:
```
...
val successFunction: TransferAgreementData => Future[Result] = { formData: TransferAgreementData =>
      val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)

      for {
        consignmentStatus <- consignmentStatusService.consignmentStatus(consignmentId, request.token.bearerAccessToken)
        transferAgreementStatus = consignmentStatus.flatMap(_.transferAgreement)
        r <- transferAgreementStatus match {
          case Some("Completed") => Future(Redirect(routes.UploadController.uploadPage(consignmentId)))
          case _ => transferAgreementService.addTransferAgreement(consignmentId, request.token.bearerAccessToken, formData)
            .map(_ => Redirect(routes.UploadController.uploadPage(consignmentId)))
        }
      } yield r
    }
...
```

TomJKing (commented):
Looks good. One minor suggestion.

app/controllers/TransferAgreementController.scala
thanhz: That Indeed looks better!

TomJKing (approved):

PR #1139, opened 2021-11-18
Change the folder structure of the views.

MancunianSam:
With most of the judgment pages, we're going to use different template files to try to keep as much logic as possible out of the templates.

The template list is already getting too long so I've created two directories, judgment and standard for each set of template files.

I'm not sure whether to have `app/views/judgment/judgmentDashboard.scala.html` instead of `app/views/judgment/dashboard.scala.html`

It would save duplicating judgment in the path but then we'd have two files called dashboard, only differentiated by the folder its in which might be more confusing.

Open to suggestions.


techncl:
Hmm, I guess having "judgment" before the page name could help when it comes to searching for the page and maybe prevent you from getting confused and accidentally editing/commiting the wrong page when you're context-switching

techncl:
What do you think @TomJKing ?

TomJKing (approved):

PR #1145, opened 2021-11-19
Add judgment file check results

MancunianSam:
Add judgment file checks page.
    
Added the new view in the judgments directory, used the same controller method and added a forAll in the tests as we're mostly checking for the same things for judgment and standard users.


app/views/judgment/judgmentFileChecksProgress.scala.html
```diff
@@ -0,0 +1,47 @@
...
+    @defining(play.core.PlayVersion.current) { version =>
+    <div class="govuk-grid-row file-check-progress">
+        <div class="govuk-grid-column-two-thirds">
+
+            @progressIndicator(Messages("checkingRecords.progress"))
+            @frontEndInputs(frontEndInfo)
+            <h1 class="govuk-heading-l">Checking court judgment record</h1>
+            <input id="consignmentId" type="hidden" value="@consignmentId">
+            <p class="govuk-body">Please wait while your court judgment is being checked. This may take a few minutes.</p>
+            <p class="govuk-body">The following checks are now being performed:</p>
```
TomJKing, line 16: I think we can remove the list of checks from the page. I seem to remember Kirsten saying this was what we're going to do for now.

app/views/judgment/judgmentFileChecksProgress.scala.html
```diff
@@ -0,0 +1,47 @@
...
+                    <p class="govuk-body">Please click 'Continue' to see your results.</p>
+
+                </div>
+            </div>
+            <form action="@routes.FileChecksResultsController.fileCheckResultsPage(consignmentId)">
+                <button type="submit" role="button" draggable="false" id="file-checks-continue" class="govuk-button @if(!fileChecks.isComplete){govuk-button--disabled}" @if(!fileChecks.isComplete){disabled}>
+                Continue
+                </button>
+            </form>
+
```
TomJKing, line 43: Unnecessary line break

TomJKing (commented):

app/views/judgment/judgmentFileChecksProgress.scala.html
MancunianSam: The design on the ticket has them but I remember now saying we'd get rid of them. I've taken them out.

TomJKing (approved):

PR #1157, opened 2021-11-22
Upgrade aws library

MancunianSam:
The main changes are that we now need a custom http handler to work with our Cloudfront distribution. We use this to remove the bucket name from the path and to add xhrWithCredentials.
The tests have also changed. We can no longer directly influence the onHttpProgress function in the tests so some of the tests checking the upload progress have had to change slightly.


npm/src/s3upload/tdr-fetch-handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+  })
+}
+
+export interface FetchHttpHandlerOptions {
+  /**
+   * The number of milliseconds a request can take before being automatically
+   * terminated.
+   */
+  requestTimeout?: number
```
TomJKing, line 23: Maybe change the name of the param to this, just to make it a bit clearer is milliseconds: `requestTimeoutMs`

npm/src/s3upload/tdr-fetch-handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    const { port, method } = request
+    const url = `${request.protocol}//${request.hostname}${
+      port ? `:${port}` : ""
+    }${path}`
+    // Request constructor doesn't allow GET/HEAD request with body
+    // ref: https://github.com/whatwg/fetch/issues/551
+    const body =
+      method === "GET" || method === "HEAD" ? undefined : request.body
+
+    const requestOptions: RequestInit = {//eslint-disable-line
```
TomJKing, line 68: Is this comment `//eslint-disable-line` needed?

npm/src/upload/index.ts
```diff
@@ -37,9 +40,21 @@ export class FileUploader {
     goToNextPage: () => void,
     keycloak: KeycloakInstance
   ) {
+    const requestTimeout = 20 * 60 * 1000
```
TomJKing, line 43: Maybe change param name to `requestTimeoutMs`?

TomJKing (commented):
What can I say, nothing like making things straightforward! ;) Just a couple of minor comments. Feel free to ignore.

npm/src/s3upload/tdr-fetch-handler.ts
MancunianSam: The RequestInit is part of the base typescript dom library so it doesn't need an explicit import which eslint doesn't like. I had a google though and you can add it to the .eslintrc and get rid of the comment so I've done that.

npm/src/s3upload/tdr-fetch-handler.ts
MancunianSam: Yeah makes sense, I've changed it here and where it's called.

TomJKing (approved):

PR #1211, opened 2021-12-09
Add judgment ta page

techncl:


app/controllers/HomepageController.scala
```diff
@@ -19,7 +19,11 @@ class HomepageController @Inject()(val controllerComponents: SecurityComponents,
   def judgmentHomepageSubmit(): Action[AnyContent] = secureAction.async {
     implicit request: Request[AnyContent] => {
       consignmentService.createConsignment(None, request.token).map(consignment => {
-        Redirect(routes.TransferAgreementController.judgmentTransferAgreement(consignment.consignmentid.get))
+        if (request.token.isJudgmentUser) {
+          Redirect(routes.TransferAgreementController.judgmentTransferAgreement(consignment.consignmentid.get))
+        } else {
```
TomJKing, line 24: I don't think we should be redirecting a non-judgment user to the standard pages. I think this should cause an error, as a non-judgment user should not be able to get to to the judgment home to submit the form in the first place. It will also provide protection if a user starts messing around with URLs.

app/controllers/TransferAgreementController.scala
```diff
@@ -22,7 +21,7 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
                                             val keycloakConfiguration: KeycloakConfiguration,
                                             langs: Langs)
                                            (implicit val ec: ExecutionContext) extends TokenSecurity with I18nSupport {
-  val transferAgreementForm: Form[TransferAgreementData] = Form(
+  val emptyTransferAgreementForm: Form[TransferAgreementData] = Form(
```
TomJKing, line 24: Why is this referred to as "empty". Can it not just remain named as: `transferAgreementForm`, or maybe `standardTransferAgreementForm`?

app/controllers/TransferAgreementController.scala
```diff
@@ -38,57 +37,65 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
         .verifying("All records must be open", b => b)
     )(TransferAgreementData.apply)(TransferAgreementData.unapply)
   )
+
   private val options: Seq[(String, String)] = Seq("Yes" -> "true", "No" -> "false")
   implicit val language: Lang = langs.availables.head
 
-  private def loadPageBasedOnTaStatus(consignmentId: UUID, httpStatus: Status, taForm: Form[TransferAgreementData])
-                                     (implicit request: Request[AnyContent]): Future[Result] = {
+  private def loadStandardPageBasedOnTaStatus(consignmentId: UUID, httpStatus: Status, taForm: Form[TransferAgreementData]=emptyTransferAgreementForm)
```
TomJKing, line 44: Spacing around the `=`

app/controllers/TransferAgreementController.scala
```diff
@@ -98,3 +105,6 @@ case class TransferAgreementData(publicRecord: Boolean,
                                  droAppraisalSelection: Boolean,
                                  droSensitivity: Boolean,
                                  openRecords: Boolean)
+
+case class JudgmentTransferAgreementData(reportingRestrictions: Boolean,
```
TomJKing, line 109: Don't need this case class at the moment, so it can be deleted.

app/controllers/TransferAgreementController.scala
```diff
@@ -38,57 +37,65 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
-    }
+  def judgmentTransferAgreement(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+      val warningMessage = Messages("judgmentTransferAgreement.warning")
+      Future(Ok(views.html.judgment.judgmentTransferAgreement(consignmentId, warningMessage)).uncache())
+  }
 
-    val successFunction: TransferAgreementData => Future[Result] = { formData: TransferAgreementData =>
-      val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
+  def transferAgreementSubmit(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    val isJudgmentUser = request.token.isJudgmentUser
```
TomJKing, line 70: Only calling this once so don't need to define as a val. Just add directly to the if statement: `if (request.token.isJudgmentUser)`

app/controllers/UploadController.scala
```diff
@@ -23,31 +23,48 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
...
           }
         case _ =>
           Redirect(routes.TransferAgreementController.transferAgreement(consignmentId))
       }
     }
   }
+
+  def judgmentUploadPage(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+    val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
+    val isJudgmentUser = request.token.isJudgmentUser
```
TomJKing, line 52: Doesn't look like `isJudgmentUser` is called any where, so can be deleted

conf/messages
```diff
@@ -10,6 +12,9 @@ fileChecksResults.description.plural=Your folder ''{0}'' containing {1} items ha
 
 transferComplete.error=There has been an error, contact us.
 
+judgmentTransferAgreement.warning=We cannot publish the court judgment without this: Name(s) of judge(s), name of parties, court, judgment date, neutral citation, case reference(s).
```
TomJKing, line 15: Not sure what the designs say, but don't think the word `this` is correct: _... the court judgment without the: Name(s) ..._

app/controllers/TransferAgreementController.scala
```diff
@@ -38,57 +37,65 @@ class TransferAgreementController @Inject()(val controllerComponents: SecurityCo
...
-      loadPageBasedOnTaStatus(consignmentId, BadRequest, formWithErrors)
-    }
+  def judgmentTransferAgreement(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
+      val warningMessage = Messages("judgmentTransferAgreement.warning")
+      Future(Ok(views.html.judgment.judgmentTransferAgreement(consignmentId, warningMessage)).uncache())
+  }
 
-    val successFunction: TransferAgreementData => Future[Result] = { formData: TransferAgreementData =>
-      val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
+  def transferAgreementSubmit(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request: Request[AnyContent] =>
```
TomJKing, line 69: I would split out the judgment transfer agreement into a separate function:

```
def judgmentTransferAgreementSubmit(consignmentId: UUID): Action[AnyContent] = secureAction.async {
  ...
  Future(Redirect(routes.UploadController.uploadPage(consignmentId)))
}
```
It makes it cleaner, also we were seeing some weird behaviours when Thanh tried using the same function for transfer complete. Should also make things a bit easier in the future should we need capture information about reporting restrictions

TomJKing (commented):
Looks good. Some minor comments and suggestions

app/controllers/TransferAgreementController.scala
techncl: I renamed it this for clarity, just to differentiate it from a filled in/partially filled in form, but I can change it back.

app/controllers/UploadController.scala
techncl: Yes; this was leftover code

TomJKing (approved):
Looks good.

PR #1226, opened 2021-12-15
Move transfer confirmation requests to a service

techncl:


test/services/ConfirmTransferServiceSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+    val graphQlResponse = GraphQlResponse[afjtc.Data](None, List(NotAuthorisedError("some auth error", Nil, Nil)))
+    when(judgmentGraphQlClient.getResult(token, afjtc.document, Some(afjtc.Variables(addFinalJudgmentTransferConfirmationInput))))
+      .thenReturn(Future.successful(graphQlResponse))
+
+    val transferConfirmation = confirmTransferService.addFinalJudgmentTransferConfirmation(consignmentId, token).failed
+      .futureValue.asInstanceOf[AuthorisationException]
+
+    transferConfirmation shouldBe a[AuthorisationException]
+  }
+}
```
thanhz, line 144: Missing newline

techncl: Good spot.

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -1,34 +1,33 @@
 package controllers
 
-import java.util.UUID
 import com.github.tomakehurst.wiremock.WireMockServer
-import com.github.tomakehurst.wiremock.client.WireMock.{equalToJson, ok, okJson, post, serverError, urlEqualTo}
+import com.github.tomakehurst.wiremock.client.WireMock.{status => _, _}
```
thanhz, line 4: Perhaps there's a reason but could we not just import as
`import com.github.tomakehurst.wiremock.client.WireMock._`?
or leave it explicit like before so we know what we're using?

techncl: Yeah, `._` makes sense. The thing you're referring to is something that IntelliJ did when I selected "Optimize imports"; I should probably not rely on it too much especially since it's not very good at handling the imports of Implicits.

thanhz (approved):

PR #1275, opened 2022-01-11
Make header content appear on multiple lines when page is zoomed in/v…

techncl:
…iewed on a small screen

When this issue was first discovered, it was using an older Mac with a 1600 x 900 resolution;
it could be replicated by zooming the page in, to 500%.

Now, when the page is zoomed in at 175+%, it will move the title and 'Sign out' button
on a new line; GDS' page does the same thing at 175%.

This solution should solve the header issue for:

1. Screens zoomed in 175% and above
2. Screens with lower resolutions
3. Small screens like phones/tablets

### DIFF
--- a/npm/css-src/sass/_header.scss
+++ b/npm/css-src/sass/_header.scss
@@ -5,12 +5,11 @@
 $govuk-header-link: govuk-colour('white');
 
 .govuk-header__tna-logo {
-  float: left;
   width: 220px;
 }
 
 .govuk-header__tna-logo-image {
-  width: 220px;
+  width: 90%;
 }
 
 .govuk-header__tna-content {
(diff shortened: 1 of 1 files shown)


npm/css-src/sass/_header.scss
```diff
@@ -58,3 +57,19 @@ $govuk-header-link: govuk-colour('white');
     text-decoration: none;
   }
 }
+
+@media (max-width: 48.06249em) {
```
TomJKing, line 61: How did you derive the max width value?

TomJKing (commented):

Also did you test this on different browsers? I tried on Firefox, and the "The National Archives" white box was missing from the standard size window at 100%. Might just be something weird on my machine though?

npm/css-src/sass/_header.scss
techncl: Because on GDS, they have a CSS rule for their header which states that when the screen width is over `48.0625em`,  the header should not display the content as vertically-stacked blocks and instead the content should be divided horizontally into sections
image

The CSS I wrote is just the opposite, any screen with a width of `48.06249em` and under, should no longer display content divided into horizontal sections and should be in blocks; I feel this this way reads better.

TomJKing (approved):
Looks good. I tried it on Chrome, Firefox and Edge seems to work fine. Also checked in responsive mode on Chrome and looks OK there too.

PR #1311, opened 2022-01-24
Add name to header

thanhz:
Will update `tdr-auth-utils` dependency in `build.sbt ` after https://github.com/nationalarchives/tdr-auth-utils/pull/46 is merged

### DIFF
--- a/app/auth/UnprotectedPageController.scala
+++ b/app/auth/UnprotectedPageController.scala
@@ -1,5 +1,7 @@
 package auth
 
+import com.nimbusds.jwt.SignedJWT
+import com.nimbusds.oauth2.sdk.token.BearerAccessToken
 import org.pac4j.core.profile.{CommonProfile, ProfileManager}
 import org.pac4j.play.PlayWebContext
 import org.pac4j.play.scala.{Security, SecurityComponents}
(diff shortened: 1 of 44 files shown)


app/auth/UnprotectedPageController.scala
```diff
@@ -10,10 +12,21 @@ import javax.inject.Inject
 class UnprotectedPageController @Inject ()(val controllerComponents: SecurityComponents) extends Security[CommonProfile] {
 
   implicit class RequestUtils(request: Request[AnyContent]) {
+    val webContext = new PlayWebContext(request, playSessionStore)
+    val profileManager = new ProfileManager[CommonProfile](webContext)
```
MancunianSam, line 16: Because the webContext is coming from the request, I don't think you can have it at the class level like this. I've tried it against the /faq page while logged in and after a few refreshes, the profile starts to come back as Option.empty. I'm not 100% sure why but if you create the webContext each time inside the two methods then it works as expected. You can put those lines into a method though, something like 
```
private def getProfile = {
      val webContext = new PlayWebContext(request, playSessionStore)
      new ProfileManager[CommonProfile](webContext)
    }
```

app/views/internalServerError.scala.html
```diff
@@ -1,4 +1,4 @@
-@()(implicit messages: Messages, request: RequestHeader)
+@(name: String)(implicit messages: Messages, request: RequestHeader)
```
MancunianSam, line 1: It doesn't look like this variable is used in the file. It just needs passing into the `@main`  method I think.

app/views/forbiddenError.scala.html
```diff
@@ -1,4 +1,4 @@
-@()(implicit messages: Messages, request: RequestHeader)
+@(name: String)(implicit messages: Messages, request: RequestHeader)
```
MancunianSam, line 1: This parameter just needs passing into the `@main` method as well.

MancunianSam (commented):
Looks good. Just a couple of small things to change.

MancunianSam (approved):

PR #1353, opened 2022-02-08
Update judgment upload page to match new content design

techncl:


test/controllers/UploadControllerSpec.scala
```diff
@@ -139,7 +139,7 @@ class UploadControllerSpec extends FrontEndTestHelper {
       status(uploadPage) mustBe OK
       headers(uploadPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
       contentAsString(uploadPage) must include("Upload a court judgment")
-      contentAsString(uploadPage) must include("You can upload your judgment by dragging and dropping it in the area below or by clicking 'Choose file'")
+      contentAsString(uploadPage) must include("You may now upload the court judgment you wish to transfer. You can only upload one file.")
```
TomJKing, line 142: Is it worth also checking that the: `We only accept Microsoft Word files (.docx).` wording is also present?

TomJKing (commented):
Looks good. One very minor suggestion, feel free to ignore. Thanks

test/controllers/UploadControllerSpec.scala
techncl: Hmm, we could, but we already have two lines from that page already being checked for content; should we add another?

TomJKing: Fair enough.

PR #1382, opened 2022-02-16
Remove tick mark & add padding to left of message

techncl:


TomJKing (approved):
Looks good. Assume this won't affect the e2e tests

techncl:
> Looks good. Assume this won't affect the e2e tests

Thanks. No, it shouldn't as we're not checking the message

PR #1395, opened 2022-02-21
Tdr-1666 new user successful registration confirmation holding page

thanhz:


app/controllers/RegistrationController.scala
```diff
@@ -0,0 +1,25 @@
...
+                                       val keycloakConfiguration: KeycloakConfiguration,
+                                       val consignmentService: ConsignmentService)
+                                      (implicit val ec: ExecutionContext) extends TokenSecurity with I18nSupport {
+
+  def complete(): Action[AnyContent] = secureAction {
+    implicit request: Request[AnyContent] => {
+      Ok(views.html.registrationComplete(request.token.name))
+    }
+  }
+
```
TomJKing, line 24: Unnecessary line break

app/views/registrationComplete.scala.html
```diff
@@ -0,0 +1,19 @@
+@(name: String)(implicit request: RequestHeader, messages: Messages)
```
TomJKing, line 1: This all look fine. I'm just wondering about the links to the FAQs and Help pages. I think we should have different pages for the judgments. It might just be worth following up with Kirsten to double check this. If so I think we will need some tasks to update the controllers for these pages, if we don't have so already.

TomJKing (commented):

app/views/registrationComplete.scala.html
thanhz: The new designs for the FAQ and Help pages for judgment still need to be reviewed & implemented. I can edit the links to those once they are done.

TomJKing (approved):
Looks good. Thanks

PR #1402, opened 2022-02-22
Change upload error message design

techncl:


npm/src/upload/form/display-warning-message.ts
```diff
@@ -20,6 +21,7 @@ export const rejectUserItemSelection = (
 
   warningMessageToReveal?.removeAttribute("hidden")
   warningMessageToReveal?.focus()
+  selectionArea?.classList.add("govuk-form-group--error")
 
   throw new Error(exceptionMessage)
```
TomJKing, line 26: Not sure we should be throwing an error here, as this is causing it to display in the console of the browser. Probably something to ask the rest of the team.

TomJKing (commented):
Looks good. Just one minor comment.

npm/src/upload/form/display-warning-message.ts
techncl: These exceptions are created by us though, for example [this](https://github.com/nationalarchives/tdr-transfer-frontend/blob/18d7b4972b9e93094c3f0acd3cf4fa601cfb92ec/npm/src/upload/form/upload-form.ts#L158). I think these were added just to ensure that the code doesn't continue and in case the message doesn't display for the user, for whatever reason. I guess we could do some refactoring and replace the throwing of exceptions with return statements and then check what was returned and return from the main function; maybe in another PR?

TomJKing: Fair enough.

TomJKing (approved):
Looks good. Thanks

PR #1406, opened 2022-02-23
Pass in empty form when navigating back to TA pages that previously d…

techncl:
…isplayed errors

Issue [1410](https://national-archives.atlassian.net/browse/TDR-1410)  has cropped up again this time on Firefox. I think the reason for this
is because when you click “back” in a browser like Chrome, you first get a page asking if you want to confirm submission; I think refreshing or
clicking back while on this page, makes a new request for the page to be safe. On Firefox, you don’t get this “confirm submission” page and I think instead,
it just resubmits the form. In our code, when the errorFunction was called and passes in the incomplete form it received, to loadStandardPageBasedOnTaStatus,
it was then passing this incomplete form into the 'transferAgreementPrivateBetaAlreadyConfirmed' and 'transferAgreementComplianceAlreadyConfirmed' views.
This commit will make it so that it will just pass in an empty form.


MancunianSam (approved):

techncl:
Thanks 👍

PR #1407, opened 2022-02-23
Changes to bucket endpoint config

MancunianSam:
This updates the TDR request handler to be almost identical to the AWS one but with `credentials: "include"` set. It also updates the way we do the S3 config so that it sends the request to the cloudfront distribution.


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
TomJKing, line 1: Is this supposed to be here? Appear to have two `package-lock.json` files now?

TomJKing (commented):
Looks good. Just one comment.

package-lock.json
MancunianSam: No, I must have run npm commands in this directory by mistake, I've deleted it.

TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):

PR #1416, opened 2022-02-25
Change link that 'Continue' button takes you on 'uploadHasCompleted' …

techncl:
…Page


app/controllers/UploadController.scala
```diff
@@ -63,7 +63,7 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
         case Some("InProgress") =>
           Ok(views.html.uploadInProgress(consignmentId, pageHeading, request.token.name)).uncache()
         case Some("Completed") =>
-          Ok(views.html.uploadHasCompleted(consignmentId, pageHeading, request.token.name)).uncache()
+          Ok(views.html.uploadHasCompleted(consignmentId, pageHeading, request.token.name, request.token.isJudgmentUser)).uncache()
```
MancunianSam, line 66: I'm just wondering whether it's best to use `isJudgmentUser` here. This action is wrapped with the `judgmentTypeAction` which means that if the consignment type is not set to judgment, then they get forbidden. What could happen, it's unlikely I know is:
* A user logs in as a judgment user and creates a consignment.
* The user later logs in as a standard user and goes to this consignment.
* They will see the judgment upload page because of the consignment type but the link will be pointing to the standard user processing page because of their token claim.

It might be safer to just put `true` here and `false` on line 41 (as that page can only be accessed by standard consignment types. 

I'm not 100% sure on this though, what do you think? There's other places where we're doing `request.token.isJudgmentUser` inside actions wrapped in `judgmentTypeAction` so if you think it's worth changing here then we can change those as well.

test/controllers/UploadControllerSpec.scala
```diff
@@ -218,6 +218,54 @@ class UploadControllerSpec extends FrontEndTestHelper {
         status(uploadPage) mustBe FORBIDDEN
       }
     }
```
MancunianSam, line 220: Is it worth adding a line to the `render the judgment upload is complete page if the upload has completed` test and the `render the upload is complete page if the upload has completed` to make sure the correct link is shown?

app/controllers/UploadController.scala
techncl: Ah, so this is the case where the user is both a judgment user and standard user? Then `request.token.isJudgmentUser` would return `true` in both cases? Yes, it makes sense then.

test/controllers/UploadControllerSpec.scala
techncl: The "Upload has completed" page and the "Upload in progress page", you mean? Do you mean like checking for text that is specific to these pages? Because I was thinking this but then realised that if you access the judgment page when you are a standard user (and vice versa), it just shows the "You are not permitted to see this page This page is restricted to certain users only." message and nothing "progress"/"completed" page-specific.

MancunianSam: For those two tests I mentioned, they're checking for the `Your upload is complete and has been saved` text. You've made changes now so the button link is different depending on the user so I was wondering whether it was worth adding a check at the end of both of those tests for the link address.

techncl: But the link address is the same for all 3, they all end in `/upload`. Should we maybe change that and have `/upload-has-completed` and `/upload-in-progress`? On the negative side, the users might save that link and try to access the platform via it, on the positive side, we can use it for the tests.

MancunianSam: I meant the link in the button, the one that has an if condition here https://github.com/nationalarchives/tdr-transfer-frontend/blob/8a33ae06d37ae358dc5eae7d59d3cfb8328893fb/app/views/uploadHasCompleted.scala.html#L12
The difference is only /consignment/{id}/records vs /judgment/{id}/records so it's probably not worth testing for.

MancunianSam (approved):
Looks good mate, cheers

PR #1430, opened 2022-03-01
Send empty directories to API

MancunianSam:
Extracts empty folders from the drag and drop and sends them to the API.

Needs https://github.com/nationalarchives/tdr-consignment-api/pull/201 to be merged and then https://jenkins.tdr-management.nationalarchives.gov.uk/job/TDR%20Graphql%20Code%20Generation/ to be run to update the npm library.


npm/test/upload-form-folder.test.ts
```diff
@@ -138,6 +138,110 @@ test("dropzone updates the page with correct folder information if there is a ne
   expect(mockDom.folderSizeElement!.textContent).toStrictEqual("2 files")
 })
 
+test("dropzone updates the page with correct folder information if there is are valid files and an empty directory", async () => {
```
TomJKing, line 141: Reword test title slightly: ... information if there are valid files ...

TomJKing (commented):

Also we're now using a mixture of "folder" and "directory" in the TS code. Is it worth renaming to use "directory" throughout to make things consistent. Think it'll mean a lot of changes, so might be worth doing in a separate PR? What do you think?

MancunianSam:
We can rename variables and names to directory instead of folder. The only thing is I think the actual database entry still needs to be Folder as that's what the archivists are expecting but that shouldn't affect the front end too much. I'll put a ticket in Jira to update them all.

TomJKing (approved):
Looks good. Thanks

PR #1431, opened 2022-03-01
Add survey link to complete page

techncl:


TomJKing (commented):
I think this link is specific to judgments only. So needs some logic to show it for judgment only. Probably best to double check with Sarra. Thanks

TomJKing (approved):
Looks good. Thanks

PR #1438, opened 2022-03-03
Add Remove file/folder button for upload page

thanhz:


app/views/judgment/judgmentUpload.scala.html
```diff
@@ -39,12 +39,13 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
                 <div class="drag-and-drop">
-                    <div class="govuk-summary-list govuk-file-upload">
+                    <div id="success-message-row" class="govuk-summary-list govuk-file-upload">
                         <div class="govuk-summary-list__row">
                             <dd id="folder-selection-success" class="govuk-summary-list__value drag-and-drop__success" hidden
                                 tabindex="-1" role="alert" aria-describedby="success-message-text">
-                                <div>
-                                    <p id="success-message-text">The file "<span id="file-name"></span>" has been selected</p>
+                                <div class="success-message-flexbox-container">
+                                    <p class="success-message-flexbox-item-1" id="success-message-text">The file "<span id="file-name"></span>" has been selected </p>
```
MancunianSam, line 47: There's no css defined for the `success-message-flexbox-item-1` so do we need that class here? I've taken it out and it looks the same as far as I can see. 

If you need it for any reason that's OK but if you don't then you can remove the class from here and also from the places in the tests that it's added as well which should clear things up a bit.

npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -64,3 +64,12 @@
 .error-messages {
   margin-bottom: 0;
 }
+
+.success-message-flexbox-container {
+  align-items: center;
+  display: flex;
+}
+
+.success-message-flexbox-item-2 {
```
MancunianSam, line 73: From my other comment about removing the `success-message-flexbox-item-1` class. If you do that, can this be renamed to `success-message-flexbox-item` ?

npm/src/upload/form/upload-form.ts
```diff
@@ -131,6 +131,20 @@ export class UploadForm {
     displaySelectionSuccessMessage(this.successMessage, this.warningMessages)
   }
 
+  removeSelectedItem = (): void => {
```
MancunianSam: We talked about this on Slack but you need to reset the form here and add an event in to prevent the url changing. Something like 
```
removeSelectedItem = (ev: Event): void => {
  ev.preventDefault()
  const successMessageRow: HTMLElement | null = document.querySelector(
    "#success-message-row"
  )
  this.selectedFiles = []
  successMessageRow?.setAttribute("hidden", "true")
  this.formElement.reset()
}
```

MancunianSam (commented):
Looks good, just a couple of things around the classes and naming.

MancunianSam (approved):

PR #1453, opened 2022-03-08
Apply bold styling only to success message in accordance with designs

techncl:


npm/css-src/sass/_drag-and-drop.scss
```diff
@@ -73,3 +72,9 @@
 .success-message-flexbox-item {
   margin-left: auto;
 }
+
+/* stylelint-disable */
```
MancunianSam, line 76: Could you add a class into the <p> tags so you can reference that and then you won't need the stylelint disable?

MancunianSam (approved):
Cheers, looks good.

PR #1456, opened 2022-03-11
Add Zip and PasswordProtected error messages

thanhz:
generatedGraphql will be updated when [this](https://github.com/nationalarchives/tdr-generated-graphql/pull/422) pr is merged


app/views/fileChecksFailed.scala.html
```diff
@@ -20,16 +20,28 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+                  These commonly have file extensions such as .zip, .iso, .7z, .rar and others.
+                  please see our <a class="govuk-link" href="@routes.FaqController.faq()" target="_blank" rel="noreferrer noopener">FAQ(Opens in new tab)</a> for a full list.
+                  Either remove or unpack your zip and archival package files and try uploading again.</p>
+                } else {
+                  <p class="govuk-body">One or more files you uploaded have failed our checks. Please
+                    <a class="govuk-link" href="@routes.ContactController.contact()">contact</a>
+                    us to resolve this issue.
+                  </p>
+                  <p class="govuk-body">Possible failure causes:</p>
+                  <ul class="govuk-list govuk-list--bullet">
```
TomJKing, line 38: Not sure what the designs have now, but wondering whether this list is needed anymore, given we're capturing password protected and zip error separately.

app/controllers/FileChecksResultsController.scala
```diff
@@ -30,8 +30,9 @@ class FileChecksResultsController @Inject()(val controllerComponents: SecurityCo
         )
         Ok(views.html.standard.fileChecksResults(consignmentInfo, consignmentId, request.token.name))
       } else {
-        Ok(views.html.fileChecksFailed(request.token.name, isJudgmentUser = false))
-      }
+        val fileStatus = fileCheck.files.find(!_.fileStatus.contains("Success")).get.fileStatus.get
```
TomJKing, line 33: Wondering what should happen if there are both zips and password protected files found? I think that is a question for Kirsten and Sarra. Not sure there are designs for such a case? I think the `fileStatus` should be a list to pass into the template.

app/views/fileChecksFailed.scala.html
```diff
@@ -1,5 +1,5 @@
 @import views.html.partials.progressIndicator
-@(name: String, isJudgmentUser: Boolean)(implicit request: RequestHeader, messages: Messages)
+@(name: String, isJudgmentUser: Boolean, fileStatus: String = "")(implicit request: RequestHeader, messages: Messages)
```
TomJKing, line 2: Think the `fileStatus` value should be list. See comments above, given there is the possibility that could have zip and password protected errors.

TomJKing (commented):
Looks good. Just some comments, I think you might need to speak to Kirsten, or Sarra, as I'm wondering what should happen if we get both zip and password protected errors, and whether that should be a separate story?

thanhz:
I am still waiting on feedback from Kirsten on what she wants to do with the old general error message and what should happen for an upload containing both password protected and zip files. But I have made a new partial html view and it now takes in a list.

thanhz:
If folder contains zip and passwordprotected go to general error page

TomJKing (commented):
Looks good. Just a couple of minor suggestions

app/views/fileChecksFailed.scala.html
```diff
@@ -1,5 +1,6 @@
-@import views.html.partials.progressIndicator
-@(name: String, isJudgmentUser: Boolean)(implicit request: RequestHeader, messages: Messages)
+@import views.html.partials.{progressIndicator, fileCheckErrorMessage}
+
+@(name: String, isJudgmentUser: Boolean, fileStatus: List[String] = Nil)(implicit request: RequestHeader, messages: Messages)
```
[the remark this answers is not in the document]
TomJKing: This instance is still referring to `fileStatus` instead of `fileStatuses`

TomJKing (commented):
Looks good. Just one very minor comment. Thanks

TomJKing (approved):
Looks good. Thanks

PR #1465, opened 2022-03-14
Add consignment ref to email subject

thanhz:


thanhz:
There is also a mail link on the error message for the filechecks failure for judgments. Not sure if this should include a consignment ref as well? But this would mean another call to db. One for the consignment ref and another for the fileStatus.

TomJKing (commented):
Looks good. It might be worth demonstrating this to Kirsten / Sarra to double check they're happy with the what is added to the subject line.

Also regarding your other comment about adding this functionality to the file checks failure page, I would again check with Kirsten /Sarra whether they want this. If so, it should be raised as a separate ticket.

TomJKing (approved):
Looks good.

PR #1469, opened 2022-03-15
Move to github actions

MancunianSam:
Move to GitHub actions
This depends on https://github.com/nationalarchives/tdr-github-actions/pull/5


.github/mergify.yml
```diff
@@ -0,0 +1,34 @@
+queue_rules:
+  - name: default
```
TomJKing, line 2: Is the `name` value correct?

.github/mergify.yml
```diff
@@ -0,0 +1,34 @@
...
+      - or:
+          - files=build.sbt
+          - files~=^(!?project/)
+    actions:
+      review:
+        type: APPROVE
+        message: Automatically approving Scala Steward
+      queue:
+        name: default
+  - name: automatic merge for dependabot
```
TomJKing, line 21: Capitalise "Dependabot" for consistency?

.github/mergify.yml
```diff
@@ -0,0 +1,34 @@
...
+        - "check-success=security/snyk (nationalarchives)"
+        - or:
+          - files=npm/package.json
+          - files=npm/package-lock.json
+    actions:
+      review:
+        type: APPROVE
+        message: Automatically approving dependabot
+      queue:
+        name: default
```
TomJKing, line 34: Missing line break

.github/workflows/test.yml
```diff
@@ -0,0 +1,22 @@
...
+permissions:
+  id-token: write
+  contents: read
+jobs:
+  test:
+    uses: nationalarchives/tdr-github-actions/.github/workflows/tdr_test.yml@main
+    with:
+      repo-name: tdr-transfer-frontend
+      test-command: |
+        npm --prefix npm ci
```
TomJKing, line 17: Do we need the npm linting commands in here as well, or is that covered by the `npm run checks` command?

TomJKing (commented):
Looks good. Just some minor comments

.github/mergify.yml
MancunianSam: It can be anything really, it's just an identifier. The mergify setup instructions call it that so I copied it. I can change it I think if you like.

.github/mergify.yml
MancunianSam: I've changed this one and the other name. The author field has to stay lower case as that's how it shows in GitHub.

.github/workflows/test.yml
MancunianSam: It's covered in the `run checks` I've updated the package json to add the checks command which runs the linting and the tests.

techncl (approved):
Looks good to me.

PR #1479, opened 2022-03-16
Remove before you start

thanhz:


techncl (changes requested):
Looks good. The only thing is the capitalised "T" in "Transfer" on each bullet show be a lowercased "t" as the bullets are still part of the sentence.

techncl (approved):

PR #1524, opened 2022-03-24
Make list of statuses exhaustive & Throw Exception if unexpected stat…

techncl:
…us is passed in


app/controllers/ConfirmTransferController.scala
```diff
@@ -55,11 +55,13 @@ class ConfirmTransferController @Inject()(val controllerComponents: SecurityComp
...
         confirmTransferStatus match {
           case Some("Completed") => Future(Ok(views.html.standard.confirmTransferAlreadyConfirmed(consignmentId, request.token.name)).uncache())
-          case _ =>
+          case None =>
             getConsignmentSummary(request, consignmentId)
               .map { consignmentSummary =>
                 httpStatus(views.html.standard.confirmTransfer(consignmentId, consignmentSummary, finalTransferForm, request.token.name)).uncache()
               }
+          case _ =>
+            throw new IllegalStateException(s"Unexpected Confirm Transfer status: $confirmTransferStatus")
```
MancunianSam, line 64: As a general comment, it might be worth adding the consignment ID into the error message for each of these so we can see which consignment has failed. What do you think?

techncl: Sure, sounds like a good idea!

app/controllers/TransferAgreementPrivateBetaController.scala
```diff
@@ -50,9 +50,11 @@ class TransferAgreementPrivateBetaController @Inject()(val controllerComponents:
           case Some("InProgress") | Some("Completed") =>
             Ok(views.html.standard.transferAgreementPrivateBetaAlreadyConfirmed(
               consignmentId, transferAgreementForm, transferAgreementFormNameAndLabel, warningMessage, request.token.name)).uncache()
-          case _ => httpStatus(
+          case None => httpStatus(
             views.html.standard.transferAgreementPrivateBeta(
               consignmentId, taForm, transferAgreementFormNameAndLabel, warningMessage, request.token.name)).uncache()
+          case _ =>
+            throw new IllegalStateException(s"Unexpected Transfer Agreement status: $transferAgreementStatus")
```
MancunianSam, line 57: Could you add the consignment id into the two in this class as well please.

MancunianSam (approved):

PR #1577, opened 2022-04-05
Rework filecheckerrormessage logic

thanhz:


app/views/partials/fileCheckErrorMessage.scala.html
```diff
@@ -1,9 +1,12 @@
+@import views.html.partials.fileCheckGenericErrorMessage
 @(fileStatuses: List[String] = Nil)
 @{
-    if(fileStatuses.size == 1 && fileStatuses.contains("PasswordProtected")) {
+    if(fileStatuses.contains("PasswordProtected") && fileStatuses.contains("Zip")){
```
TomJKing, line 4: There are actually some additional possible [file statuses](https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/service/FileStatusService.scala#L37):

- `Mismatch`
- `VirusDetected`

These will also need to be accounted for initially, as could have a combination of say, `PasswordProteced` and `Mismatch` which would need to display the generic error message. 

I think something like this should cover all the possible combinations:
```
...
@{    
    val isPasswordProtected = fileStatuses.contains("PasswordProtected")
    val isZip = fileStatuses.contains("Zip")
    val isMismatch = fileStatuses.contains("Mismatch")
    val isAntivirus = fileStatuses.contains("Antivirus")

    if((isPasswordProtected && isZip) || isMismatch || isAntivirus){
        ...  
    } else if(isPasswordProtected) {
       ...
    } else if(isZip) {
       ...                
    } else {
       ...
    }
}
...
```
The unit test will need updating as well to account for all of the possible combinations. I would suggest also in the unit tests, making sure a `Success` status is also included, to make it more realistic, ie have multiple files, some succeeding and others failing.

TomJKing (commented):
Thanks. Had a chance to think about this a bit more, and looking at it, its a bit more complicated. See comment. Hope that makes sense. This is turning into one of those very tricky little bits of functionality.

app/views/partials/fileCheckErrorMessage.scala.html
thanhz: Unless I am mistaken I think the api that fetches the filestatuses currently only has FFID errors so I think Mismatch and Antivirus won't be caught because in the api we have 
`  def getFileStatus(consignmentId: UUID): Future[Map[UUID, String]] = {
    for {
      ffidStatus <- fileStatusRepository.getFileStatus(consignmentId, FFID)
      fileStatusMap = ffidStatus.flatMap(fileStatusRow => Map(fileStatusRow.fileid -> fileStatusRow.value)).toMap
    } yield fileStatusMap
  }`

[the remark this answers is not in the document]
TomJKing: Ah fair enough, didn't notice that. In that case I'd just move the "contains" into `val`s as they are both used twice, ie:

```
val isPasswordProtected = fileStatuses.contains("PasswordProtected") 
val isZip = fileStatuses.contains("Zip")
```

TomJKing (approved):
Looks good. Thanks. Apologies this turned into a bit of a mission.

PR #1582, opened 2022-04-06
Tdr-1307 Series completed frontend

thanhz:
generated graphql depenedency will be updated after this [pr](https://github.com/nationalarchives/tdr-generated-graphql/pull/476) is merged


app/controllers/SeriesDetailsController.scala
```diff
@@ -27,35 +29,49 @@ class SeriesDetailsController @Inject()(val controllerComponents: SecurityCompon
...
     val successFunction: SelectedSeriesData => Future[Result] = { formData: SelectedSeriesData =>
-      consignmentService
-        .createConsignment(Some(UUID.fromString(formData.seriesId)), request.token)
-        .map { consignment =>
-          if (request.token.isJudgmentUser) {
-            Redirect(routes.BeforeUploadingController.beforeUploading(consignment.consignmentid.get))
-          } else {
-            Redirect(routes.TransferAgreementPrivateBetaController.transferAgreement(consignment.consignmentid.get))
-          }
+      if (request.token.isJudgmentUser) {
```
MancunianSam, line 63: This bit isn't needed. The action prevents any judgment users from getting to this page so we don't need to check for it here. You can just have lines 66 to 73 instead of the if/else

test/controllers/SeriesDetailsControllerSpec.scala
```diff
@@ -67,47 +72,35 @@ class SeriesDetailsControllerSpec extends FrontEndTestHelper {
     }
```
MancunianSam, line 72: If you make the changes above to the controller where you use the `standardTypeAction`  instead of the `standardUserAction` then it will break the tests in this class because there's an extra API call you're not mocking.

I've got them working on a branch `change-user-action` if you want to have a quick look. You don't have to do it exactly like that but it might help.

MancunianSam (commented):
Looks good thanks. Just a couple of things.

MancunianSam (approved):

PR #1583, opened 2022-04-06
Remove MANAGEMENT_ACCOUNT

MancunianSam:
This is now optional in tdr-github-actions. Once all the repos have removed it, we can take it out of that repo.


TomJKing (approved):
Looks good. Thanks

PR #1593, opened 2022-04-08
Tdr 1722 judgments faq page

TomJKing:


app/controllers/HomepageController.scala
```diff
@@ -36,7 +36,7 @@ class HomepageController @Inject()(val controllerComponents: SecurityComponents,
       } else if(request.token.isStandardUser){
         Ok(views.html.standard.homepage(request.token.name))
       } else {
-        Ok(views.html.registrationComplete(request.token.name))
+        Ok(views.html.registrationComplete(request.token.name, request.token.isJudgmentUser))
```
MancunianSam, line 39: Isn't this page only showed to judgment users? And won't `request.token.isJudgmentUser` always be false as well because where it's true is covered in the first `if` So we'd have a judgments only page with `judgmentsUser = false` passed to it.
I might be talking nonsense though.

app/controllers/UploadController.scala
```diff
@@ -36,7 +36,7 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
         case Some("Completed") =>
           uploadStatus match {
             case Some("InProgress") =>
-              Ok(views.html.uploadInProgress(consignmentId, pageHeading, request.token.name)).uncache()
+              Ok(views.html.uploadInProgress(consignmentId, pageHeading, request.token.name, request.token.isJudgmentUser)).uncache()
```
MancunianSam, line 39: This is wrapped in the `standardTypeAction` so this will always be false. The line below hard codes false in for isJudgmentUser so it might be worth doing that here.

app/controllers/UploadController.scala
```diff
@@ -65,7 +65,7 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
 
       uploadStatus match {
         case Some("InProgress") =>
-          Ok(views.html.uploadInProgress(consignmentId, pageHeading, request.token.name)).uncache()
+          Ok(views.html.uploadInProgress(consignmentId, pageHeading, request.token.name, request.token.isJudgmentUser)).uncache()
```
MancunianSam, line 68: Same here but in reverse, isJudgmentUser is always true.

app/views/judgment/judgmentFaq.scala.html
```diff
@@ -0,0 +1,258 @@
...
+            <h2 class="govuk-heading-m" id="getting-started">Getting started</h2>
+            <h3 class="govuk-heading-s">What is TDR?</h3>
+            <p class="govuk-body">
+                Transfer Digital Records (TDR) is The National Archives’ secure digital transfer service that enables Public Record Bodies to upload, prepare and then transfer selected digital public records to The National Archives (TNA) for permanent archiving and publication.
+            </p>
+            <p class="govuk-body">
+                For judgments, TDR is used to transfer the approved judgment to TNA for publication via their Case Law service.
+            </p>
+
+            <h3 class="govuk-heading-s">How does transferring a judgment record work?</h3>
```
MancunianSam, line 16: Bit random but is it worth putting ids on the heading so we can reference a particular bit of the page.
`<h3 id=how-does-it-work...` and then we can do `...gov.uk/faq#how-does-it-work`
Don't know if it's worth the effort so feel free to ignore me.

MancunianSam (commented):
Looks good. Just a couple of suggestions

app/controllers/HomepageController.scala
TomJKing: It's actually shown to any user without a user type. Given that at the moment this only applies to the judgment users, have removed this, but changed the html template to only show the judgment's faqs. Which is what I think Kirsten wants.

app/controllers/UploadController.scala
TomJKing: good point

app/controllers/UploadController.scala
TomJKing: good point

techncl (commented):
Looks good, just a few comments:

1. The landing page (AKA index.scala.html) and "notFoundError.scala" (and other error) pages where we can't determine the type of user) still have the FAQ link -  I was thinking, maybe in the main.scala, instead of the if-else, you could make it if-elseif-else and the "else if" checks if it is a standard user and the "else", just loads the page without the link? But that would require you to update the standard pages too. **Alternatively,** you could pass in another variable (a sort of "undetermined" one) and wrap the current if-else in another if-else.
2. I don't think the "signedOut" page can determine whether the user is a judgment user? Maybe we could somehow pass it information that the user is a judgment user and pass that into the controller (then into the view)? Or remove the FAQ link.

TomJKing:
@techncl There is a separate story for removing the FAQ link where the user type cannot be determined: https://national-archives.atlassian.net/browse/TDR-1862 Might be worth making a note of these comments there, and making these changes separately, as this PR is already pretty big. What do you think?

MancunianSam (approved):

PR #1607, opened 2022-04-11
Remove help and faq links

MancunianSam:
- Add new judgments help page
- Add images and bold text to the help page
- Don't show contact and FAQ if not signed in
- Add help page checks to the tests
- Add id to all header tags
- Update tests


techncl (approved):
Looks good.

PR #1608, opened 2022-04-11
Fix uploadHasCompleted and fileChecksCompleted steps by refactoring p…

techncl:
…rogress indicator function


techncl:
> Looks good. Would you be able to update the unit tests where sensible. See the unit test updates for the FAQ and Help page PRs. I'm going to put an additional card in the backlog to improve the unit test coverage for the page content, given we've had to add this additional logic to support the judgments UI.

The only issue is that the HTML in the unit tests doesn't render the stuff that's in the "Messages" file, it instead displays the raw HTML for e.g. "progressIndicator.step" instead of "Step 1 of 5".

TomJKing:
> > Looks good. Would you be able to update the unit tests where sensible. See the unit test updates for the FAQ and Help page PRs. I'm going to put an additional card in the backlog to improve the unit test coverage for the page content, given we've had to add this additional logic to support the judgments UI.

> 
> The only issue is that the HTML in the unit tests doesn't render the stuff that's in the "Messages" file, it instead displays the raw HTML for e.g. "progressIndicator.step" instead of "Step 1 of 5".

OK, fair enough. Thanks for taking a look.

TomJKing (approved):

PR #1610, opened 2022-04-11
Tdr-1838 Add transfer reference to judgment workflow

thanhz:
I have added the consignment reference to all the pages attached in this [story](https://national-archives.atlassian.net/browse/TDR-1838).
There are some other pages I am not sure whether it should have the consignment reference or not such as:

- UploadInProgress (Your upload was interrupted and could not be completed)
- Upload Completed page (When you go back after already finishing an upload)
- FileCheckProgressAlreadyConfirmed
- File-Checks completed page

### DIFF
--- a/app/controllers/FileChecksResultsController.scala
+++ b/app/controllers/FileChecksResultsController.scala
@@ -39,11 +39,14 @@ class FileChecksResultsController @Inject()(val controllerComponents: SecurityCo
   def judgmentFileCheckResultsPage(consignmentId: UUID): Action[AnyContent] = judgmentTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
     consignmentService.getConsignmentFileChecks(consignmentId, request.token.bearerAccessToken).flatMap(fileCheck => {
       if (fileCheck.allChecksSucceeded) {
-        consignmentService.getConsignmentFilePath(consignmentId, request.token.bearerAccessToken).map(files => {
-          val filename = files.files.head.metadata.clientSideOriginalFilePath
+        for {
+          filepath <- consignmentService.getConsignmentFilePath(consignmentId, request.token.bearerAccessToken)
+          ref <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
+        } yield {
+          val filename = filepath.files.head.metadata.clientSideOriginalFilePath
             .getOrElse(throw new IllegalStateException(s"Filename cannot be found for judgment upload: '$consignmentId'"))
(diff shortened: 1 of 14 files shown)


app/controllers/FileChecksResultsController.scala
```diff
@@ -39,11 +39,14 @@ class FileChecksResultsController @Inject()(val controllerComponents: SecurityCo
   def judgmentFileCheckResultsPage(consignmentId: UUID): Action[AnyContent] = judgmentTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
     consignmentService.getConsignmentFileChecks(consignmentId, request.token.bearerAccessToken).flatMap(fileCheck => {
       if (fileCheck.allChecksSucceeded) {
-        consignmentService.getConsignmentFilePath(consignmentId, request.token.bearerAccessToken).map(files => {
-          val filename = files.files.head.metadata.clientSideOriginalFilePath
+        for {
+          filepath <- consignmentService.getConsignmentFilePath(consignmentId, request.token.bearerAccessToken)
+          ref <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
```
TomJKing, line 44: Maybe put this val name out in full, ie: `reference`

app/controllers/UploadController.scala
```diff
@@ -57,11 +57,13 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
...
   }
 
-  def judgmentUploadPage(consignmentId: UUID): Action[AnyContent] = judgmentTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
+  def judgmentUploadPage(consignmentId: UUID): Action[AnyContent] = judgmentTypeAction(consignmentId) {
+    implicit request: Request[AnyContent] =>
     val consignmentStatusService = new ConsignmentStatusService(graphqlConfiguration)
 
     for {
       consignmentStatus <- consignmentStatusService.consignmentStatus(consignmentId, request.token.bearerAccessToken)
+      ref <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
```
TomJKing, line 66: Maybe put this val name out in full, ie: `reference`

TomJKing (commented):
Looks good. Thanks. Just a couple of very minor comments.

With regard to the additional pages displaying the reference, I would say yes, they should. However, this is one for Kirsten to answer. So I would ask here and Sarra about it. I would merge these changes without waiting for an answer and raise a new JIRA task if they do want the reference on the other pages. 

We need to get as much done before Tues next week as possible.

PR #1614, opened 2022-04-12
Tdr-1887/1352/1901-add transfer reference to error page

thanhz:
- Added consignment reference to error page for both standard and judgment as they share the same html file.
- Show transfer reference by default


techncl (approved):
Looks good.

PR #1616, opened 2022-04-13
Fix for service error not appearing if user skips to fileChecksResults

thanhz:


TomJKing (approved):
Looks good. Thanks

techncl (commented):
Looks good. I was thinking, maybe the `parentFolder` can be checked before the `getConsignmentRef` to save a call to the API if getting the parent folder fails?

techncl (approved):

PR #1619, opened 2022-04-14
Add Consignment Reference to standard pages & Judgment Completed Pages

thanhz:
- Add Consignment Reference to standard pages
- Add Transfer Reference to Completed Judgment Pages


MancunianSam (commented):
This looks good mate. Just a suggestion but there's a lot of instances of `reference.consignmentReference` because `getConsignmentRef` in `ConsignmentService` returns a `GetConsignment` object instead of a `String`. 

I know you haven't changed that method in this PR but I've had a look and we're not using that full object anywhere so is it worth changing `getConsignmentRef` to return `Future[String]` by returning `data.getConsignment.get.consignmentReference` and then all of the instances of `reference.consignmentReference` can be changed to just `reference`

What do you think?

thanhz:
> This looks good mate. Just a suggestion but there's a lot of instances of `reference.consignmentReference` because `getConsignmentRef` in `ConsignmentService` returns a `GetConsignment` object instead of a `String`.
> 
> I know you haven't changed that method in this PR but I've had a look and we're not using that full object anywhere so is it worth changing `getConsignmentRef` to return `Future[String]` by returning `data.getConsignment.get.consignmentReference` and then all of the instances of `reference.consignmentReference` can be changed to just `reference`
> 
> What do you think?

This sounds like a good idea! :+1:

MancunianSam (approved):

PR #1678, opened 2022-04-27
TDR-1965 Rewording of error page and add reference to the email subject

thanhz:


app/views/partials/fileCheckGenericErrorMessage.scala.html
```diff
@@ -1,6 +1,7 @@
+@(consignmentRef: String)
 <p class="govuk-body">
-    One or more files you uploaded have failed our checks. Please
-    <a class="govuk-link" href="mailto:tdr@@nationalachives.gov.uk?subject=Problem with Results of checks">Contact us</a>
+    One or more files you uploaded have failed our checks. Contact us at
+    <a class="govuk-link" href="mailto:tdr@@nationalachives.gov.uk?subject=Ref: @consignmentRef Problem with Results of checks">tdr@@nationalachives.gov.uk</a>
```
techncl, line 4: I was thinking that maybe there should be a hyphen in-between the "consignmentRef" and the "Problem with Results of checks" like `@consignmentRef  - Problem with Results of checks` just for clarity.

techncl (commented):
Looks good. There was just one comment about adding a hyphen. Also, I was thinking maybe the link could be added to the unit tests, just to make it more robust?

app/views/partials/fileCheckGenericErrorMessage.scala.html
thanhz: Yeah that would make it easier to read :+1:

thanhz:
After discussion with Kirsten we have decided to word the error as such **"One or more files you uploaded have failed our checks. Contact us at  [tdr@nationalarchives.gov.uk](mailto:tdr@nationalarchives.gov.uk) if the problem persists."**

techncl (approved):

PR #1714, opened 2022-05-06
TDR-793 - Make the "You can only drop a single folder" message more s…

vimleshtna:
…pecific


conf/messages
```diff
@@ -2,10 +2,10 @@
 
 transferAgreement.warning=This service is in private beta. During this time, the Transfer Digital Records service can only accept records that are in English, are Public Records and are Crown Copyright. For more information, please see our <a href="@routes.FaqController.faq()#other-languages" target="_blank" rel="noopener noreferrer" class="govuk-link">FAQ (opens in new tab)</a> if you need to transfer records that this service cannot currently handle.
 
-upload.dragAndDropFolderErrorMessage=You can only drop a single folder
+upload.dragAndDropFolderErrorMessage=You can only drop a single folder.
 upload.dragAndDropFileErrorMessage=You must upload a file
-upload.dragAndDropMultipleFilesErrorMessage=You must upload a single file
-upload.dragAndDropMultipleFoldersErrorMessage=You must upload a single folder
+upload.dragAndDropMultipleFilesErrorMessage=You can not upload single files. Please add your files to a folder and try uploading again.
```
TomJKing, line 7: The `upload.dragAndDropMultipleFilesErrorMessage` is used by both the standard and judgment templates. However the wording needs to be different.

For:

- **judgments:** `You must upload a single file`
- **standard:** `You can not upload single files. Please add your files to a folder and trying again`

So as a result of the changes the judgment template is displaying the incorrect wording. Would suggest splitting this out into two messages, and ensuring the correct message is called depending on the template.

TomJKing (changes requested):
Looks good. Just one comment. The error message needs to be different for standard and judgment templates.

conf/messages
vimleshtna: Done.

TomJKing (approved):
Looks good. Thanks

TomJKing (approved):
Re-approving

PR #1722, opened 2022-05-09
Tdr-625 Export consignment Report CSV

thanhz:
This is a sample of a CSV that is produced. 
Screenshot from 2022-05-09 14-39-08


thanhz:
I tried a bulk upload of 1000 files and the the downloading of the csv still worked only had to wait a second or 2 :+1:

thanhz:
There doesn't seem to be a way to display the filesize on the page because we are creating the .csv on the fly when the user clicks the link. We won't know what the size of the file will be when we first land on the page.

app/services/ConsignmentService.scala
```diff
@@ -100,4 +102,11 @@ class ConsignmentService @Inject()(val graphqlConfiguration: GraphQLConfiguratio
     sendApiRequest(updateConsignmentSeriesIdClient, updateConsignmentSeriesId.document, token, variables)
       .map(data => data.updateConsignmentSeriesId.isDefined)
   }
+
+  def getConsignmentExport(consignmentId: UUID, token: BearerAccessToken): Future[getConsignmentForExport.GetConsignment] = {
```
TomJKing, line 106: Should add some unit tests for this new method to the `ConsignmentServiceSpec.scala` as per the other methods.

app/controllers/TransferCompleteController.scala
```diff
@@ -1,16 +1,16 @@
...
+import play.api.mvc.{Action, AnyContent, Request}
 import services.ConsignmentService
 
-import scala.concurrent.{ExecutionContext, Future}
+import java.time.{LocalDateTime, ZonedDateTime}
+import java.util.UUID
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
 
 class TransferCompleteController @Inject()(val controllerComponents: SecurityComponents,
```
TomJKing, line 15: I'd add additional test to check that the new HTML for the download report is appearing. See line 50

TomJKing (commented):
Looks good. Just some comments around testing.

app/controllers/TransferCompleteController.scala
thanhz: ah yes I forgot about those one. I've added a test for that now

app/services/ConsignmentService.scala
thanhz: I've added some test for the getConsginmentExport method

TomJKing (approved):
Looks good. Thanks

PR #1739, opened 2022-05-11
Add GET route for judgment homepage

thanhz:
When the users session expires and they try to click the start transfer button they are directed to /judgment/homepage as a GET request but this route does not exist as a GET request but rather a POST request. This PR adds this route and makes sure the url is displayed correctly for judgment homepage.

### DIFF
--- a/test/controllers/HomepageControllerSpec.scala
+++ b/test/controllers/HomepageControllerSpec.scala
@@ -84,6 +84,16 @@ class HomepageControllerSpec extends FrontEndTestHelper {
       contentAsString(homepagePage) must include (s"""" href="/judgment/help">""")
     }
 
+    "return a redirect to the judgment homepage page with an authenticated judgment" in {
+      val controller = new HomepageController(
+        getAuthorisedSecurityComponents,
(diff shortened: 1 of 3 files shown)


test/controllers/HomepageControllerSpec.scala
```diff
@@ -71,7 +71,7 @@ class HomepageControllerSpec extends FrontEndTestHelper {
         getAuthorisedSecurityComponents,
         getValidJudgmentUserKeycloakConfiguration,
         consignmentService)
-      val homepagePage = controller.homepage().apply(FakeRequest(GET, "/homepage").withCSRFToken)
+      val homepagePage = controller.judgmentHomepage().apply(FakeRequest(GET, "/judgment/homepage").withCSRFToken)
       status(homepagePage) mustBe OK
```
MancunianSam, line 75: Is it worth adding a test in to check that if you go to '/homepage` as a judgment user, you get redirected to `/judgment/homepage` ?

MancunianSam (commented):
Looks good, just a suggestion for the extra test, feel free to ignore me.

PR #1740, opened 2022-05-12
Update judgment help to remove MFA config 'device name' instructions

TomJKing:
'device name' input box removed from MFA config page as was causing confusion to users

Help page guidance updated to reflect this change


MancunianSam (approved):

thanhz:
Just a heads up that I will probably be replacing the image in my PR to reflect the reorder of the of the authenticator

PR #1746, opened 2022-05-12
TDR-1979 - Move batching logic from frontend to API to fix the "Dupli…

vimleshtna:
…cate folder entries created in DB" issue


TomJKing (approved):
Looks good. Thanks

PR #1747, opened 2022-05-13
Tdr 37 tag uploaded s3 objects with file path

techncl:


TomJKing (approved):
Looks good. Are there any other changes required in other repos to get this to work?

techncl:
> Looks good. Are there any other changes required in other repos to get this to work?

No, I don't think so, when I run this locally, it tags the file on S3 with the path. There's a task in this sprint (TDR-1946) which is going to update the lambda code to use this information.

TomJKing (approved):
Looks good. Thanks

PR #1770, opened 2022-05-20
TDR-2003 - Change frontend to check for "InProgress" instead of "Comp…

vimleshtna:
…leted" on transfer already completed page


app/controllers/ConfirmTransferController.scala
```diff
@@ -54,7 +54,7 @@ class ConfirmTransferController @Inject()(val controllerComponents: SecurityComp
       consignmentStatus =>
         val exportTransferStatus = consignmentStatus.flatMap(_.export)
         exportTransferStatus match {
-          case Some("Completed") => Future(Ok(views.html.transferAlreadyCompleted(consignmentId, request.token.name)).uncache())
+          case Some("InProgress") => Future(Ok(views.html.transferAlreadyCompleted(consignmentId, request.token.name)).uncache())
```
TomJKing, line 57: We've now got three potential status after the user clicks the export/transfer button:

- `InProgress`
- `Completed`
- `Failed`

I think for any of these statuses the user needs to be taken to the completed pages. @techncl What do you think?

TomJKing (commented):
Looks good. I just have one question on this. See comment. I've included Nigel in the comment to see his thoughts on this.

app/controllers/ConfirmTransferController.scala
techncl: Ah yes, because the user could possibly come back to the page after the export has completed/failed, right? So I guess line 57 could/should be `case Some("InProgress") | Some("Completed") | Some("Failed")`?

TomJKing: @techncl Yes. I think that is correct.

vimleshtna: Thanks Tom and Nigel. I have updated the code it as per the suggestion.

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -471,28 +478,30 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
...
+        playStatus(transferControllerPage) mustBe OK
+        contentType(transferControllerPage) mustBe Some("text/html")
+        headers(transferControllerPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
+        confirmTransferPageAsString must include(
+          s"""href="/consignment/$consignmentId/transfer-complete">Continue""".stripMargin)
+        confirmTransferPageAsString must include("Your transfer has already been completed")
+      }
     }
 
     "render the confirm transfer 'already confirmed' page with an authenticated user if the standard user navigates back to the " +
```
TomJKing, line 500: Should not this test be updated to cover all the statuses: `Failed`, `InProgress`, `Completed` as with the other tests

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -527,28 +536,30 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
...
+        playStatus(transferAlreadyCompletedPage) mustBe OK
+        contentType(transferAlreadyCompletedPage) mustBe Some("text/html")
+        headers(transferAlreadyCompletedPage) mustBe TreeMap("Cache-Control" -> "no-store, must-revalidate")
+        transferAlreadyCompletedPageAsString must include(
+          s"""href="/judgment/$consignmentId/transfer-complete">Continue""".stripMargin)
+        transferAlreadyCompletedPageAsString must include("Your transfer has already been completed")
+      }
     }
 
     "render the confirm transfer 'already confirmed' page with an authenticated judgment user if the user navigates back to the " +
```
TomJKing, line 558: Should not this test be updated to cover all the statuses: `Failed`, `InProgress`, `Completed` as with the other tests

TomJKing (commented):
Looks good. Couple of additional comments

TomJKing (approved):
Looks good. Thanks

PR #1788, opened 2022-05-24
Tdr 2025 improve test coverage on file checks results page

techncl:


app/views/fileChecksFailed.scala.html
```diff
@@ -10,7 +10,7 @@
         } else {
             @progressIndicator(Messages("checkingRecords.progress"), isJudgmentUser)
         }
-      <h1 class="govuk-heading-l">Results of your checks</h1>
+      <h1 class="govuk-heading-l">Results of checks</h1>
```
TomJKing, line 13: Was this change of page heading signed off, or has it always been incorrect and no one noticed?

TomJKing (commented):
Looks good. One minor comment

app/views/fileChecksFailed.scala.html
techncl: I'm not actually sure. What I'll do is pass in the title via the controller so that this "failed" page displays one that is consistent with the success page's title and so that it's consistent depending on the type of user.

TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):

PR #1804, opened 2022-05-29
Tdr-1986 Dynamic imports

thanhz:
The index.ts has been refactored to use conditional dynamic formatting. This his reduced the size of the main.js as seen in the screenshot.
BeforeAfter

It seems however that implementing dynamic imports to any of the other typescript files provides no reduction to the size of the main.js and makes the code a bit harder to read and thus those changes haven't been included.


MancunianSam (commented):
Looks good, that's a big reduction in the js file size. Just a couple of comments but if they don't make sense or make it harder to read then ignore me.

PR #1844, opened 2022-06-10
Tdr-1130 session expire dialog

thanhz:
Screenshot from 2022-06-16 17-46-21


thanhz:
Should the message be changed to `"You have been inactive for more than 55 Minutes"` since the session expiry is currently set to 1 hour?

thanhz:
As per comment on the [jira card](https://national-archives.atlassian.net/browse/TDR-1108) the message has been changed to 55 minutes to reflect the 1 hour session timeout.

thanhz:
As per comment on the [jira card](https://national-archives.atlassian.net/browse/TDR-1108) the message has been changed to 55 minutes to reflect the 1 hour session timeout.

npm/src/index.ts
```diff
@@ -38,6 +38,8 @@ export const renderModules = async () => {
   const fileChecksContainer: HTMLDivElement | null = document.querySelector(
     ".file-check-progress"
   )
+  const timeoutDialog: HTMLDialogElement | null =
```
TomJKing, line 41: The `HTMLDialogElement` is showing its deprecated. Having a quick look, it appears this is a bug with Typescript: https://github.com/microsoft/TypeScript/issues/48267 Could you double check that this is the case.

npm/src/index.ts
```diff
@@ -106,5 +108,49 @@ export const renderModules = async () => {
     } else {
       errorHandlingModule.handleUploadError(frontEndInfo)
     }
+  } else if (timeoutDialog) {
+    const authModule = await import("./auth")
```
TomJKing, line 112: There is some complicated logic in this function. Would it be possible to abstract this out into the existing `auth/index.ts` file. Or maybe a new file `àuth/sesssiontimeout.ts`? Can then add some more unit testing around the functionality

app/views/main.scala.html
```diff
@@ -73,6 +73,21 @@
         </main>
     </div>
 
+    @if(isLoggedIn) {
+    <dialog class="timeout-dialog">
```
TomJKing, line 77: At the moment on the `/upload` page the token is continuously updated, for 12 hours, to ensure a user does not get logged out whilst their files are being uploaded. There is a question as to whether:

1. this behaviour should be retained; or
2. it should be removed, as they would then get this session timeout warning.

I think this question should be asked to the wider team, including Kirsten and Pauline.

npm/test/index.test.ts
```diff
@@ -91,3 +91,28 @@ test("renderModules does not call authorisation when the front end info is missi
 
   keycloakInstance.mockRestore()
 })
+
```
TomJKing, line 94: Probably need some more extensive unit tests. See suggestion about, about abstracting the functionality into separate functions to make unit testing easier.

TomJKing (commented):
Looks good. Just some comments around unit testing, and other stuff.

app/views/main.scala.html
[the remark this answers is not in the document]
thanhz: I have tested the screenreader for the timeout message. When the model appears the screenreader will read out the message and the options.

npm/src/index.ts
thanhz: The HTMLDialogElement isn't showing as deprecated for me. It does say experimental, but from what I have gathered many browsers do support this. I have also tested it out with various browsers myself.

app/views/main.scala.html
thanhz: As discussed today. We will be retaining the behavior of the upload page.

npm/src/index.ts
thanhz: As suggested, I have moved the session timeout logic to its own file. Still working on the tests though

npm/test/index.test.ts
thanhz: Tests finally working

npm/test/session-timeout.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  const mockUpdateToken = jest
+    .fn()
+    .mockImplementation((_: number) => new Promise((res, _) => res(true)))
+  const isTokenExpired = false
+  const now = new Date()
+  const refreshTokenParsed: KeycloakTokenParsed = {
+    exp: Math.round(now.getTime() / 1000) + 300
+  }
+
+  const mockKeycloak: Keycloak.KeycloakInstance = createMockKeycloak(
```
TomJKing, line 71: This is repeated code. Could it be moved out of the functions?

npm/test/session-timeout.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  const mockUpdateToken = jest
+    .fn()
+    .mockImplementation((_: number) => new Promise((res, _) => res(true)))
+  const isTokenExpired = false
+  const now = new Date()
+  const refreshTokenParsed: KeycloakTokenParsed = {
+    exp: Math.round(now.getTime() / 1000) - 300
+  }
+
+  const mockKeycloak: Keycloak.KeycloakInstance = createMockKeycloak(
```
TomJKing, line 103: This is repeated code. Could it be moved out of the functions?

TomJKing (commented):
Looks good. Thanks for getting those jest tests in place. Just one very minor comment.

TomJKing (approved):
Looks good. Thanks

PR #1880, opened 2022-06-23
Merge judgment and standard export status tests

techncl:


MancunianSam (approved):
That's much tidier, thanks.

PR #1906, opened 2022-07-04
Tdr 1923 update standard help guide

TomJKing:


app/views/help.scala.html
```diff
@@ -10,81 +10,231 @@ <h2 class="govuk-heading-m">This user help guide includes:</h2>
...
+                Please be aware that at this stage, we can only accept records that are Crown Copyright, in English, open (i.e. no FOI exemptions apply) and from your shared drive.
             </p>
             <p class="govuk-body">
                 Over time, we will extend the service to accept a wider range of records such as those in a different language, closed records and other complex types of records.
             </p>
+
+            <h3 class="govuk-heading-s" id="before-web-browsers">Web browsers</h3>
             <p class="govuk-body">
-                To ensure you have the best experience, and can use all of our available features, we recommend you use the latest version of Google Chrome. Please be aware that Internet Explorer is not currently supported at all. 
+                We do not support Internet Explorer (IE11).
```
techncl, line 83: Maybe we should exclude "(IE11)" just in case someone thinks we support earlier versions of Internet Explorer?

app/views/help.scala.html
```diff
@@ -10,81 +10,231 @@ <h2 class="govuk-heading-m">This user help guide includes:</h2>
...
+                We advise you to use a different browser such as Edge, Chrome, Safari or Firefox to ensure you can use this service and all of its features.
```
techncl, line 84: Should we add Safari and a browser we advise them using? I remember it having a few weird issues.

app/views/help.scala.html
```diff
@@ -10,81 +10,231 @@ <h2 class="govuk-heading-m">This user help guide includes:</h2>
...
+
+            <h3 class="govuk-heading-s" id="secured-access">Secured access</h3>
+            <p class="govuk-body">
+                In order to keep the service and your account secure from attack, TDR uses Multi-factor Authentication (MFA), which needs two separate pieces of information in order to authenticate and access the service.
+                This means you will be required to sign in each time using a unique generated 'one-time password' (OTP) code that confirms who you are.
+                This can be set up with a free-to-download authenticator app on your smartphone, such as Google Authenticator, Microsoft Authenticator, or FreeOTP.
+                Other authenticators are available.
+            </p>
+            <p class="govuk-body">
+                Free-to-download apps are available from the App Store for iPhone, or 'Google Play' and 'Play Store' for Android.
```
techncl, line 95: 'Google Play' and 'Play Store' are synonyms; maybe we can call it just 'Google Play Store' or "'Google Play' ('Play Store')"

app/views/help.scala.html
```diff
@@ -100,146 +250,203 @@ <h3 class="govuk-heading-s">Step two: Transfer agreement</h3>
...
-            </p>
-            <ul class="govuk-list govuk-list--bullet">
-                <li>You can upload one file or one top-level folder per transfer. However, your folder can contain multiple files and sub-folders</li>
-                <li>We cannot accept files that are password protected or encrypted</li>
-                <li>During private beta, we cannot accept zip files</li>
-                <li>Files such as thumbnail images (thumbs.db) and system files are not usually selected for permanent preservation. We strongly advise you do not upload these</li>
-                <li>During the upload process, TDR runs DROID to capture basic file-level metadata. Such as file path, file name, date (usually last modified), file format, and checksum</li>
-            </ul>
-            <img src="@routes.Assets.versioned("images/upload.png")" class="screenshots" alt="Upload page with a multi-select area for dropping your folder or selecting through a button." />
+                The transfer agreement is in two parts.
```
techncl, line 253: Lowercased t for "transfer" instead of uppercase

app/views/help.scala.html
```diff
@@ -100,146 +250,203 @@ <h3 class="govuk-heading-s">Step two: Transfer agreement</h3>
...
+            </p>
+            <ol class="govuk-body">
+                <li>The first part covers the requirements that records must be public, crown copyright and in English.
+                    Click each box to confirm and then 'Continue' when ready to proceed.</li>
+                <img src="@routes.Assets.versioned("images/standard-TA-1.png")" class="screenshots" alt="Transfer Agreement part 1 page" />
+                <li>The second part of the transfer agreement is to confirm that the Departmental Records Officer (DRO) has signed off the sensitivity review, appraisal and selection decision, and that the records are open for publication.
+                    (Closed records cannot currently be accepted whilst TDR is operating in private beta phase.)
```
techncl, line 260: I think the full stop after "publication" might be a mistake as well as the full stop after "beta phase"; I think it's supposed to be after the closing bracket

app/views/help.scala.html
```diff
@@ -100,146 +250,203 @@ <h3 class="govuk-heading-s">Step two: Transfer agreement</h3>
...
+
+            <h3 class="govuk-heading-s" id="upload-records">Step three: Upload your records</h3>
             <p class="govuk-body">
-                Please note, when you upload your records, you still have legal custody over them and are still responsible for any FOI requests.
+                Before you begin, ensure your internet connection is stable and strong.
             </p>
-            <h3 class="govuk-heading-s">Step four: Upload and file checks</h3>
+            <ol class="govuk-body">
+                <img src="@routes.Assets.versioned("images/standard-upload-records.png")" class="screenshots" alt="Upload records page" />
+                <li>To upload your records, you can either drag and drop a file or folder OR click 'Choose Folder' to browse and select a folder for upload.
```
techncl, line 271: This is for standard users only, right? So maybe we should remove "a file or"

app/views/help.scala.html
```diff
@@ -100,146 +250,203 @@ <h3 class="govuk-heading-s">Step two: Transfer agreement</h3>
...
+                <p class="govuk-body">The checking process is likely to take several minutes.</p>
+                <img src="@routes.Assets.versioned("images/standard-checks-complete.png")" class="screenshots" alt="Checks complete page" />
+                <li>On completion, you will be advised on screen when you can proceed to the next step.
+                    Click 'Continue' to see the results of your checks.</li>
+                <img src="@routes.Assets.versioned("images/standard-results-check.png")" class="screenshots" alt="Results of checks page" />
+                <li>You will be informed on screen of the results of your checks. </li>
+                <li>If your checks have been successful, click 'Continue' to progress to the next step.</li>
+            </ol>
+            <p class="govuk-body">If your files fail TDR checks, we will inform you.
+                Failing checks may mean you cannot proceed with your transfer until the problems are resolved.
```
techncl, line 330: Should it be "Failing checks *will* mean" instead of "Failing checks *may* mean"? Or are there actually some failed checks that we will allow?

TomJKing: At some future point the users will be able to remove files that fail from the consignment for certain reasons. I think this is fine to leave as is for now.

app/views/help.scala.html
TomJKing: This is a question for Kirsten/Linda/Pauline. Added it to the JIRA task

app/views/help.scala.html
TomJKing: Do you mean remove Safari from the recommended list of browsers? If so I think this is a question for Kirsten/Linda/Pauline. Added it to the JIRA task

app/views/help.scala.html
TomJKing: This is a question for Kirsten/Linda/Pauline. Added it to the JIRA task

app/views/help.scala.html
TomJKing: Removed as per Linda's comment

app/views/help.scala.html
TomJKing: Removed as per Linda's comment

app/views/help.scala.html
TomJKing: Changed to read "Google Play Store"

techncl (approved):
Looks good! Thanks.

PR #1922, opened 2022-07-07
Proxy API requests through the front end.

MancunianSam:
* Add controller endpoints to replace the API endpoints.
* Add a service to call the original API endpoints.
* Add tests for the new controller and service methods.
* Update the javascript to call the front end instead of the API
* Update the jest tests to work with the new fetch calls
* Remove apollo client and the graphql module.


app/controllers/UploadController.scala
```diff
@@ -1,27 +1,56 @@
...
+  def startUpload(): Action[AnyContent] = secureAction.async { implicit request =>
+    request.body.asJson.flatMap(body => {
+      decode[StartUploadInput](body.toString).toOption
+    }) match {
+      case None => Future.failed(new Exception(s"Incorrect data provided ${request.body}"))
+      case Some(input) => uploadService.startUpload(input, request.token.bearerAccessToken).map(Ok(_))
+    }
+  }
+
+  def saveMetadata(): Action[AnyContent] = secureAction.async { implicit request =>
```
TomJKing, line 45: Maybe rename this: `saveClientMetadata`

app/controllers/UploadController.scala
```diff
@@ -62,29 +91,29 @@ class UploadController @Inject()(val controllerComponents: SecurityComponents,
...
-    } yield {
-      val uploadStatus: Option[String] = consignmentStatus.flatMap(_.upload)
-      val pageHeading1stHalf = "Upload judgment"
-      val pageHeading2ndHalf = "Uploading judgment"
+      for {
+        consignmentStatus <- consignmentStatusService.getConsignmentStatus(consignmentId, request.token.bearerAccessToken)
+        reference <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
+      } yield {
+        val uploadStatus: Option[String] = consignmentStatus.flatMap(_.upload)
+        val pageHeading1stHalf = "Upload judgment"
```
TomJKing, line 101: Maybe rename these vals:

- `pageHeadingUpload`
- `pageHeadingUploading`

app/services/UploadService.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  def startUpload(startUploadInput: StartUploadInput, token: BearerAccessToken): Future[String] = {
+    val variables = su.Variables(startUploadInput)
+    sendApiRequest(startUploadClient, su.document, token, variables).map(data =>
+      data.startUpload
+    )
+  }
+
+  def saveMetadata(addFileAndMetadataInput: AddFileAndMetadataInput, token: BearerAccessToken): Future[List[afam.AddFilesAndMetadata]] = {
```
TomJKing, line 36: Maybe rename: `saveClientMetadata`

test/services/UploadServiceSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.UUID
+import scala.concurrent.{ExecutionContext, Future}
+
+class UploadServiceSpec extends AnyFlatSpec {
+  implicit val ec: ExecutionContext = ExecutionContext.global
+  private val graphQlConfig = mock[GraphQLConfiguration]
+  private val token = new BearerAccessToken("some-token")
+
+  "startUpload" should "returns the correct data" in {
```
TomJKing, line 25: Reword slightly: `... return the correct data`

test/services/UploadServiceSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    val graphQlResponse =
+      GraphQlResponse(Some(su.Data("ok")),Nil)
+    when(graphQlClientForStartUpload.getResult(token, su.document, Some(su.Variables(input))))
+      .thenReturn(Future.successful(graphQlResponse))
+
+    val response = new UploadService(graphQlConfig).startUpload(input, token).futureValue
+    response should equal("ok")
+  }
+
+  "saveMetadata" should "returns the correct data" in {
```
TomJKing, line 40: Reword slightly: `... return the correct data`

app/services/UploadService.scala
```diff
@@ -0,0 +1,42 @@
...
+    val variables = ucs.Variables(consignmentStatusInput)
+    sendApiRequest(updateConsignmentStatusClient, ucs.document, token, variables).map(data => {
+      data.updateConsignmentStatus match {
+        case Some(response) => response
+        case None => throw new RuntimeException(s"No data returned when updating the consignment status for ${consignmentStatusInput.consignmentId}")
+      }
+    })
+  }
+
+  def startUpload(startUploadInput: StartUploadInput, token: BearerAccessToken): Future[String] = {
```
TomJKing, line 29: Is it possible to have no response here? If so, should an error be thrown?

TomJKing (commented):
Looks good. Some minor comments.

Also when testing locally seeing the following error when trying to upload a judgment. Standard uploads working OK.

Screenshot from 2022-07-11 07-49-43

MancunianSam:
I only tested this for a standard user and forgot there's different views for the judgment users. The CSRF input was missing so I've added that in and added in a missing entry to the routes file which was causing an error further on. You can now do a complete judgment export.

app/services/UploadService.scala
MancunianSam: The `ApiErrorHandling` object calls `result.data.get` so if data was optional, we'd get a (slightly ugly) error and it also throws a `GraphQlException` if the API populates the errors field. I think that covers everything that could happen. What do you think?

TomJKing (approved):
Looks good. Thanks

PR #1928, opened 2022-07-11
Add additional logging.

MancunianSam:
This uses play filters https://www.playframework.com/documentation/2.8.x/ScalaHttpFilters

The /assets and /keycloak paths were generating a lot of noise so I've
removed them. We can always add them back in later.


TomJKing (approved):
Looks good. Thanks

PR #1935, opened 2022-07-12
Fix logging for logged out users.

MancunianSam:
This was calling `getProfile.get` but `getProfile` is empty if the user
is not logged in.

This just uses maps on the option so we get something sensible for a
logged out user.


TomJKing (approved):
Looks good. Thanks

PR #1950, opened 2022-07-15
Prevent the attempted redirect when unauthorised.

MancunianSam:
The Ajax requests to the new front end endpoints were being redirected when the user was logged out. This is the standard behaviour for the browser flow but makes no sense here.

Pac4j will return 401 instead of redirecting if it recognises the request as ajax and it does this by checking the header. 
https://www.pac4j.org/docs/clients.html#5-ajax-requests

I've added the header to each of the requests.

Also, the custom request handler was redirecting even when pac4j was sending the correct 401 response. I can't remember what this class is for and it doesn't seem to do anything but I don't want to remove it just in case so I've put in an exception for these ajax requests.

They now return 401 for unauthorised although I'm not 100% sure how they got to be logged out as they should have been warned by the modal.


MancunianSam:
Yeah, I think that timeout was causing the problem so this change should hopefully never be used.

PR #1985, opened 2022-07-29
Tdr 2089 file navigation without pagination

thanhz:
The file selection page for additional metadata without any pagination.
We are currently able to select and deselect files.

Current problems include 
- No ability to go back once you've gone into a subfolder. (Whilst it may be possible to implement, the html is starting to get harder to read as it will require additional hidden fields.)
- You can tick a folder, however if you go into the subfolder none of the files will be checked as there is no way at the moment to know that the parent has already been ticked and to get it's decedents.
- It was decided that the files shouldn't be link as we were under the impression that the link was supposed to allow users to edit the metadata. But after looking at the prototype. These should be buttons as they should allow you to see what metadata is already captured for the file. This would mean that the page it leads to will need to process the form because we need to retain the users current selections in addition to passing in the id of the file to view it's metadata..

Screenshot 2022-07-28 at 15-57-20 Additional Metadata


app/controllers/AdditionalMetadataController.scala
```diff
@@ -25,4 +45,106 @@ class AdditionalMetadataController @Inject () (val consignmentService: Consignme
...
+  private def handleDeselection(selected: List[String], allNodes: List[String], previouslySelectedIds: String): Set[String] = {
+    //Filter out selected Nodes
+    val allNodesWithoutSelection = allNodes.filter(x => !selected.contains(x)).mkString(",")
+    //Remove the Filtered Nodes from the PreviouslySelected
+    previouslySelectedIds.split(",").toSet.filter(previouslySelectedId => !allNodesWithoutSelection.contains(previouslySelectedId))
+  }
+
+  private def generateNodesToDisplay(edges: List[PaginatedFiles.Edges], ids: Set[String]): List[NodesToDisplay] = {
+    edges.map(edge => {
+      val isSelected = ids.contains(edge.node.fileId.toString)
```
TomJKing, line 131: I would create a new val of: `edge.node` as this is being called in multiple places in this function.

app/controllers/AdditionalMetadataController.scala
```diff
@@ -25,4 +45,106 @@ class AdditionalMetadataController @Inject () (val consignmentService: Consignme
...
+    val formValidationResult: Form[NodesFormData] = navigationForm.bindFromRequest()
+    formValidationResult.fold(
+      errorFunction,
+      successFunction
+    )
+  }
+
+  private def handleDeselection(selected: List[String], allNodes: List[String], previouslySelectedIds: String): Set[String] = {
+    //Filter out selected Nodes
+    val allNodesWithoutSelection = allNodes.filter(x => !selected.contains(x)).mkString(",")
```
TomJKing, line 124: Just a style thing, but I wouldn't use `x` as a val. Would prefer to use `s` here given its a string.

TomJKing (commented):
Just some initial comments. Mostly about code organisation

thanhz:
I'm going to create a new PR with the suggestions. As that one will fix the issues folder selection and urls

PR #2013, opened 2022-08-05
Tdr 2171 add dynamic form validation

techncl:


app/controllers/AddClosureMetadataController.scala
```diff
@@ -2,50 +2,92 @@ package controllers
...
-                                       propertyName: Set[String],
-                                       valueNameWithDependencies: String): Set[CustomMetadata] = {
+                                       propertyName: ListSet[String],
+                                       valueToGetDependenciesFrom: String): ListSet[CustomMetadata] = {
     val valuesByProperties: Map[String, List[CustomMetadata.Values]] = customMetadataUtils.getValuesOfProperties(propertyName)
     val allValuesForProperty: Seq[CustomMetadata.Values] = valuesByProperties(propertyName.head)
-    val valueWithDependencies: Option[CustomMetadata.Values] = allValuesForProperty.find(_.value == valueNameWithDependencies)
-    val dependencyNames: Seq[String] = valueWithDependencies.map(_.dependencies.map(_.name)).getOrElse(Nil)
-    customMetadataUtils.getCustomMetadataProperties(dependencyNames.toSet)
+    val value: CustomMetadata.Values = allValuesForProperty.find(_.value == valueToGetDependenciesFrom).get
```
MancunianSam, line 89: We should try and avoid calling `get` on an Option where we can help it, even if we're sure there's something in there. You can get rid of the get by changing `find` to `filter` and flatMapping. Something like 

```scala
    val value: Seq[CustomMetadata.Values] = allValuesForProperty.filter(_.value == valueToGetDependenciesFrom)
    val dependencyNames: Seq[String] = value.flatMap(_.dependencies.map(_.name))
```

app/controllers/AddClosureMetadataController.scala
```diff
@@ -2,50 +2,92 @@ package controllers
...
-        value = "closed_for"
+        orderedFieldsForForm <- { // currently not ordered, but will be soon; please delete this comment when it is
+          cache.set(s"$consignmentId", consignmentRef, 1.hour)
+          getDefaultFieldsForForm(consignmentId, request)
+        }
+      }
+        yield Ok(views.html.standard.addClosureMetadata(consignmentId, consignmentRef, orderedFieldsForForm, request.token.name))
+  }
+
+  def addClosureMetadataSubmit(consignmentId: UUID): Action[AnyContent] = standardTypeAction(consignmentId) {
```
MancunianSam, line 40: I guess you'll add tests in for this endpoint once everything is completed?

app/controllers/util/DynamicFormUtils.scala
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    fieldValueBelongingToMetadataName match {
+      case Some((fieldValue, _)) =>
+        validateFieldType(inputName, fieldType, metadataName, inputInfo, valueSelected, fieldValue)
+      case None => throw new IllegalArgumentException(s"Metadata name $metadataName, extracted from field input name, does not exist.")
+    }
+  }
+
+  private def validateFieldType(inputName: String, fieldType: String, metadataName: String,
```
MancunianSam, line 103: There are text fields in the database, are they covered in this method?

test/controllers/AddClosureMetadataControllerSpec.scala
```diff
@@ -180,3 +185,17 @@ class AddClosureMetadataControllerSpec extends FrontEndTestHelper {
             Values("mock code1", List()), Values("mock code2", List())))))
   }
 }
+
+case class MockAsyncCacheApi()(implicit val ec: ExecutionContext) extends AsyncCacheApi {
```
MancunianSam, line 189: I'm not sure if this will be useful later on so feel free to ignore me but you don't need a full case class here, you can use Mockito to pass in a mocked `AsyncCacheApi` so the controller constructor becomes
```scala
    new AddClosureMetadataController(
     ...
      customMetadataService,
      mock[AsyncCacheApi]
    )
```
If you do need to mock the responses from calls to the cache API then that can be done.

test/controllers/util/DynamicFormUtilsSpec.scala
```diff
@@ -0,0 +1,519 @@
...
+case class MockFormValues(day: List[String] = List("3"),
+                          month: List[String] = List("4"),
+                          year: List[String] = List("2021"),
+                          numericTextBoxValue: List[String] = List("5"),
+                          dropdownValue: List[String] = List("TestValue 3"),
+                          radioValue: List[String] = List("Yes"),
+                          day2: List[String] = List("7"),
+                          month2: List[String] = List("9"),
+                          year2: List[String] = List("2022"),
+                          csrfToken: List[String] = List("12345"))
```
MancunianSam, line 519: Can you put a line break at the end of the file please.

MancunianSam (commented):
Looks good thanks. Just a couple of small comments and questions.

app/controllers/AddClosureMetadataController.scala
techncl: Yes.

app/controllers/util/DynamicFormUtils.scala
techncl: No, they aren't. I haven't made any checks for input text yet since we do not have a `text` input box on any pages yet (just numeric and date); I was planning on adding it/letting someone else add it when I/they get to it as it's a little more involved and requires more thought (UTF-8 checking, escaping characters?). For now, it will throw an exception forcing someone to create an implementation for it. Should I add  a check for plain text now?

MancunianSam: I think that's best. Just put something in that allows any text. We can worry about input validation later once we know what to do. It would be good to get this in a state where we can put any valid entries in the database and it will work.

test/controllers/AddClosureMetadataControllerSpec.scala
techncl: Great! Thanks.

MancunianSam (approved):
:+1:

PR #2017, opened 2022-08-08
Order fields based on ordinal.

MancunianSam:
The API has this as a mandatory field that returns Int.MaxValue if it's
not set in the database.

I've added in a test for the sorting and updated the existing tests with
the new case class parameter.

Needs https://github.com/nationalarchives/tdr-generated-graphql/pull/619 to be merged first.


techncl (dismissed):
Looks good! Just one comment (feel free to ignore), do you think it worth adding more numbers (than just "1") as the Ordinal value in the tests? Just so whoever is reading this in the future will know that these numbers will not just be "1" but a range.

MancunianSam:
Good idea. I've updated the tests to show sequential order numbers.

techncl (approved):

PR #2018, opened 2022-08-08
Style buttons as links for navigation page

TomJKing:
Styling for pagination:
Screenshot from 2022-08-08 14-26-55


thanhz (commented):
Tried this on my local. Looks great. Thanks :+1: 
Was wondering if we could increase the clickable area for the number links? but no worries if it becomes too much of a hassle.

MancunianSam (approved):
Looks good, I don't know CSS but I tried it and it looks like a link :+1:

PR #2026, opened 2022-08-09
TDR 2236 add file totals

MancunianSam:
- Add file totals
- Update generated-graphql dependency


thanhz (approved):
Tested Locally, looks good! Thanks for doing this :+1:

PR #2032, opened 2022-08-10
TDR-2112 add file status

vimleshtna:
call `addFileStatus` mutation with 'Failed' status if the s3 file upload fails


test/controllers/UploadControllerSpec.scala
```diff
@@ -467,11 +484,85 @@ class UploadControllerSpec extends FrontEndTestHelper {
     }
   }
 
+  "UploadController addFileStatus" should {
+    "call the addFileStatus endpoint" in {
+      val graphQLConfiguration: GraphQLConfiguration = new GraphQLConfiguration(app.configuration)
+      val client = graphQLConfiguration.getClient[addFileStatus.Data, addFileStatus.Variables]()
+      val consignmentService: ConsignmentService = new ConsignmentService(graphQLConfiguration)
+      val fileStatusService = new FileStatusService(new GraphQLConfiguration(app.configuration))
```
MancunianSam, line 492: You can reuse the `graphQLConfiguration` variable you've got here instead of calling `new GraphQLConfiguration(app.configuration)`
It's the same for the other two tests as well.

test/controllers/UploadControllerSpec.scala
```diff
@@ -433,9 +448,10 @@ class UploadControllerSpec extends FrontEndTestHelper {
     "throw an error for invalid input data" in {
       val graphQLConfiguration: GraphQLConfiguration = new GraphQLConfiguration(app.configuration)
       val consignmentService: ConsignmentService = new ConsignmentService(graphQLConfiguration)
+      val fileStatusService = new FileStatusService(new GraphQLConfiguration(app.configuration))
```
MancunianSam, line 451: This is the same as the comment below but you can reuse the `graphQLConfiguration` variable here. I know you didn't add it in but if you could also update the ones for the `uploadService` variable as we're calling `new GraphqlConfiguration` there as well. It keeps things a bit tidier.

MancunianSam (commented):
Looks good thanks, just a couple of small tidying up things.

test/controllers/UploadControllerSpec.scala
vimleshtna: Sorry, I missed this. I have updated it, thanks.

test/controllers/UploadControllerSpec.scala
vimleshtna: Updated for `uploadService` too.

MancunianSam (approved):

PR #2038, opened 2022-08-11
Change the frontend to allow a /27 subnet mask.

MancunianSam:
The private subnet which the apps sit in is on a /28 which is only 11
usable addresses. We're hitting this limit at the moment because there's
two database instances running but even when we drop the original
databases, we still get problems if there are two deployments and the
bastion is running so we need change this.

Play needs every possible address from the load balancer health checks
so this is every address from two AZs worth of /27 IPs.


conf/application.intg.conf
```diff
@@ -34,17 +34,50 @@ play.filters.hosts {
...
+        "10.0.0.54:9000",
+        "10.0.0.55:9000",
+        "10.0.0.56:9000",
+        "10.0.0.57:9000",
+        "10.0.0.58:9000",
+        "10.0.0.59:9000",
+        "10.0.0.60:9000",
+        "10.0.0.61:9000",
+        "10.0.0.62:9000"
+
```
TomJKing, line 80: Missing line break

TomJKing (commented):
Looks good. One very minor comment

TomJKing (commented):
Looks. Good thanks.

MancunianSam:
Moving subnets is very difficult so I'm going to close this.

PR #2060, opened 2022-08-22
TDR-2241 Implement Card Component

thanhz:
Most the css was taken from the original nhs frontend card, slightly modified so it doens't require JavaScript
Screenshot from 2022-08-23 11-34-00


techncl:
Hey Thanh, (I haven't managed to review this properly but) I think instead of JS, you can just hyperlink the card i.e. just put the card's html in the `a` [tag](https://www.bennadel.com/blog/3693-anchor-tags-can-contain-block-level-elements-as-of-html5.htm)

thanhz:
> Hey Thanh, (I haven't managed to review this properly but) I think instead of JS, you can just hyperlink the card i.e. just put the card's html in the `a` [tag](https://www.bennadel.com/blog/3693-anchor-tags-can-contain-block-level-elements-as-of-html5.htm)

Ah that's a good shout. For some reason I thought we couldn't do that :/
Just need to make some changes with the CSS. Thanks :+1:

techncl (commented):
Looks good, just a few things:

1. I noticed that the links are a different type blue. GDS has a section on [their site](https://design-system.service.gov.uk/styles/colour/) that specifies the colours we should use. Could you add the same colour for all TDR page links to the top of the main.scss file? This should work:

```
a {
  color:  $govuk-link-colour;
}
```
Maybe just a quick check on all the pages to see if it's had a negative effect on them (it shouldn't).

2. The text looks a bit cramped, I noticed it's because the line height and font differ from the design, maybe you could add these:
```
line-height: 1.5;
font-size: 1.1875rem;
```
We should probably use rem instead of px for better scaling.

3. the underlining of the titles doesn't disappear when you hover over the card, the closest thing you can do with CSS is hide the underlining when you hover over the title (not card) 
```
.tdr-card__heading:hover {
  text-decoration: none;
}
```
This would give a close enough effect.

thanhz:
1. I have added $govuk-link-colour to the main.scss file. I had to change the colour of the username to white as it was now turning blue whilst keeping it as a <a> tag to keep the styling. 
2. Added the line height and font size as suggested
3. Added the css hover psuedo class as suggested for the heading

techncl (commented):
Looks good. Just one more thing, I noticed that the heading is a little small; it's `1.25 rem` and in the design it is `1.5 rem`. Could you change it?

techncl (approved):
Looks good. Thanks. 👍

PR #2064, opened 2022-08-23
Refactor the closure metadata form code and call graphql query to get…

vimleshtna:
… file metadata for add-closure-metadata and overview pages


app/views/partials/inputDropdown2.scala.html
```diff
@@ -0,0 +1,46 @@
+@* dropdownOptions as Tuple until data model clearer *@
```
MancunianSam, line 1: Why do we need another inputDropdown form. Can you not change the original and update the other pages that are using it. Or if not can you give it a name which describes what it's used for.

app/controllers/AdditionalMetadataSummaryController.scala
```diff
@@ -23,9 +25,20 @@ class AdditionalMetadataSummaryController @Inject ()(val consignmentService: Con
...
       response <- consignment.files match {
         case first :: _ =>
-          Future(Ok(views.html.standard.additionalMetadataSummary(fileName, consignment.consignmentReference, first.metadata, request.token.name)))
+          Future(Ok(views.html.standard.additionalMetadataSummary(fileName, consignment.consignmentReference, getMetadata(first.metadata), request.token.name)))
         case Nil => Future.failed(new IllegalStateException(s"Can't find selected files for the consignment $consignmentId"))
       }
     } yield response
   }
+
+  private def getMetadata(metaData: GetConsignment.Files.Metadata): Map[String, String] = {
```
MancunianSam, line 34: This is a suggestion so feel free to ignore me but would it be better to have another case class here with `String` for each of these attributes rather than the Map. Then the additionalMetadataSummary view can use it as well. What do you think?

MancunianSam (commented):
Looks good thanks. Just a couple of questions.

app/controllers/util/DynamicFormUtils.scala
```diff
@@ -15,247 +14,44 @@ class DynamicFormUtils(request: Request[AnyContent], defaultFieldValues: List[(F
...
 
-  lazy val monthsWithLessThan31Days = Map(
-    2 -> "February",
-    4 -> "April",
-    6 -> "June",
-    9 -> "September",
-    11 -> "November"
-  )
-
   def formAnswersContainAnError(validatedFormAnswers: Map[String, (Option[Any], List[String])]): Boolean =
```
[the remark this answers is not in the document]
vimleshtna: Removed thanks.

app/views/partials/inputDropdown2.scala.html
vimleshtna: Actually, I thought to remove the old one later, but now, I have updated the old pages to use the new dropdown and kept the same name, thanks

app/controllers/AdditionalMetadataSummaryController.scala
vimleshtna: Yeah, good suggestion. I have updated this.

MancunianSam (approved):
Looks good, thanks

PR #2068, opened 2022-08-24
Add Scalastyle checks for tests

techncl:


test/controllers/AdditionalMetadataNavigationControllerSpec.scala
```diff
@@ -131,12 +131,13 @@ class AdditionalMetadataNavigationControllerSpec extends FrontEndTestHelper {
       checkCommonFileNavigationElements(fileSelectionPageAsString, parentFolder, folderId, fileId, selectedFolderId)
       fileSelectionPageAsString.contains(
         s"""
-           |                <button name="pageSelected" data-prevent-double-click="true" class="govuk-button" type="submit" data-module="govuk-button" role="button" value="${currentPage + 1}">
+           |                <button name="pageSelected" data-prevent-double-click="true" class="govuk-button" type="submit """.stripMargin +
```
MancunianSam, line 134: So you need the string concatenation here? Can you add another new line as it's a multi line string?

test/controllers/AdditionalMetadataNavigationControllerSpec.scala
```diff
@@ -163,12 +164,13 @@ class AdditionalMetadataNavigationControllerSpec extends FrontEndTestHelper {
       checkCommonFileNavigationElements(fileSelectionPageAsString, parentFolder, folderId, fileId, selectedFolderId)
       fileSelectionPageAsString.contains(
         s"""
-           |                                <button name="pageSelected" data-prevent-double-click="true" class="govuk-button__tna-button-link" type="submit" data-module="govuk-button" role="link" value="${currentPage - 1}">
+           |                                <button name="pageSelected" data-prevent-double-click="true" """.stripMargin +
```
MancunianSam, line 167: The same thing here. Can you have these two strings as part of the same multiline string? There's a couple more of these in the file if you can check them.

test/controllers/UploadControllerSpec.scala
```diff
@@ -707,9 +707,12 @@ class UploadControllerSpec extends FrontEndTestHelper {
          |            </a>""".stripMargin
     )
     pageAsString must include(
-      """|                    <p class="upload-progress-error-timeout__message" hidden>Your upload has timed out. Click 'Return to start' to begin a new transfer.</p>
-         |                    <p class="upload-progress-error-authentication__message" hidden>You have been signed out. Click 'Return to start' to begin a new transfer.</p>
-         |                    <p class="upload-progress-error-general__message" hidden>Click 'Return to start' to begin a new transfer.</p>""".stripMargin
+      """                    <p class="upload-progress-error-timeout__message" hidden>""" +
```
MancunianSam, line 710: These lines will be short enough if you remove the whitespace between | and < so can you do that? I'm not sure if that will still work but it does look easier to read when they're not split out over separate lines. Don't worry if the whitespace is needed.

MancunianSam (commented):
Looks good thanks. There's a few places where you're concatenating two multiline strings. It might be better to have one string with extra line breaks. I've not put a comment on each one but have a look.

Also, I think some of the lines will be short enough if you moved some of the whitespace. I'm just thinking that with all the line breaks, it might be harder to read the html strings. 

The last option to think about is to increase the allowed line length in the scalastyle config if it's causing problems but that's up to you :smile:

test/controllers/UploadControllerSpec.scala
techncl: Yes, I tried this but it doesn't work; the issue is the new line character directly after the `hidden>`. Maybe I could add a `.replace("hidden>\n",  "hidden>")` at the end of the string?

test/controllers/FileChecksControllerSpec.scala
```diff
@@ -224,8 +226,13 @@ class FileChecksControllerSpec extends FrontEndTestHelper with TableDrivenProper
     "call the fileCheckProgress endpoint" in {
       val graphQLConfiguration: GraphQLConfiguration = new GraphQLConfiguration(app.configuration)
       val consignmentService: ConsignmentService = new ConsignmentService(graphQLConfiguration)
-      val controller = new FileChecksController(getAuthorisedSecurityComponents, new GraphQLConfiguration(app.configuration),getValidStandardUserKeycloakConfiguration,
-        consignmentService,frontEndInfoConfiguration)
+      val controller = new FileChecksController(
+        getAuthorisedSecurityComponents,
+        new GraphQLConfiguration(app.configuration),
```
thanhz, line 231: Instead of creating a new '`GraphQLConfiguration`' you can use the one declared on line 227

thanhz (commented):
Looks good, just some minor comments

test/controllers/UploadControllerSpec.scala
MancunianSam: No, don't worry about it if it makes it more complicated. I was just wondering.

test/controllers/FileChecksControllerSpec.scala
```diff
@@ -242,8 +249,13 @@ class FileChecksControllerSpec extends FrontEndTestHelper with TableDrivenProper
     "throw an error if the API returns an error" in {
       val graphQLConfiguration: GraphQLConfiguration = new GraphQLConfiguration(app.configuration)
       val consignmentService: ConsignmentService = new ConsignmentService(graphQLConfiguration)
-      val controller = new FileChecksController(getAuthorisedSecurityComponents, new GraphQLConfiguration(app.configuration),getValidStandardUserKeycloakConfiguration,
-        consignmentService,frontEndInfoConfiguration)
+      val controller = new FileChecksController(
+        getAuthorisedSecurityComponents,
+        new GraphQLConfiguration(app.configuration),
```
thanhz, line 254: Same thing here

MancunianSam (approved):
@thanhz I'll approve this as your comments have been resolved as well.

PR #2101, opened 2022-08-31
Redirect to homepage if cache session is missing.

MancunianSam:
Sometimes, AWS will remove items from the cache for reasons of its own. If this happens, pac4j can't find the session it's looking for and throws an exception.

This will intercept the exception, check the message (as there are other pac4j ones we do want to fail) and redirect to the homepage. 

Because this is the result of an error, the previous page isn't stored so we can't redirect to the page the user was on before they got to the sign in page.

I've marked this as draft until we've checked with Kirsten.


TomJKing (approved):
Looks good. Thanks.

Confirmed by Kirsten that should redirect to the /homepage until the "view history" page is available.

PR #2108, opened 2022-09-02
TDR-2287 Make filename not show glove icon on hover

thanhz:
This makes it so that when you hover over the filenames you will no longer see a glove icon. Folders will still display glove icon.


TomJKing (approved):
Looks good. Thanks

PR #2109, opened 2022-09-02
Tdr-2290 Display correct file count for empty folders

thanhz:
...and hide next button on the file navigation page if there are no files.


TomJKing (approved):
Looks good. Thanks

PR #2146, opened 2022-09-14
TDR-2313 - Show empty value if there is no default value for closure …

vimleshtna:
Show empty value if there is no default value for closure period


MancunianSam (approved):
Looks good, thanks.

techncl (approved):
Looks good! Thanks. Edit (should have refreshed the page before reviewing 😄)

PR #2148, opened 2022-09-15
Prevent last page number from showing on empty folder

thanhz:
The navigation page was showing '0' when navigating into an empty folder


vimleshtna (commented):
Looks good however If possible, can we add a test for this, please?
Not sure, if we have an end-to-end test to cover this scenario.

thanhz:
> Looks good however If possible, can we add a test for this, please? Not sure, if we have an end-to-end test to cover this scenario.

Thanks, I've added more cases to the empty folder test. I believe we won't be adding e2e tests until all the pages are linked

app/views/standard/additionalMetadataNavigation.scala.html
```diff
@@ -96,7 +96,7 @@ <h2 class="app-task-list__section">Folder uploaded: @folderUploaded</h2>
                                 @if(totalPages - currentPage > 3) {
                                         <li class="govuk-pagination__item govuk-pagination__item--ellipses">&ctdot;</li>
                                 }
-                            @if(totalPages != 1) {
+                            @if(totalPages != 1 && totalPages != 0) {
```
vimleshtna, line 99: Can we add one check which will work for both conditions: `@if(totalPages > 1)`?

vimleshtna (commented):
Look good, just a question, thanks

app/views/standard/additionalMetadataNavigation.scala.html
thanhz: Yes you're right, that is much simpler :+1:

vimleshtna (approved):
Looks good, thanks,

PR #2149, opened 2022-09-15
Populate description using values from db

thanhz:
The property `closed_for` was recently changed to `Closed` so this has also been included in order for the form to show up

### DIFF
--- a/app/controllers/util/CustomMetadataUtils.scala
+++ b/app/controllers/util/CustomMetadataUtils.scala
@@ -40,7 +40,7 @@ class CustomMetadataUtils(allCustomMetadataProperties: List[CustomMetadata]) {
 
 //    Use this until descriptions are added, then use property.description.getOrElse("")
     val fieldLabel = dbAndFieldLabel.getOrElse(property.name, property.fullName.getOrElse(""))
-    val fieldDescription = dbAndFieldDescription.getOrElse(property.name, property.description.getOrElse(""))
+    val fieldDescription = property.description.getOrElse("")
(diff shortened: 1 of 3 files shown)


TomJKing (approved):
Looks good. Could you hold off merging until we've sorted out the DB migrations, as I think the changes that this needs in the DB haven't gone in yet.

app/controllers/util/CustomMetadataUtils.scala
```diff
@@ -40,7 +40,7 @@ class CustomMetadataUtils(allCustomMetadataProperties: List[CustomMetadata]) {
 
 //    Use this until descriptions are added, then use property.description.getOrElse("")
     val fieldLabel = dbAndFieldLabel.getOrElse(property.name, property.fullName.getOrElse(""))
-    val fieldDescription = dbAndFieldDescription.getOrElse(property.name, property.description.getOrElse(""))
+    val fieldDescription = property.description.getOrElse("")
```
vimleshtna, line 43: Thanh, can you please remove `dbAndFieldDescription` map if we are not using this anywhere else?

https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/controllers/util/CustomMetadataUtils.scala#L20

techncl: And also the comment on line 41

thanhz: It's not being used. I have removed it

vimleshtna (approved):
Looks good, thank you.

PR #2150, opened 2022-09-15
TDR-2305 - Show an error if user enters a future date for the FOI dec…

vimleshtna:
…ision asserted field


test/controllers/util/FormFieldSpec.scala
```diff
@@ -117,140 +117,174 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
         year = InputNameAndValue("Year", "2015", "YYYY"))
     }
 
-    "validate should not return any error when the given date is valid" in {
+    "validate should not return any error when the given date is valid and future date is allowed" in {
```
techncl, line 120: I think the test should remain how it was and instead, you add a new test to the bottom of the file ("validate should not return an error when future date is allowed and the given date is in future") because this test was originally testing whether the date is entered correctly i.e. what the user entered was an official date; adding the future date part on top of it makes it test for two things.

test/controllers/util/FormFieldSpec.scala
```diff
@@ -117,140 +117,174 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
...
+        getDate(LocalDateTime.now().plusMonths(1)),
+        getDate(LocalDateTime.now().plusYears(1))
+      )
+      date.foreach {
+        case (day, month, year) =>
+          DateField.validate(day, month, year, dateField.copy(isFutureDateAllowed = false)) shouldBe Some(s"${dateField.fieldName} date cannot be a future date.")
+      }
+    }
+
+    "validate should return an error when future date is not allowed but the given date is present or past" in {
```
techncl, line 272: Should this be `validate should NOT return an error`?

test/controllers/util/FormFieldSpec.scala
```diff
@@ -117,140 +117,174 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
...
+      date.foreach {
+        case (day, month, year) =>
+          DateField.validate(day, month, year, dateField.copy(isFutureDateAllowed = false)) shouldBe None
+      }
+    }
+
+    def getDate(dateTime: LocalDateTime): (String, String, String) =
+      (dateTime.getDayOfMonth.toString, dateTime.getMonthValue.toString, dateTime.getYear.toString)
   }
 }
```
techncl, line 290: You've covered 3 scenarios:

1. validate should **not** return an error when future date **is allowed** and the given date is in future
2. validate should return an error when future date **is not allowed** but the given date is in future
3. validate should **not** return an error when future date **is not allowed** and the given date is present or past

I think maybe one more, just as a sanity check, would be good:

1. validate should **not** return an error when future date **is allowed** and the given date is present or past

test/controllers/util/CustomMetadataUtilsSpec.scala
```diff
@@ -136,13 +138,26 @@ class CustomMetadataUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeA
...
             verifyText(field.asInstanceOf[DropdownField], property)
+          case unknownType => throw new IllegalArgumentException(s"Invalid type $unknownType")
         }
         field.fieldDescription should equal(property.description.getOrElse(""))
         field.fieldName should equal(property.fullName.get)
         field.isRequired should equal(property.propertyGroup.contains("MandatoryMetadata"))
     }
   }
 
+  "convertPropertiesToFields" should "convert date property to field and it shouldn't allow future date when property name is foiExemptionAsserted" in {
```
techncl, line 149: Similar comment to the one I made at the very top about hardcoding foiExemptionAsserted

techncl (commented):
Looks good! Just a few comments.

app/controllers/util/CustomMetadataUtils.scala
```diff
@@ -52,7 +51,7 @@ class CustomMetadataUtils(allCustomMetadataProperties: List[CustomMetadata]) {
       case DateTime =>
         DateField(property.name, fieldLabel, fieldDescription,
           InputNameAndValue("Day", "", "DD"), InputNameAndValue("Month", "", "MM"), InputNameAndValue("Year", "", "YYYY"),
-          isRequired
+          isRequired, isFutureDateAllowed = property.name != foiExemptionAsserted
```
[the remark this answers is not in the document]
vimleshtna: Good point but I don't think we should create a new data type. We may add a flag `futureDateAllowed` in `FileProperty` table.
We can have a discussion with teams and then we can decide if we can go with this approach.

test/controllers/AddClosureMetadataControllerSpec.scala
```diff
@@ -311,6 +312,56 @@ class AddClosureMetadataControllerSpec extends FrontEndTestHelper {
...
+          .withFormUrlEncodedBody(formSubmission: _*)
+          .withCSRFToken
+        )
+      val addClosureMetadataPageAsString = contentAsString(addClosureMetadataPage)
+
+      checkPageForStaticElements.checkContentOfPagesThatUseMainScala(addClosureMetadataPageAsString, userType = "standard")
+      checkForExpectedClosureMetadataFormPageContent(addClosureMetadataPageAsString)
+      formTester.checkHtmlForOptionAndItsAttributes(addClosureMetadataPageAsString, formSubmission.toMap, formStatus = "PartiallySubmitted")
+
+      addClosureMetadataPageAsString must include(
```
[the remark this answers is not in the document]
vimleshtna: The test is specific to test the error related to future date on the page. So, we cannot add in that method otherwise all other tests will fail.

test/controllers/util/DynamicFormUtilsSpec.scala
```diff
@@ -171,28 +172,54 @@ class DynamicFormUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeAndA
...
-      val (_, dynamicFormUtils): (ListMap[String, List[String]], DynamicFormUtils) = generateFormAndSendRequest(
-        MockFormValues(month = List("b4"),
-          year = List("2r"),
-          numericTextBoxValue = List("1.5"),
-          day2 = List("000"),
-          month2 = List("e"),
-          year2 = List("-")
-        )
+  "validateAndConvertSubmittedValuesToFormFields" should "return an error when future date is not allowed" in {
+    val dateTime = LocalDateTime.now().plusDays(1)
```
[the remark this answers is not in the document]
vimleshtna: Actually, we are already testing all these scenarios in the `FormFieldSpec.scala` class. So, I just kept only one test to avoid writing too many duplicate tests. Hope that makes sense.

test/controllers/util/CustomMetadataUtilsSpec.scala
vimleshtna: Yeah, we will change this one if we go with a data-driven approach.

test/controllers/util/FormFieldSpec.scala
vimleshtna: Yeah, I'll add those checks in the new test, thanks.

test/controllers/util/FormFieldSpec.scala
vimleshtna: Yeah, good catch. I changed that, thanks.

test/controllers/util/FormFieldSpec.scala
vimleshtna: Yeah, that makes sense. I added another test, thanks

test/controllers/AddClosureMetadataControllerSpec.scala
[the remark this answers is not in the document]
techncl: Ah yes, sorry, you're correct. But then, do we need this whole unit test since we are already testing that the correct error message is being returned in the FormFieldSpec? These controller tests are just mainly to test if the controller did the right thing i.e. rendered a page that is displaying/hiding the correct elements (when you GET, POST incorrectly, POST correctly) rather than the specific error messages as these come from FormFields. As long as FormFields tests pass, it should always display the correct text since the views are all calling `fieldErrors`. This is why the error messages I put in `MockInputOption` are only the default ones because these are hidden when you GET but are unhidden when you incorrectly POST.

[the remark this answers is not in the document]
vimleshtna: Yeah, you are right. It's not required actually. I'll delete it.

techncl (approved):
Looks good! Thanks.

PR #2164, opened 2022-09-21
Sort drop down by ordinal value.

MancunianSam:
We're retrieving the ordinal for file property values from the API and using it to sort the values.

This is set as draft until https://github.com/nationalarchives/tdr-generated-graphql/pull/661 is released.


conf/application.conf
```diff
@@ -6,6 +6,6 @@ auth.domain="auth.tdr-integration.nationalarchives.gov.uk"
 
 upload.url="https://upload.tdr-integration.nationalarchives.gov.uk"
 
-consignmentapi.domain="https://api.tdr-integration.nationalarchives.gov.uk"
+consignmentapi.domain="http://localhost:8080"
```
TomJKing, line 9: Should this be change to the `localhost`. This configuration is used for local development, and I think generally devs will want to point to intg api.

TomJKing (commented):

TomJKing (approved):
Looks good. Thanks

PR #2174, opened 2022-09-23
Add validation to Radio buttons for completeness

techncl:
Previously, radio buttons didn't have validation because on the closure metadata form, an option is always selected by default, but it just makes sense to have it for the future and also to check what would happen if a user sent us something unexpected via the form.


MancunianSam (approved):
:+1:

PR #2199, opened 2022-10-03
Add multiValue to FormField

techncl:
We are going to need 'multiValue' for when we start saving 'multiValue' properties like FOI exemption codes


app/views/standard/addClosureMetadata.scala.html
```diff
@@ -53,7 +53,7 @@ <h2>@file.name</h2>
                   ),
                   fieldErrors
                 )
-                case DateField(fieldId, fieldName, fieldDescription, day, month, year, _, fieldErrors, _) => inputDate(
+                case DateField(fieldId, fieldName, fieldDescription, multiValue, day, month, year, _, fieldErrors, _) => inputDate(
```
MancunianSam, line 56: This parameter isn't being used, can you swap it out with `_` ? It's the same on line 40 and line 66 as well.

MancunianSam (commented):
Looks good thanks, just a small question.

app/views/standard/addClosureMetadata.scala.html
techncl: Ah, yes, you are correct.

MancunianSam (approved):

PR #2225, opened 2022-10-07
Revert changes to the API url.

MancunianSam:
This should be set to the integration url and not localhost


techncl (approved):
Looks good!

PR #2237, opened 2022-10-14
Return error if files fail to upload

techncl:


npm/test/s3upload.test.ts
```diff
@@ -160,18 +164,62 @@ test("a single file upload returns the correct key and should call addFileStatus
...
+      fileId: tdrFileWithPath.fileId,
+      statusType: "Upload",
+      statusValue: "Failed"
+    })
+  )
+  s3Mock.on(PutObjectCommand).resolves({ $metadata: { httpStatusCode: 500 } })
+  const s3Upload = new S3Upload(s3Mock as unknown as S3Client, "")
+
+  const result:
+    | {
```
MancunianSam, line 199: This isn't directly related to your changes but this is the third place now that we're writing the type out like this. Do you think it would be clearer to create an interface in `s3Upload/index.ts` like
```
interface IUploadResult {
  sendData: ServiceOutputTypes[]
  processedChunks: number
  totalChunks: number
}
```
Then replace this type and the one in `s3Upload/index.ts` and the one in `clientfileprocessing.test.ts` It doesn't really change anything but it might look a bit neater. What do you think?

MancunianSam (commented):
Looks good thanks. Just once question about tidying up the types.

MancunianSam (approved):

PR #2261, opened 2022-10-19
Make ESLint search within subfolders & fix ESLint errors

techncl:
When we were running 'npm run lint', it wasn't searching the subfolder of 'upload', so this should fix that

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -14,7 +14,7 @@
     "build-css": "npm-run-all add-stylesheet-dir sass-compile compress-css",
     "build:local": "npm-run-all copy-assets build-css build-ts:prod",
     "build": "npm-run-all copy-assets build-css build-ts:prod",
-    "lint": "eslint src/**/*.ts src/**/**/*.ts",
+    "lint": "eslint src --ext .ts",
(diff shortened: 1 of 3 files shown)


npm/package.json
```diff
@@ -14,7 +14,7 @@
     "build-css": "npm-run-all add-stylesheet-dir sass-compile compress-css",
     "build:local": "npm-run-all copy-assets build-css build-ts:prod",
     "build": "npm-run-all copy-assets build-css build-ts:prod",
-    "lint": "eslint src/**/*.ts",
+    "lint": "eslint src/**/*.ts src/**/**/*.ts",
```
vimleshtna, line 17: I think if you use just a single quote then it works recursively.

`"lint": "eslint 'src/**/*.ts'"`

vimleshtna (commented):
Looks good, just a suggestion, it should work.

npm/package.json
techncl: Oh, cool. Thanks.

vimleshtna (approved):
Looks good, thanks.

PR #2264, opened 2022-10-20
TDR-2168 - Implement "Advisory Council Sign Off" page

vimleshtna:


app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  def submitClosureStatus(consignmentId: UUID, fileIds: List[UUID]): Action[AnyContent] = standardTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
+
+    val errorFunction: Form[ClosureStatusFormData] => Future[Result] = { formWithErrors: Form[ClosureStatusFormData] =>
+      for {
+        (consignmentRef, filePaths, parentFolderId) <- cache.getOrElseUpdate[(String, List[String], UUID)](s"$consignmentId-data", 1.hour)(
+          for {
+            details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
+            consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Option(FileFilters(None, Option(fileIds), None)))
+            filePaths = consignment.files.flatMap(_.fileMetadata).filter(_.name == clientSideOriginalFilepath).map(_.value)
```
TomJKing, line 66: Why are the "file paths" needed?

app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    val filters = Option(FileFilters(None, Option(fileIds), None))
+
+    for {
+      details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
+      consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, filters)
+      response <- if (consignment.files.nonEmpty) {
+        val filePaths = consignment.files.flatMap(_.fileMetadata).filter(_.name == clientSideOriginalFilepath).map(_.value)
+        val closureStatus = consignment.files.flatMap(_.fileMetadata).find(_.name == closureType.name).exists(_.value == closureType.value)
+        cache.set(s"$consignmentId-data", (consignment.consignmentReference, filePaths, details.parentFolderId.get), 1.hour)
```
TomJKing, line 48: Is there a reason for caching this information?

test/controllers/AdditionalMetadataClosureStatusControllerSpec.scala
```diff
@@ -0,0 +1,309 @@
...
+
+      status(response) mustBe OK
+      contentType(response) mustBe Some("text/html")
+
+      verifyClosureStatusPage(contentAsString(response), consignmentId)
+
+      wiremockServer.verify(postRequestedFor(urlEqualTo("/graphql")))
+    }
+
+    "render the closure status page and closure status checkbox should be checked" in {
```
TomJKing, line 64: Maybe reword this test a bit to make it clear what the scenario is for the checkbox being checked: `render the closure status page with the closure status checkbox checked when closure status already set to 'closed'`

test/controllers/AdditionalMetadataClosureStatusControllerSpec.scala
```diff
@@ -0,0 +1,309 @@
...
+
+      status(response) mustBe BAD_REQUEST
+      contentType(response) mustBe Some("text/html")
+
+      verifyClosureStatusPage(contentAsString(response), consignmentId, isChecked = false, true)
+
+      wiremockServer.verify(postRequestedFor(urlEqualTo("/graphql")))
+    }
+
+    "should save the metadata and redirect to the next page" in {
```
TomJKing, line 189: Maybe change the test title to make clear when this would happen: `should save the metadata and redirect to the next page when the closure status checkbox is checked`

TomJKing (commented):
Looks good. Just a couple of minor questions and comments.

app/controllers/AdditionalMetadataClosureStatusController.scala
vimleshtna: I added this because if the user submits the page without confirming the checkbox then it will again call APIs to get that information to show on the page. So, if we cache then it will pick from the cache instead of calling the APIs.
Let me know your opinion on this please.

app/controllers/AdditionalMetadataClosureStatusController.scala
vimleshtna: It's required because we are showing the name of the file as per UI design https://tdr-prototype.herokuapp.com/metadata/closure-metadata/T2-flour/check-closure-status

app/controllers/AdditionalMetadataClosureStatusController.scala
TomJKing: OK. That sounds reasonable

TomJKing (approved):
Looks good. Thanks

app/controllers/AdditionalMetadataClosureStatusController.scala
TomJKing, line 48: Should the cache be checked first to see if it exists? If so use the data from the cache, if not, only then make the calls to the API and set it?

TomJKing (commented):
Added another comment. Just noticed something.

app/controllers/AdditionalMetadataClosureStatusController.scala
vimleshtna: Yeah, I thought the same but then I realized if we cache it then it won't show the updated page if the user comes back. It should show the checkbox as checked as the user is already confirmed.
If we really want to go with cache then we have to update the cache when the user confirms it but I don't think it's a good idea. What do you think?

TomJKing: Yeah, good point. Lets go without the cache. We can always revisit if needed.

TomJKing (approved):
Looks good. Thanks

PR #2287, opened 2022-10-27
Allow 'Description Public' to be displayed on the form

techncl:
Previously 'Description Public' was filtered out and not displayed on the form; this commit removes this filtering. Field name is wrong and label is missing as we need to change this in the DB


vimleshtna (approved):
Looks good, thanks.

PR #2311, opened 2022-11-03
Add nested navigation

MancunianSam:
This uses the https://github.com/nationalarchives/tdr-components/ library which provides the javascript and css.

Most of the rest of the changes are removing the old code for the cache based solution.


app/controllers/AdditionalMetadataNavigationController.scala
```diff
@@ -1,252 +1,35 @@
...
-      ids <- folderDescendantsCache.getDescendantFileIds(consignmentId, folderId, token)
-      _ = if (selectedFiles.containsSome(ids)) partSelectedCache.add(folderId)
-      descendantFilesSelected = selectedFiles.toSet.intersect(ids.toSet).size
-    } yield descendantFilesSelected
+  def submitFiles(consignmentId: UUID, metadataType: String): Action[AnyContent] = standardTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
+    val fileIds = request.body.asFormUrlEncoded.getOrElse(Map()).keys.toList
+      .filter(_ != "csrfToken")
+      .map(UUID.fromString)
+    if(metadataType == "closure") {
+      Future(Redirect(routes.AddClosureMetadataController.addClosureMetadata(consignmentId, fileIds)))
```
TomJKing, line 29: This should be redirecting to the closure status page: `routes.AdditionalMetadataClosureStatusController.getClosureStatusPage`

TomJKing (commented):

app/services/ConsignmentService.scala
```diff
@@ -134,6 +137,34 @@ class ConsignmentService @Inject()(val graphqlConfiguration: GraphQLConfiguratio
       .map(data => data.getConsignment.get)
   }
 
+  def getAllConsignmentFiles(consignmentId: UUID, token: BearerAccessToken): Future[File] = {
```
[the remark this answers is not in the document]
MancunianSam: Good spot, I've removed it, the associated client and the `getAllDescendants` can be removed now as well as it was only called in `getConsignmentPaginatedFile`

app/controllers/AdditionalMetadataNavigationController.scala
MancunianSam: I've changed it and updated the tests.

TomJKing (approved):
Looks good. Thanks.

PR #2312, opened 2022-11-03
TDR-2520 Dynamically display fileNavigation page depending on metadataType

thanhz:
The “Add Descriptive Metadata” button should take the user to the summary page but with the removal of the old file navigation. We should probably link it in this pr #2311 or this one depending on which gets merged first as there will be conflicts


TomJKing (approved):
Looks good. Thanks

PR #2320, opened 2022-11-07
Increase max length of posted content.

MancunianSam:
For 10000 files, the size of the posted json is around 1.7 Mb so 5Mb
should give us a bit of space.


TomJKing (approved):
Looks good. Thanks

PR #2333, opened 2022-11-09
Tdr 2412 update html and controller for alternative title and description

techncl:


test/testUtils/FormTester.scala
```diff
@@ -1,15 +1,16 @@
...
+  def generateWaysToIncorrectlySubmitAForm(value: String = "true", combineOptionNameWithValue: Boolean = false): Seq[Seq[(String, String)]] = {
     val possibleOptions: Seq[String] = defaultOptions.map(_.name)
     val optionsToSelectToGenerateFormErrors =
       for {
         numberRangeOfOptionsToSelect <- (1 until possibleOptions.length).toList
         optionsToSelect <- possibleOptions.combinations(numberRangeOfOptionsToSelect)
-      } yield optionsToSelect.map(option => (option, "true"))
+      } yield optionsToSelect.map(option => (option, if (combineOptionNameWithValue) s"$option $value" else value))
 
     optionsToSelectToGenerateFormErrors
```
MancunianSam, line 15: I know you didn't change this but you don't need the variable here, you can just have the for comprehension as the last line.

test/controllers/AddClosureMetadataControllerSpec.scala
```diff
@@ -547,11 +766,30 @@ class AddClosureMetadataControllerSpec extends FrontEndTestHelper {
     )
     addClosureMetadataPageAsString must include(
       """        <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
-        |            Description Public
+        |            Is the description closed?
         |        </legend>""".stripMargin
     )
   }
 
+  private def checkForExpectedClosureMetadataDependenciesFormPageContent(addClosureMetadataPageAsString: String, fullNameAndName: Map[String, String]): Unit = {
```
MancunianSam, line 774: I'm not sure if this is an issue but these checks are very dependant on white space. I had to refactor a html page recently and I spent a lot of time trying to get the whitespace right in the test. You could try something like `addClosureMetadataPageAsString.replaceAll("\\s{2,}", "")` which will remove all spaces apart from single ones which might help. 
I'm not sure that it will though so if this makes things more complicated then ignore me.

test/controllers/util/CustomMetadataUtilsSpec.scala
```diff
@@ -106,8 +106,16 @@ class CustomMetadataUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeA
           field.isInstanceOf[RadioButtonGroupField] should be(true)
           verifyBoolean(field.asInstanceOf[RadioButtonGroupField], property.defaultValue)
         case Text =>
-          field.isInstanceOf[DropdownField] should be(true)
-          verifyText(field.asInstanceOf[DropdownField], property)
+          property.propertyType match {
```
MancunianSam, line 109: This is a duplicate of the block on line 146. Is it worth splitting this into a `checkText` method?

MancunianSam (commented):
Looks good thanks, just a couple of questions/suggestions.

app/controllers/AddClosureMetadataController.scala
```diff
@@ -87,36 +83,156 @@ class AddClosureMetadataController @Inject() (
...
+  private def getValuesThatWereSelectedIfTheyHaveDependencies(
+      dependencyProperties: Set[CustomMetadata],
+      metadataInput: List[UpdateFileMetadataInput]
+  ): Set[ValueSelectedAndDepsToDel] = {
+    val propertiesThatHaveValuesWithDependencies: Set[CustomMetadata] = getPropertiesWhereValuesHaveDependencies(dependencyProperties)
+    val valuesThatWereSelectedThatHaveDependencies: Set[ValueSelectedAndDepsToDel] = getValueSelectedIfItHasDependencies(metadataInput, propertiesThatHaveValuesWithDependencies)
+    valuesThatWereSelectedThatHaveDependencies
+  }
+
+  private def getDefaultFieldsForForm(staticMetadata: Set[StaticMetadata], consignmentId: UUID, request: Request[AnyContent]): Future[(Set[CustomMetadata], List[FormField])] = {
```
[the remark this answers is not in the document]
techncl: Yes, that sounds like a good idea.

test/testUtils/FormTester.scala
techncl: Ah, yes. I think I wrote this originally. The reason why was just to give the reader a little understanding as to what it was returning but I will change the method name to something like `generateOptionsToSelectToGenerateFormErrors` so that it's no longer necessary. 👍

MancunianSam (approved):
Thanks for doing this :+1:

PR #2342, opened 2022-11-10
Remove 'Completed' from Upload status update

techncl:


vimleshtna (commented):
Do we still need to call this API `"update-consignment-status"` from the frontend if we are not setting any value?
If yes, then we should set it as InProgress. What do you think?

techncl:
> Do we still need to call this API `"update-consignment-status"` from the frontend if we are not setting any value? If yes, then we should set it as InProgress. What do you think?

@vimleshtna, yes we do. When a user clicks the continue button on the upload page, we call [startUpload](https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/npm/src/clientfilemetadataupload/index.ts), in the frontend which calls the API, which eventually calls the [StartUpload method](https://github.com/nationalarchives/tdr-consignment-api/blob/cc35705452239b7525f3e2243caade4c5e1141ed/src/main/scala/uk/gov/nationalarchives/tdr/api/service/ConsignmentService.scala#L45), which sets it to "InProgress". Then after the S3 file uploads, the code (that I edited) used to call the API and set the Upload consignment status to "Completed" but now (after my change) it will call the API, which will check the file UploadStatuses and if they are all completed, marks the consignment UploadStatus as complete. So the API now sets it to "Completed" rather than the frontend; the frontend just triggers this process. I think the plan later is to have a lambda do the triggering (there is a spike card), but for now, we're letting the frontend do it.

npm/src/updateconsignmentstatus/index.ts
```diff
@@ -12,8 +12,7 @@ export class UpdateConsignmentStatus {
   ): Promise<number | void | Error> {
     const updateConsignmentStatusInput: ConsignmentStatusInput = {
       consignmentId: uploadFilesInfo.consignmentId,
-      statusType: "Upload",
-      statusValue: "Completed"
+      statusType: "Upload"
```
vimleshtna, line 15: Can we rename the main method `markUploadStatusAsCompleted` to something more meaningful, maybe `setUploadStatusBasedOnFileStatuses`?

vimleshtna (commented):

vimleshtna (approved):
Looks good, thanks.

PR #2343, opened 2022-11-10
Update download metadata report.

MancunianSam:
Originally, this had the filename hard coded in at the beginning of the
report using the original file path. The API has been updated now to
store the file name and the file type in the metadata table and the
relevant file properties now have an ordering on them so this report can
be entirely data driven. I've updated the tests as well.


vimleshtna (approved):
Looks good, thanks

techncl (commented):
Looks good. I downloaded the CSV, but the filename column is blank. Is that fine?

MancunianSam:
@techncl  It needs the changes from https://github.com/nationalarchives/tdr-consignment-api/pull/481 to populate the file name in the FileMetadata table. That' merged now so once it's deployed, the file name should be populated.

PR #2357, opened 2022-11-15
Skip closure type confirmation if already confirmed.

MancunianSam:
This does a few things.

If all files are already confirmed as closed, the closure status page is
skipped.

If you go to the closure status page anyway, the checkbox is checked and
disabled and the submit button is changed to a link which takes you to
the custom metadata page.

I've put the logic for whether all files are closed or not in a separate
method so we can change the logic easily.

### DIFF
--- a/app/services/ConsignmentService.scala
+++ b/app/services/ConsignmentService.scala
@@ -3,11 +3,13 @@ package services
 import com.google.common.collect.Ordering.natural
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken
 import configuration.GraphQLConfiguration
+import controllers.util.MetadataProperty.closureType
 import graphql.codegen.AddConsignment.addConsignment
 import graphql.codegen.GetConsignment.getConsignment
 import graphql.codegen.GetConsignmentExport.getConsignmentForExport
 import graphql.codegen.GetConsignmentFiles.getConsignmentFiles
 import graphql.codegen.GetConsignmentFiles.getConsignmentFiles.GetConsignment.Files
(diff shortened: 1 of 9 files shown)


test/services/ConsignmentServiceSpec.scala
```diff
@@ -493,4 +503,34 @@ class ConsignmentServiceSpec extends AnyWordSpec with MockitoSugar with BeforeAn
       results.failed.futureValue shouldBe a[HttpError]
     }
   }
+
+  "areAllFilesClosed" should {
+    "return true if all file files have a closure type of Closed" in {
```
techncl, line 508: Is "file files" a mistake?

app/services/ConsignmentService.scala
```diff
@@ -66,6 +68,14 @@ class ConsignmentService @Inject() (val graphqlConfiguration: GraphQLConfigurati
       )
   }
 
+  def areAllFilesClosed(consignment: GetConsignment): Boolean = {
+    consignment.files
+      .filter(file => file.fileMetadata.exists(metadata => metadata.name == "FileType" && metadata.value == "File"))
+      .filter(file => file.fileMetadata.exists(metadata => metadata.name == closureType.name && metadata.value != closureType.value))
+      .map(_.fileId)
```
techncl, line 75: Do you need to `.map` over the files?

techncl (approved):
Looks good. Just 2 comments; feel free to ignore.

app/services/ConsignmentService.scala
MancunianSam: No I don't good point. I've removed the map and replaced the second filter with exists, thanks.

techncl (approved):

PR #2370, opened 2022-11-17
Use existing metadata report

MancunianSam:
The transfer complete report was using the old metadata fields in the
consignment export query.

We already have an endpoint for creating a metadata csv file so I've
linked to that one on the transfer complete page and removed the old
one.

They are slightly different in that the metadata report has more columns
but all of the information that was in the transfer complete report is
still there and it now has headings and columns driven from the
database.


vimleshtna (approved):
Looks good, thanks

PR #2387, opened 2022-11-23
TDR-2462 - Show an error message on the navigation page

vimleshtna:
Show an error message if a user clicks on the 'Add or edit metadata' button without selecting any files

### DIFF
--- a/app/controllers/AdditionalMetadataNavigationController.scala
+++ b/app/controllers/AdditionalMetadataNavigationController.scala
@@ -4,23 +4,25 @@ import auth.TokenSecurity
 import configuration.KeycloakConfiguration
 import graphql.codegen.types.FileFilters
 import org.pac4j.play.scala.SecurityComponents
+import play.api.cache.AsyncCacheApi
 import play.api.mvc.{Action, AnyContent, Request}
 import services.ConsignmentService
+import viewsapi.Caching.preventCaching
 
 import java.util.UUID
 import javax.inject.Inject
 import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
 
 class AdditionalMetadataNavigationController @Inject() (
     val consignmentService: ConsignmentService,
     val keycloakConfiguration: KeycloakConfiguration,
-    val controllerComponents: SecurityComponents
+    val controllerComponents: SecurityComponents,
+    val cache: AsyncCacheApi
 ) extends TokenSecurity {
 
   def getAllFiles(consignmentId: UUID, metadataType: String): Action[AnyContent] = standardTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
-    consignmentService
-      .getAllConsignmentFiles(consignmentId, request.token.bearerAccessToken)
-      .map(allFiles => {
+    getCachedFiles(consignmentId, metadataType, request).map(allFiles => {
         Ok(views.html.standard.additionalMetadataNavigation(consignmentId, request.token.name, allFiles, metadataType))
       })
   }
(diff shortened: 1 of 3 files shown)


test/controllers/AdditionalMetadataNavigationControllerSpec.scala
```diff
@@ -102,42 +103,84 @@ class AdditionalMetadataNavigationControllerSpec extends FrontEndTestHelper {
...
+            FakeRequest(POST, s"/consignment/$consignmentId/additional-metadata/files/descriptive/")
+              .withFormUrlEncodedBody(Seq((fileId, "checked")): _*)
+              .withCSRFToken
+          )
         playStatus(result) must equal(SEE_OTHER)
-        redirectLocation(result).get must equal(s"/consignment/$consignmentId/additional-metadata/closure/selected-summary")
+        redirectLocation(result).get must equal(s"/consignment/$consignmentId/additional-metadata/closure/selected-summary?fileIds=$fileId")
+      }
+
+      "redirect to the file navigation page with an error message if a user submits the page without selecting any files" in {
```
TomJKing, line 153: Maybe re-word the test slightly: `... without selecting any files or folders`

app/controllers/AdditionalMetadataNavigationController.scala
```diff
@@ -32,20 +33,34 @@ class AdditionalMetadataNavigationController @Inject() (
...
-        .map(consignment => {
-          val areAllClosed = consignmentService.areAllFilesClosed(consignment)
-          if (areAllClosed) {
-            Redirect(routes.AddClosureMetadataController.addClosureMetadata(consignmentId, fileIds))
-          } else {
-            Redirect(routes.AdditionalMetadataClosureStatusController.getClosureStatusPage(consignmentId, fileIds))
-          }
-        })
+
+    if (fileIds.nonEmpty) {
```
TomJKing, line 37: This nested if else statement is getting a bit long and deep. Wondering if its abstracting out a bit into a separate function to handle the different views for "closure" and "descriptive". Something like `selectView`?

app/controllers/AdditionalMetadataNavigationController.scala
```diff
@@ -32,20 +33,34 @@ class AdditionalMetadataNavigationController @Inject() (
...
+      }
     } else {
-      Future(Redirect(routes.AdditionalMetadataSummaryController.getSelectedSummaryPage(consignmentId, fileIds)))
+      getCachedFiles(consignmentId, metadataType, request).map(allFiles => {
+        BadRequest(views.html.standard.additionalMetadataNavigation(consignmentId, request.token.name, allFiles, metadataType, displayError = true))
+      })
     }
   }
+
+  private def getCachedFiles(consignmentId: UUID, metadataType: String, request: Request[AnyContent]): Future[ConsignmentService.File] = {
```
TomJKing, line 60: Does this cache need distinguishing between the different metadata types? Its always going to be the same regardless of the metadata type, as it contains all the filed ids for the consignment.

TomJKing (commented):
Looks good. Thanks. Just some minor comments.

app/controllers/AdditionalMetadataNavigationController.scala
vimleshtna: Yeah, that makes sense. I moved to a new method.

app/controllers/AdditionalMetadataNavigationController.scala
vimleshtna: Yeah, I was thinking if we are gonna show the green tick then this will be useful to distinguish between the different metadata types. But, if we decide to use not to show or use a different approach then we will remove the metadata type from the cache key.

TomJKing (approved):
Looks good. Thanks

PR #2391, opened 2022-11-28
TDR-2399 Add multiple foi exemption codes

thanhz:


app/controllers/util/CustomMetadataUtils.scala
```diff
@@ -62,6 +64,16 @@ class CustomMetadataUtils(allCustomMetadataProperties: List[CustomMetadata]) {
         TextField(property.name, fieldLabel, fieldDescription, property.multiValue, InputNameAndValue("years", property.defaultValue.getOrElse("")), "numeric", isRequired)
       case Text =>
         property.propertyType match {
+          case Defined if property.multiValue =>
```
TomJKing, line 67: This function is pretty long, and there is the nested `match` statement for the `Text` case. I'd be tempted to abstract out the logic for the `Text` case to a separate function.

app/views/partials/inputSingleCheckbox.scala.html
```diff
@@ -13,7 +13,9 @@
 @isSmallCheckbox = {@attributes.get(Symbol("_smallCheckbox"))}
 @isDynamicForm = @{attributes.getOrElse(Symbol("_dynamicForm"), false).asInstanceOf[Boolean]}
 @fieldId = @{if(isDynamicForm) s"inputcheckbox-$fieldName" else fieldName}
-    <div class="govuk-checkboxes @{if(fieldError.nonEmpty) "govuk-form-group--error" else ""}">
+    @if(!isDynamicForm) {
```
TomJKing, line 16: I'd avoid splitting the HTML up like this using `if` statements. It gets confusing to read. Probably better to keep it all in one `if` statement if that is possible, eg:
```
...
@if(!isDynamicForm) { 
   .... all the HTML needed for this ...
} else { 
   ... all the HTML for the default ...
}
...
```

app/views/partials/inputSingleCheckbox.scala.html
```diff
@@ -28,4 +30,6 @@
                 @label
             </label>
         </div>
-    </div>
+    @if(!isDynamicForm) {
```
TomJKing, line 33: See comment above about splitting the HTML across two `if` statements

app/views/partials/multipleCheckbox.scala.html
```diff
@@ -0,0 +1,49 @@
...
+        <div class="govuk-hint">
+        @fieldDescription
+        </div>
+
+        @errorMessage(fieldId, fieldErrors)
+        <div class="govuk-checkboxes" data-module="govuk-checkboxes" id="@fieldId">
+            <div class="govuk-grid-row">
+        @options.zipWithIndex.map {
+            case (foiCode, index) => {
+                @if(index == 0 || index % 15 == 0) {<div class="govuk-grid-column-one-third">}
```
TomJKing, line 29: What is this doing?

npm/css-src/sass/_multipleDropdown.scss
```diff
@@ -0,0 +1,207 @@
+.govuk-multi-autocomplete__wrapper {
```
TomJKing, line 1: Has this SASS been copied from a source library?

test/controllers/util/DynamicFormUtilsSpec.scala
```diff
@@ -176,17 +177,17 @@ class DynamicFormUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeAndA
     val dateTime = LocalDateTime.now().plusDays(1)
     val rawFormToMakeRequestWith =
       ListMap(
-        "inputdate-TestProperty3-day" -> List(dateTime.getDayOfMonth.toString),
-        "inputdate-TestProperty3-month" -> List(dateTime.getMonthValue.toString),
-        "inputdate-TestProperty3-year" -> List(dateTime.getYear.toString)
+        "inputdate-FoiExemptionAsserted-day" -> List(dateTime.getDayOfMonth.toString),
```
TomJKing, line 180: Is it only the FOI exemption asserted input date that has the future date validation on it?

TomJKing (commented):

When running locally got the following error in the console, although the functionality still worked OK.

image

Also  the download and export CSVs show all the selected FOI exemptions codes. Looks like the data is being saved to the DB OK.

app/views/partials/multipleCheckbox.scala.html
thanhz: In the prototype there are 3 columns for the check boxes, each column consisting of 15 check boxes. This just ensures there's a new column created for every 15 check boxes.

test/controllers/util/DynamicFormUtilsSpec.scala
thanhz: At the moment only _foiExemptionAsserted_ has a future date validation

npm/css-src/sass/_multipleDropdown.scss
TomJKing: @thanhz OK, would you mind checking with Gareth if he copied this from somewhere. If so we should put a comment in the file with the URL to where this SASS comes from.

app/controllers/util/CustomMetadataUtils.scala
thanhz: I have extracted the logic of the `Text` case to a separate function.

app/views/partials/inputSingleCheckbox.scala.html
thanhz: This makes sense. I've extracted out the common html and put everything into a single if else statement.

npm/css-src/sass/_multipleDropdown.scss
thanhz: I've put a comment in the SASS file with the link.

thanhz:
Closing this Pull request due to the use of inline style in the library causing the console error.

PR #2398, opened 2022-11-28
TDR-2595 Summary page save and return to file navigation

thanhz:
I branched of @vimleshtna  pr for the summary page as it already has the metadataType there.


TomJKing (approved):
Looks good. Thanks.

PR #2400, opened 2022-11-29
Update description for FOIExemptionCode

thanhz:
Updated the description for FOIExemptionCode for the form. I have hard coded the description because the properties of the form is driven from the database but we cannot have html in the database for the link.

### DIFF
--- a/app/views/partials/inputDropdown.scala.html
+++ b/app/views/partials/inputDropdown.scala.html
@@ -25,7 +25,9 @@ fieldErrors: List[String] = Nil
             </label>
         </h2>
         <div id="@formFieldId-hint" class="govuk-hint">
-            @fieldDescription
+             @* hard coding the description as the database cannot contain html*@
+            Add one or more exemption code to this closure. Here is a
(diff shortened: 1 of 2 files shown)


app/views/partials/inputDropdown.scala.html
```diff
@@ -25,7 +25,9 @@ <h2 class="govuk-label-wrapper">
             </label>
         </h2>
         <div id="@formFieldId-hint" class="govuk-hint">
-            @fieldDescription
+             @* hard coding the description as the database cannot contain html*@
```
TomJKing, line 28: Will this not override the any dropdown field that does not have a description? Think it would need to include an if statement to just apply the additional content for the FOI exemption code field?

TomJKing (commented):
Looks good. Just one comment. Thanks

app/views/partials/inputDropdown.scala.html
thanhz: You're right. I've added an if statement now.

TomJKing (approved):
Looks good. Thanks.

PR #2407, opened 2022-11-30
Refactor customMetadata Tests to use new FormTestData

thanhz:
I have replaced the `allProperties` declaration with a new `FormTestData` class. This should hopefully make it a bit easier to test the custom metadata form because it is easier to tell which property you are testing i.e. 'Dropdown' instead of 'TestProperty4'.


TomJKing (approved):
Looks good. Thanks for doing this.

PR #2421, opened 2022-12-05
TDR-2609 Retain selection for TitleClosed & DescriptionClosed when submitting partial form

thanhz:
This change makes it so that when submitting a partial form, the users selection for TitleClosed & Description Closed is retained.


TomJKing (approved):
Looks good. Thanks

PR #2430, opened 2022-12-07
TDR-2655 Add new card component to show selected files

thanhz:
Added a partial view to be reused for displaying the users selected files
This has been added to the 'Confirm closure status' and 'Edit closure metadata' and 'delete metadata' page.


app/views/partials/card.scala.html
```diff
@@ -0,0 +1,15 @@
+@(fileNames: List[String], metadataType: String, isEditing: Boolean = true)
```
TomJKing, line 1: I'd suggest passing the text into the partial template, instead of a parameter. That way the controllers can sort out what test should be displayed. Also the partial template holds the minimal amount of logic

npm/css-src/sass/_card.scss
```diff
@@ -1,3 +1,5 @@
+//This card component can be replaced by the '_card-metadata.scss' when the 'card' component on the additional metadata start page is updated
```
TomJKing, line 1: Could you check with Gareth whether the "card components" scss will be the same across all the pages. If so might aswell make this change in this PR.

TomJKing (commented):

npm/css-src/sass/_card.scss
thanhz: Confirmed with Gareth that the card components scss will be the same across all the pages.

app/views/partials/card.scala.html
thanhz: As suggested I am now passing the text into the partial template.

TomJKing (approved):
Looks good. Thanks

PR #2433, opened 2022-12-08
Remove slashes from the end of URLs so that URL parameters look nicer

techncl:
URL parameters looked like 'url/?parameter'; this change will make it look like this 'url?parameter'


thanhz (approved):
Looks good

thanhz (approved):

PR #2438, opened 2022-12-12
TDR-2717/2716 Display multivalues as a single entry on the summary page and sort by Ordinal

thanhz:
Screenshot from 2022-12-12 10-56-27
All fields are sorted by 'UiOrdinal'. The 'DescriptionClosed' and 'ClosureType' don't currently have a value in the 'FileProperty' table so they are displayed last.


TomJKing:
Merging on behalf of @thanhz

PR #2448, opened 2022-12-13
TDR-2642 - Implement - Alternative title, alternative description & description

vimleshtna:
- Revert closure metadata dependency page
- Implement Alternative title, alternative description & description on the main page
- Display DescriptionClosed field if the description is not empty


app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
+      formFields
     }
   }
 
-  private def getFilesFromConsignment(files: List[getConsignmentFilesMetadata.GetConsignment.Files]): List[File] = {
-    files.map(file => {
-      val filePath = file.fileMetadata.find(_.name == clientSideOriginalFilepath).map(_.value).getOrElse("")
-      File(file.fileId, filePath)
-    })
+  private def getDependenciesFromValue(customMetadataUtils: CustomMetadataUtils, propertyName: Set[String], valueToGetDependenciesFrom: String): Set[CustomMetadata] = {
```
TomJKing, line 129: Is there a reason the `propertyName` parameter is a `Set[String]`? Can it not be just a `String`?

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
+    val valuesByProperties: Map[String, List[CustomMetadata.Values]] = customMetadataUtils.getValuesOfProperties(propertyName)
```
TomJKing, line 130: Maybe rename this: `propertyToValues`

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
+    val allValuesForProperty: Seq[CustomMetadata.Values] = valuesByProperties(propertyName.head)
+    val value: Seq[CustomMetadata.Values] = allValuesForProperty.filter(_.value == valueToGetDependenciesFrom)
```
TomJKing, line 132: Maybe rename this val: `values`

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
-          consignmentId,
-          request,
-          metadataType,
-          isMainForm = true,
-          propertyNameAndFieldSelected,
-          formData.formFields,
-          fileIds
-        )
+        consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Option(FileFilters(None, Option(fileIds), None, None)))
+        defaultFieldForm <- getDefaultFieldsForForm(consignmentId, request)
```
TomJKing, line 42: Not sure this should be called `defaultFieldForm`, maybe better to call it: `formFields`, as it is only a single page, don't think these are "default" fields any more

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
   }
 
-  private def getDefaultFieldsForForm(
-      metadataType: String,
-      isMainForm: Boolean,
-      staticMetadata: Set[StaticMetadata],
-      consignmentId: UUID,
-      request: Request[AnyContent]
-  ): Future[FormData] = {
+  private def getDefaultFieldsForForm(consignmentId: UUID, request: Request[AnyContent]): Future[List[FormField]] = {
```
TomJKing, line 115: Maybe rename this: `getFieldsForForm`, as there is only a single page, so the fields aren't really "default"

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -34,248 +34,104 @@ class AddAdditionalMetadataController @Inject() (
...
-              (mainFormPageTitle.format(metadataType), mainFormPageDescription.format(metadataType))
-            }
+  }
 
-          getInfoForAddAdditionalMetadataPage(consignmentId, request, metadataType, isMainForm, fieldsAndValuesSelectedOnPrevPage, validatedFields, fileIds).map {
-            case (pageInfo, controllerInfo) =>
+  def addAdditionalMetadataSubmit(consignmentId: UUID, metadataType: String, fileIds: List[UUID]): Action[AnyContent] = standardTypeAction(consignmentId) {
+    implicit request: Request[AnyContent] =>
+      for {
+        defaultFieldValues <- cache.getOrElseUpdate[List[FormField]]("fieldValues") {
```
TomJKing, line 60: Maybe rename this: `formFields`as it fields that are returned, not "values"

TomJKing (commented):
I'm still working through this PR, as there are quite a few changes, but I've added some initial comments. Mostly these are around the naming of vals and methods.

app/controllers/util/DynamicFormUtils.scala
```diff
@@ -16,38 +16,49 @@ class DynamicFormUtils(request: Request[AnyContent], defaultFieldValues: List[Fo
 
   def validateAndConvertSubmittedValuesToFormFields(submittedValues: Map[String, Seq[String]]): List[FormField] = {
```
TomJKing, line 17: Maybe rename this to `convertSubmittedValuesToFormFields` The validation bit is implicit from calling the validation method.

TomJKing (commented):
Some additional comments. Thanks

app/controllers/AddAdditionalMetadataController.scala
vimleshtna: Yeah, that makes sense.

test/controllers/util/DynamicFormUtilsSpec.scala
```diff
@@ -173,6 +184,83 @@ class DynamicFormUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeAndA
     dateField.fieldErrors should equal(List(s"There was no number entered for the Day."))
   }
 
+  "validateAndConvertSubmittedValuesToFormFields" should "exclude to validate the field if it is hidden" in {
```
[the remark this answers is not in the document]
vimleshtna: Right, updated it.

app/controllers/AddAdditionalMetadataController.scala
vimleshtna: Actually, it's an existing code but yeah, we can use `String`.

app/controllers/AddAdditionalMetadataController.scala
vimleshtna: This is an old code and will be removed in the future. We will be using data from the new `DisplayProperties` table.
Anyway, I have updated this for now.

TomJKing (approved):
Looks good. Thanks

PR #2454, opened 2022-12-14
Display only files in list of files

techncl:
Previously, we weren't filtering just files in the list of files on the metadata pages; it was displaying the files and the folder. This PR removes the folder from the list and adds the filter that is sent to the API (when a request is made) to a utils file so that filtering is consistent on each page


TomJKing (approved):
Looks good. Thanks

PR #2456, opened 2022-12-15
TDR-2622 - Call the delete metadata service to delete the alternate title and description

vimleshtna:
Call the delete metadata service to delete the alternate title and alternate description if the user selected the "no" option for the TitleClosed and DescriptionClosed


TomJKing (approved):
Looks good. Thanks

PR #2470, opened 2022-12-20
Tdr 2649 add descriptive metadata form

TomJKing:


app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -125,14 +126,24 @@ class AddAdditionalMetadataController @Inject() (
...
-      dependencyProperties: Set[CustomMetadata] = getDependenciesForValue(customMetadataUtils, closureType.name, closureType.value)
-
-      formFields = customMetadataUtils.convertPropertiesToFormFields(dependencyProperties)
+      formFields =
+        if (closure) {
+          val customMetadataUtils = new CustomMetadataUtils(customMetadata)
+          val dependencyProperties: Set[CustomMetadata] = getDependenciesForValue(customMetadataUtils, closureType.name, closureType.value)
+          customMetadataUtils.convertPropertiesToFormFields(dependencyProperties)
+        } else {
+          val descriptiveProperties = displayProperties.filter(dp => dp.active && dp.propertyType.toLowerCase == metadataType)
```
vimleshtna, line 144: Can we move this filter part in the service so that we can directly assign the properties in the `displayProperties` variable at line number 133 and we can delete the if-else part at line number 133?

vimleshtna (commented):
Looks good, just a suggestion.

app/controllers/AddAdditionalMetadataController.scala
TomJKing: Yeah, that makes sense. Thanks

vimleshtna: I think we forgot to remove the filter at line number 140. It's not required now.

`displayProperties.filter(dp => dp.active && dp.propertyType.toLowerCase == metadataType)`

vimleshtna (commented):

vimleshtna (commented):
Looks good, thanks.

vimleshtna (approved):

PR #2471, opened 2022-12-20
Tdr 2759 non required field validation

TomJKing:


app/controllers/util/FormField.scala
```diff
@@ -138,20 +138,29 @@ object DropdownField {
...
+    textValue match {
+      case t if allCatch.opt(t.toInt).isEmpty => Some(numberError.format(inputName))
+      case t if t.toInt < 0                   => Some(negativeNumberError.format(inputName))
+      case _                                  => None
     }
+  }
+
+  private def toEmptyValueError(textField: TextField): Option[String] = {
+    // some fields are not required, but still require text if they are submitted
+    val overriddenFields: Set[String] = Set("TitleAlternate", "DescriptionAlternate")
```
vimleshtna, line 151: I was wondering, do we really need to hardcode the field Ids?
It should work without `overriddenFields`. If not, can we try some other way to make sure it should validate the empty value?

vimleshtna (commented):

app/controllers/util/FormField.scala
TomJKing: It doesn't work without the overridden fields. Otherwise it allows an empty value for the alternate title and description fields, which shouldn't be allowed. Can't make those fields "required" as they are only required if the is title/description closed fields are set to "true".

vimleshtna: I think we can make `isRequired` true while validating the dependencies.

https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/controllers/util/FormField.scala#L110

`case textField: TextField => TextField.validate(dependencies(textField.fieldId), textField.copy(isRequired = true))`

Based on the selected option, we know that this field is required so we can make the `textField` mandatory explicitly.

PR #2504, opened 2023-01-05
Use additional properties with metadata filter to get additional prop…

vimleshtna:
…erties that are required to show on the pages

### DIFF
--- a/app/controllers/AdditionalMetadataClosureStatusController.scala
+++ b/app/controllers/AdditionalMetadataClosureStatusController.scala
@@ -3,7 +3,7 @@ package controllers
 import auth.TokenSecurity
 import configuration.KeycloakConfiguration
 import controllers.util.InputNameAndValue
-import controllers.util.MetadataProperty.{clientSideOriginalFilepath, closureType}
+import controllers.util.MetadataProperty.{clientSideOriginalFilepath, closureType, fileType}
 import graphql.codegen.types.UpdateFileMetadataInput
 import org.pac4j.play.scala.SecurityComponents
 import play.api.cache.AsyncCacheApi
@@ -38,7 +38,13 @@ class AdditionalMetadataClosureStatusController @Inject() (
     implicit request: Request[AnyContent] =>
       for {
         details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
-        consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Some(metadataType), Some(fileIds))
+        consignment <- consignmentService.getConsignmentFileMetadata(
+          consignmentId,
+          request.token.bearerAccessToken,
+          Some(metadataType),
+          Some(fileIds),
+          Some(List(clientSideOriginalFilepath, fileType))
+        )
         response <-
           if (consignment.files.nonEmpty) {
             val filePaths = consignment.files.flatMap(_.fileMetadata).filter(_.name == clientSideOriginalFilepath).map(_.value)
@@ -73,7 +79,13 @@ class AdditionalMetadataClosureStatusController @Inject() (
           (consignmentRef, filePaths, parentFolderId) <- cache.getOrElseUpdate[(String, List[String], UUID)](s"$consignmentId-data", 1.hour)(
             for {
               details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
-              consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Some(metadataType), Some(fileIds))
+              consignment <- consignmentService.getConsignmentFileMetadata(
+                consignmentId,
+                request.token.bearerAccessToken,
+                Some(metadataType),
+                Some(fileIds),
+                Some(List(clientSideOriginalFilepath, fileType))
+              )
               filePaths = consignment.files.flatMap(_.fileMetadata).filter(_.name == clientSideOriginalFilepath).map(_.value)
             } yield (consignment.consignmentReference, filePaths, details.parentFolderId.get)
           )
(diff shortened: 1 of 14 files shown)


app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -38,7 +38,13 @@ class AdditionalMetadataClosureStatusController @Inject() (
     implicit request: Request[AnyContent] =>
       for {
         details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
-        consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Some(metadataType), Some(fileIds))
+        consignment <- consignmentService.getConsignmentFileMetadata(
+          consignmentId,
+          request.token.bearerAccessToken,
+          Some(metadataType),
+          Some(fileIds),
+          Some(List(clientSideOriginalFilepath, fileType))
```
TomJKing, line 46: The list of additional properties is used twice so I would extract it to a private val

app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -73,7 +79,13 @@ class AdditionalMetadataClosureStatusController @Inject() (
           (consignmentRef, filePaths, parentFolderId) <- cache.getOrElseUpdate[(String, List[String], UUID)](s"$consignmentId-data", 1.hour)(
             for {
               details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
-              consignment <- consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Some(metadataType), Some(fileIds))
+              consignment <- consignmentService.getConsignmentFileMetadata(
+                consignmentId,
+                request.token.bearerAccessToken,
+                Some(metadataType),
+                Some(fileIds),
+                Some(List(clientSideOriginalFilepath, fileType))
```
TomJKing, line 87: See comment above about abstracting list to a val

app/controllers/AdditionalMetadataSummaryController.scala
```diff
@@ -2,6 +2,7 @@ package controllers
 
 import auth.TokenSecurity
 import configuration.KeycloakConfiguration
+import controllers.util.MetadataProperty.fileType
```
TomJKing, line 5: Unused imports:
```
...
import controllers.util.MetadataProperty.fileType
...
import graphql.codegen.types.FileFilters
...
```

test/controllers/AddAdditionalMetadataControllerSpec.scala
```diff
@@ -590,6 +599,61 @@ class AddAdditionalMetadataControllerSpec extends FrontEndTestHelper {
     }
   }
 
+  "AddAdditionalMetadataController formFieldOverrides" should {
+    "override the field when the description is not empty" in {
```
TomJKing, line 603: Maybe rename the test to make clear which field is being overridden: `override the 'descriptionClosed' field when the ...`

test/controllers/AddAdditionalMetadataControllerSpec.scala
```diff
@@ -590,6 +599,61 @@ class AddAdditionalMetadataControllerSpec extends FrontEndTestHelper {
...
+        "some value",
+        false,
+        Seq(InputNameAndValue("Yes", "Yes")),
+        "Yes",
+        false
+      )
+      actualFormField should equal(expectedField)
+    }
+
+    "not override the field when the description is empty" in {
```
TomJKing, line 624: Maybe rename the test to make clear which field is being overridden: `not override the 'descriptionClosed' field when ...`

test/controllers/DeleteAdditionalMetadataControllerSpec.scala
```diff
@@ -65,6 +71,16 @@ class DeleteAdditionalMetadataControllerSpec extends FrontEndTestHelper {
         .apply(FakeRequest(GET, s"/consignment/$consignmentId/additional-metadata/confirm-delete-metadata/$closureMetadataType"))
       val deleteMetadataPage = contentAsString(response)
 
+      case class GraphqlRequestData(query: String, variables: gcfm.Variables)
```
TomJKing, line 74: This case class is used with three different sets of tests. Might be worth abstracting it to the `FrontEndTestHelper` class

test/services/ConsignmentServiceSpec.scala
```diff
@@ -269,42 +269,46 @@ class ConsignmentServiceSpec extends AnyWordSpec with MockitoSugar with BeforeAn
...
 
-      val getConsignmentDetails = consignmentService.getConsignmentFileMetadata(consignmentId, bearerAccessToken, None, None).futureValue
+      val getConsignmentDetails = consignmentService.getConsignmentFileMetadata(consignmentId, bearerAccessToken, None, None, additionalProperties).futureValue
 
       getConsignmentDetails.files.size should be(1)
       getConsignmentDetails.files.head.fileMetadata.head.value should be(exemptionCode)
     }
 
-    "return consignment with closure metadata when metadata type as 'closure' and fileIds are passed" in {
+    "return consignment with closure metadata when metadata type as 'closure', fileIds and additionalProperties are passed" in {
```
TomJKing, line 297: Fix grammar in the test name: `... metadata type is 'closure', ...`

TomJKing (commented):
Looks good. Some minor comments. I'm wondering whether to create a `trait` to encapsulate that some controllers require additional properties for the UI display. Something like this:

```
package controllers

trait AdditionalDisplayProperties {
  val additionalDisplayProperties: List[String]
}

```
Then have the controllers implement this, with the additional display properties clearly stated as a class val.

Not sure what you think?

vimleshtna:
> return consignment with closure metadata when metadata type as

@TomJKing Since, it's just a method call and we are passing it as a method argument so I don't think we need to create a `trait`. I think it's fine to pass it as list or create a local variable if required.

TomJKing (approved):
Looks good. Thanks

PR #2512, opened 2023-01-09
TDR-2767 Correctly format datetime datatypes for download csv

thanhz:
Similar to how the bagit export handles datetime fields. This allows the date fields to be viewable in the downloaded csv


TomJKing (approved):
Looks good. Thanks

PR #2524, opened 2023-01-10
Add feature access block to additional metadata features

TomJKing:


app/controllers/FileChecksResultsController.scala
```diff
@@ -20,13 +20,18 @@ class FileChecksResultsController @Inject() (
...
-    val frontEndInfoConfiguration: FrontEndInfoConfiguration
+    val frontEndInfoConfiguration: FrontEndInfoConfiguration,
+    val configuration: Config
 )(implicit val ec: ExecutionContext)
     extends TokenSecurity
     with I18nSupport {
 
   def fileCheckResultsPage(consignmentId: UUID): Action[AnyContent] = standardTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
     val pageTitle = "Results of your checks"
+
```
thanhz, line 31: Unnecessary line break

test/controllers/FileChecksResultsControllerSpec.scala
```diff
@@ -150,6 +158,75 @@ class FileChecksResultsControllerSpec extends FrontEndTestHelper {
...
+        val filePathResponse: String =
+          getConsignmentFilesClient.GraphqlData(Option(filePathData), List()).asJson.printWith(Printer(dropNullValues = false, ""))
+
+        mockGraphqlResponse(userType, fileStatusResponse, filePathResponse)
+        setConsignmentReferenceResponse(wiremockServer)
+
+        val fileCheckResultsController = new FileChecksResultsController(
+          getAuthorisedSecurityComponents,
+          keycloakConfiguration,
+          new GraphQLConfiguration(app.configuration),
```
thanhz, line 211: You can use the graphQLConfiguration used on line 177. Ideally we can extract out the instantiation of the FileChecksResultsController to a private method as this would reduce the amount code in this test spec but that's for another time.

thanhz (commented):
Looks good, just a few minor comments

test/controllers/FileChecksResultsControllerSpec.scala
TomJKing: The configuration is passed in a class constructor parameter so it needs to be created anyway.

PR #2528, opened 2023-01-11
TDR-2774 Add view transfers feature block

thanhz:


app/controllers/HomepageController.scala
```diff
@@ -39,7 +46,7 @@ class HomepageController @Inject() (val controllerComponents: SecurityComponents
   def judgmentHomepage(): Action[AnyContent] = secureAction { implicit request: Request[AnyContent] =>
     {
       if (request.token.isJudgmentUser) {
-        Ok(views.html.judgment.judgmentHomepage(request.token.name))
+        Ok(views.html.judgment.judgmentHomepage(request.token.name, blockViewTransfers))
```
TomJKing, line 49: I'm not sure we're releasing this functionality for judgment users, so could you just default this to `true` for the judgment template, until this is decided.

test/controllers/HomepageControllerSpec.scala
```diff
@@ -57,7 +60,10 @@ class HomepageControllerSpec extends FrontEndTestHelper {
     }
 
     "render the homepage page with an authenticated standard user" in {
-      val controller = new HomepageController(getAuthorisedSecurityComponents, getValidStandardUserKeycloakConfiguration, consignmentService)
+      val config: Config = ConfigFactory
```
TomJKing, line 63: Should make the the `false` the default config, as this will be how the page will be once the feature is released. 

The tests testing the page without the button should be the ones which override the config

test/resources/application.conf
```diff
@@ -11,6 +11,7 @@ consignmentapi.url="http://localhost:9006/graphql"
 featureAccessBlock {
   closureMetadata=true
   descriptiveMetadata=true
+  viewTransfers=true
```
TomJKing, line 14: This should be set to `false` as the page will show the button once the feature has been released, so this will reflect the final state, ie the feature will not be blocked following its release.

The final end start of the code will be to remove the feature blocks and the corresponding logic. This will occur once the feature has been successfully launched.

TomJKing (commented):

test/resources/application.conf
thanhz: As suggested, I have set the feature to false and updated the various tests

TomJKing (approved):
Looks good. Thanks

PR #2533, opened 2023-01-12
Collapse folder beforehand preventing delayed collapsing after page load

techncl:
Previously, when you load the page, the folder is expanded then the JavaScript (once the page has loaded) will collapse it; this is annoying because when you go to click a file, the folder then collapses


MancunianSam (approved):
Thanks for fixing this, it was very annoying.

PR #2547, opened 2023-01-17
Add call to the backend checks API

MancunianSam:
This can't be deployed until it's time to switch everything over to the
new backend checks process otherwise we'll end up with duplicate
metadata.


app/services/BackendChecksService.scala
```diff
@@ -0,0 +1,27 @@
...
+    val url = s"${configuration.get[String]("backendchecks.baseUrl")}/backend-checks/$consignmentId"
+    wsClient
+      .url(url)
+      .addHttpHeaders(("Authorization", token), ("Content-Type", "application/json"))
+      .post("{}")
+      .flatMap(r =>
+        r.status match {
+          case 200 => Future(true)
+          case _ =>
+            logger.error(s"Export api response ${r.status} ${r.body}")
```
TomJKing, line 22: Is that the correct log error, as this is about the backend checks, not the export api?

TomJKing (commented):
Looks good. Just one minor comment.

app/services/BackendChecksService.scala
MancunianSam: It isn't, I copied it from the export class. I've changed it.

MancunianSam:
There has been some javascript changes since you last looked at this if you want to check those over as well.

TomJKing (approved):
Looks good. Thanks

PR #2548, opened 2023-01-18
Update GraphQL version and DeleteFileMetadataInput propertyNames field

techncl:


TomJKing (approved):
Looks good. Thanks. Please hold off deploying until we get the all clear from the pen testers

PR #2565, opened 2023-01-23
TDR-2741 - Content changes on the additional metadata page

vimleshtna:


app/controllers/AdditionalMetadataController.scala
```diff
@@ -16,12 +16,24 @@ class AdditionalMetadataController @Inject() (
     val controllerComponents: SecurityComponents
 ) extends TokenSecurity {
 
+  val metadataProperties: Map[String, List[String]] = Map(
```
TomJKing, line 19: Could these values be added to the display properties table, if they are not there already?

TomJKing (commented):
Looks good. One minor question. Thanks

app/controllers/AdditionalMetadataController.scala
vimleshtna: Yeah, we can add them for closure and descriptive metadata types.
Not sure about the name of the attribute.

 PropertyName |   Attribute       |                 Value           | AttributeType 
--------------+---------------+---------------------------------------+---------------
 ClosurePeriod  | PropertyInfo   | Closure period             | text

Is that fine?

TomJKing: Can you ask Phil and Julius about this, but I'm happy with the `PropertyInfo` attribute name.

app/views/standard/additionalMetadataStart.scala.html
```diff
@@ -1,39 +1,64 @@
-@import views.html.partials.{transferReference, progressIndicator}
-@import java.util.UUID
+@import views.html.partials.{notificationBanner, progressIndicator, transferReference, warningMessage}
 
-@(parentFolder: String, parentFolderId: UUID, consignmentRef: String, consignmentId: UUID, name: String)(implicit messages: Messages, request: RequestHeader)
+@import java.util.UUID
+@(consignmentRef: String, consignmentId: UUID, name: String, closure: List[String], descriptive: List[String])(implicit messages: Messages, request: RequestHeader)
```
TomJKing, line 4: As per earlier comment maybe rename the `closure` and `descriptive` params to something a bit more descriptive.

conf/messages
```diff
@@ -46,3 +47,7 @@ transferComplete.progress=8
 
 progress.totalSteps=8
 judgmentProgress.totalSteps=3
+
+notification.title=Important
```
TomJKing, line 51: I'd make these a bit more specific than just `notification.`. Something like:

- `notification.savedProgress.title`
- `notification.savedProgress.heading`
- `notification.savedProgress.metadataInfo`

Potentially there could be other sorts of notifications in the future.

test/controllers/AdditionalMetadataClosureStatusControllerSpec.scala
```diff
@@ -346,25 +356,19 @@ class AdditionalMetadataClosureStatusControllerSpec extends FrontEndTestHelper {
...
-        |                                    FOI exemption code
-        |                                </li>
-        |                                <li>
-        |                                    Title public
-        |                                </li>
-        |                            </ul>""".stripMargin
-    )
+    closureStatusPage must include("""
+        |                            <ul class="govuk-list govuk-list--bullet govuk-list--spaced">
+        |XX
```
TomJKing, line 361: Whats the `XX` for, is it to ensure the correct spacing is applied on the HTML fragment? If so, could be named something a bit more descriptive?

test/controllers/AdditionalMetadataControllerSpec.scala
```diff
@@ -121,66 +145,60 @@ class AdditionalMetadataControllerSpec extends FrontEndTestHelper {
...
+                         |        <summary class="govuk-details__summary">
+                         |          <span class="govuk-details__summary-text">What ${name.toLowerCase} you can provide</span>
+                         |        </summary>
+                         |        <div class="govuk-details__text">
+                         |          <ul class="govuk-list govuk-list--bullet govuk-list--spaced">""".stripMargin)
+
+    if (metadataType.equals("closure")) {
+      page must include(
+        """          <ul class="govuk-list govuk-list--bullet govuk-list--spaced">
+          |XX
```
TomJKing, line 181: See earlier comment about `XX`

app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -78,9 +79,9 @@ class AdditionalMetadataClosureStatusController @Inject() (
     implicit request: Request[AnyContent] =>
       val errorFunction: Form[ClosureStatusFormData] => Future[Result] = { formWithErrors: Form[ClosureStatusFormData] =>
         for {
-          (consignmentRef, filePaths, parentFolderId) <- cache.getOrElseUpdate[(String, List[String], UUID)](s"$consignmentId-data", 1.hour)(
+          (consignmentRef, filePaths, closureProperties) <- cache.getOrElseUpdate(s"$consignmentId-data", 1.hour)(
             for {
-              details <- consignmentService.getConsignmentDetails(consignmentId, request.token.bearerAccessToken)
+              closureProperties <- displayPropertiesService.getDisplayProperties(consignmentId, request.token.bearerAccessToken, Some(metadataType)).map(_.map(_.summary))
```
TomJKing, line 84: You're setting this in the cache at line 55, so maybe this should be used?

app/controllers/AdditionalMetadataClosureStatusController.scala
```diff
@@ -51,7 +52,7 @@ class AdditionalMetadataClosureStatusController @Inject() (
           if (consignment.files.nonEmpty) {
             val filePaths = consignment.files.flatMap(_.fileMetadata).filter(_.name == clientSideOriginalFilepath).map(_.value)
             val areAllFilesClosed = consignmentService.areAllFilesClosed(consignment)
-            cache.set(s"$consignmentId-data", (consignment.consignmentReference, filePaths, details.parentFolderId.get), 1.hour)
+            cache.set(s"$consignmentId-data", (consignment.consignmentReference, filePaths, closureProperties), 1.hour)
```
TomJKing, line 55: I'd be tempted to move the caching into the display properties services, on the `getDisplayProperties` function.
Can then set the cache name based metadata type passed in, or simply cache the whole set of properties an filter the result based on the metadata type.

app/services/DisplayPropertiesService.scala
```diff
@@ -81,16 +83,37 @@ class DisplayPropertiesService @Inject() (val graphqlConfiguration: GraphQLConfi
...
   }
 
-  private def displayPropertyFilter(dp: DisplayProperty, metadataType: String): Boolean = {
-    dp.active && dp.propertyType.toLowerCase == metadataType.toLowerCase
+  private def displayPropertyFilter(dp: DisplayProperty, metadataType: Option[String]): Boolean = {
+    dp.active && metadataType.forall(`type` => dp.propertyType.equalsIgnoreCase(`type`))
   }
 
-  def getDisplayProperties(consignmentId: UUID, token: BearerAccessToken, metadataType: String): Future[List[DisplayProperty]] = {
+  def getDisplayProperties(consignmentId: UUID, token: BearerAccessToken, metadataType: Option[String]): Future[List[DisplayProperty]] = {
```
TomJKing, line 109: As mentioned earlier could move the caching to here, if its possible.

TomJKing (commented):
Looks good. Thanks. Just some minor comments. I've put something about the caching, but not sure how feasible it is.

conf/messages
vimleshtna: Yeah, that makes sense. I have renamed it, thanks.

app/controllers/AdditionalMetadataClosureStatusController.scala
vimleshtna: Yeah, we are using the cache at line 82.

app/services/DisplayPropertiesService.scala
vimleshtna: We can do this definitely but I feel it isn't required at the moment.

app/controllers/AdditionalMetadataClosureStatusController.scala
vimleshtna: We are already using the cache at some other places as well. Since we have to cache some other data too along with display properties, so it's better to use it at the controller level.

TomJKing (approved):
Looks good. Thanks.

PR #2566, opened 2023-01-23
TDR-2599 Closure metadata form to use display property table

thanhz:
Depends on https://github.com/nationalarchives/tdr-consignment-api-data/pull/246 being merged first.


app/controllers/util/DisplayPropertiesUtils.scala
```diff
@@ -55,6 +55,38 @@ class DisplayPropertiesUtils(displayProperties: List[DisplayProperty], customMet
...
+          property.displayName,
+          property.description,
+          property.multiValue,
+          InputNameAndValue("Day", "", "DD"),
+          InputNameAndValue("Month", "", "MM"),
+          InputNameAndValue("Year", "", "YYYY"),
+          required
+        )
+      case "radial" =>
+        val defaultOption = if (customMetadata.defaultValue.toBoolean) "yes" else "no"
```
TomJKing, line 71: There is a amount of logic for this field that justifies abstracting it out into its own function, eg `generateRadialField`. Same as you've done for the "small text" field

app/controllers/util/DisplayPropertiesUtils.scala
```diff
@@ -80,4 +112,23 @@ class DisplayPropertiesUtils(displayProperties: List[DisplayProperty], customMet
...
+      property.propertyName,
+      property.displayName,
+      property.description,
+      property.multiValue,
+      InputNameAndValue(inputName, customMetadata.defaultValue),
+      dataType,
+      required
+    )
+  }
+
```
TomJKing, line 133: Unnecessary line break.

test/controllers/util/DisplayPropertiesUtilsSpec.scala
```diff
@@ -17,17 +17,17 @@ class DisplayPropertiesUtilsSpec extends AnyFlatSpec with MockitoSugar with Befo
...
 
   "convertPropertiesToFormFields" should "return the list of properties requested in ordinal order" in {
-    val fields: Seq[FormField] = displayPropertiesUtils.convertPropertiesToFormFields
+    val fields: Seq[FormField] = displayPropertiesUtils.convertPropertiesToFormFields(displayProperties)
 
-    fields.size should equal(3)
-    fields.head.fieldId should equal("Dropdown1")
-    fields.tail.head.fieldId should equal("Dropdown2")
-    fields.last.fieldId should equal("Dropdown3")
+    fields.size should equal(6)
```
TomJKing, line 22: You can use the `.find` function on the list to find all the individual fields and check their properties, eg:
```
val closurePeriodField: FormField = fields.find(_.fieldName == "ClosurePeriod").get
closurePeriodField.fieldId should equal ... etc ...
```

TomJKing (commented):

test/controllers/util/DisplayPropertiesUtilsSpec.scala
thanhz: Since the test is checking if the properties are returned in ordinal order I've gone with something like `fields should equal(List("FoiExemptionAsserted", "ClosureStartDate", "ClosurePeriod", "Radio", "TestProperty2", "Dropdown"))`

TomJKing (approved):
Looks good. Thanks

PR #2575, opened 2023-01-25
Refactor dynamicFormUtils in order to make it easier to follow

techncl:
Replace Map[String, Seq[String]], for submitted values, with a class.

### DIFF
--- a/app/controllers/util/DynamicFormUtils.scala
+++ b/app/controllers/util/DynamicFormUtils.scala
@@ -3,12 +3,15 @@ package controllers.util
 import play.api.mvc.{AnyContent, Request}
 
 class DynamicFormUtils(request: Request[AnyContent], defaultFieldValues: List[FormField]) {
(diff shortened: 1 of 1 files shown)


app/controllers/util/DynamicFormUtils.scala
```diff
@@ -51,50 +54,49 @@ class DynamicFormUtils(request: Request[AnyContent], defaultFieldValues: List[Fo
           .copy(fieldErrors = RadioButtonGroupField.validate(selectedOption, dependencies, radioButtonGroupField))
 
       case textField: TextField =>
-        val text = fieldValue.head._2.head
+        val text = fieldValue.head.optionsSelected.head
```
MancunianSam, line 57: Do you think it's worth moving this variable out of the match statement? It's repeated a few times here.

MancunianSam (commented):
Looks good, just one small suggestion.

MancunianSam (approved):

PR #2581, opened 2023-01-27
TDR-2871 - Fix conditional element for radio button

vimleshtna:
Remove explicit code to show/hide conditional elements as the GovUK library will handle them automatically


TomJKing (approved):
Looks good. Thanks

PR #2583, opened 2023-01-27
Tdr 2754 implement new confirm transfer content changes

techncl:


TomJKing (commented):
Looks good. Thanks

Can you double check with @garethfoote about whether we're implementing the "back with the previous" step name just yet. I thought we were just going with a plain back button for the time being.

techncl:
> Looks good. Thanks
> 
> Can you double check with @garethfoote about whether we're implementing the "back with the previous" step name just yet. I thought we were just going with a plain back button for the time being.

It was in the design initially but he has confirmed that it was a mistake, so I'll change it back.

vimleshtna (approved):
Looks good, thanks.

PR #2588, opened 2023-01-30
TDR 2826 remove all english open records

MancunianSam:
Remove allEnglish option and initialOpenRecords parameters.

The controller tests are now done in a loop with the additional metadata set to true and false. It looks like a large change but it's mostly whitespace.

This will need https://github.com/nationalarchives/tdr-consignment-api/pull/547 to be merged and a new version of the generated graphql project before it can be merged.


TomJKing (commented):
Looks good. Thanks. I've not approved it as there is a merge conflict. 

Also, pretty sure this change will break the e2e tests. Could you make sure there is a PR ready to update the e2e tests based on this change.

app/controllers/TransferAgreementComplianceController.scala
```diff
@@ -21,28 +22,52 @@ class TransferAgreementComplianceController @Inject() (
...
+
+  private def form: Form[TransferAgreementComplianceData] =
+    if (blockClosureMetadata && blockDescriptiveMetadata) {
+      transferAgreementFormWithOpenRecords
+    } else {
+      transferAgreementForm
+    }
+
+  val blockClosureMetadata: Boolean = configuration.get[Boolean]("featureAccessBlock.closureMetadata")
+  val blockDescriptiveMetadata: Boolean = configuration.get[Boolean]("featureAccessBlock.descriptiveMetadata")
```
vimleshtna, line 62: It's just a suggestion. Feel free to ignore it.

Rename `FrontEndInfoConfiguration` to `ApplicationConfig` and get these config values from this class:

```

class ApplicationConfig @Inject()(configuration: Configuration) {

  private def get(location: String) = configuration.get[String](location)

  def frontEndInfo: FrontEndInfo = FrontEndInfo(
    get("consignmentapi.url"),
    get("environment"),
    get("region"),
    get("upload.url")
  )

  val blockClosureMetadata: Boolean = configuration.get[Boolean]("featureAccessBlock.closureMetadata")
  val blockDescriptiveMetadata: Boolean = configuration.get[Boolean]("featureAccessBlock.descriptiveMetadata")
}

```
Now, we can use these variables on the controllers. Actually, I don't want to hardcode the same string everywhere.

test/testUtils/FrontEndTestHelper.scala
```diff
@@ -83,6 +85,18 @@ trait FrontEndTestHelper extends PlaySpec with MockitoSugar with Injecting with
     }
   }
 
+  def getConfig(blockAdditionalMetadata: Boolean): Configuration = {
+    val config: Map[String, ConfigValue] = ConfigFactory
+      .load()
+      .withValue("featureAccessBlock.closureMetadata", ConfigValueFactory.fromAnyRef(blockAdditionalMetadata.toString))
+      .withValue("featureAccessBlock.descriptiveMetadata", ConfigValueFactory.fromAnyRef(blockAdditionalMetadata.toString))
+      .entrySet()
```
vimleshtna, line 93: I'm not sure if we already have test scenarios to test if `featureAccessBlock.closureMetadata` is true but `featureAccessBlock.descriptiveMetadata` is false or other way around. So in that case, the controllers (`TransferAgreementComplianceController` and `TransferAgreementPrivateBetaController`) should show the `transferAgreementForm`.

vimleshtna (commented):
Looks good, just a few suggestions, thanks.

test/testUtils/FrontEndTestHelper.scala
MancunianSam: Ah good point yeah. I've added in a table with each of the options for the two booleans and run the tests for all of them. It's in a loop so when we get rid of the feature block, we can just remove the outer loop.

app/controllers/TransferAgreementComplianceController.scala
MancunianSam: Yep, good point. I've moved these out the config, updated the controllers and updated the tests. I've also updated the FileCheckResults controller which is also using these feature block variables.

MancunianSam:
I've created a draft PR for the E2E tests but that won't work until this is merged https://github.com/nationalarchives/tdr-e2e-tests/pull/421

app/controllers/FileChecksResultsController.scala
```diff
@@ -20,16 +19,14 @@ class FileChecksResultsController @Inject() (
     val graphqlConfiguration: GraphQLConfiguration,
     val consignmentService: ConsignmentService,
     val consignmentStatusService: ConsignmentStatusService,
-    val frontEndInfoConfiguration: FrontEndInfoConfiguration,
-    val configuration: Config
+    val frontEndInfoConfiguration: ApplicationConfig,
+    val applicationConfig: ApplicationConfig
```
vimleshtna, line 23: You can remove the old one `frontEndInfoConfiguration`.

app/controllers/FileChecksController.scala
```diff
@@ -19,7 +19,7 @@ class FileChecksController @Inject() (
     val graphqlConfiguration: GraphQLConfiguration,
     val keycloakConfiguration: KeycloakConfiguration,
     val consignmentService: ConsignmentService,
-    val frontEndInfoConfiguration: FrontEndInfoConfiguration
+    val frontEndInfoConfiguration: ApplicationConfig
```
vimleshtna, line 22: Can you please rename `frontEndInfoConfiguration` to `applicationConfig`?

vimleshtna (commented):
Looks good, just a few minor comments, thanks.

app/controllers/FileChecksResultsController.scala
MancunianSam: Sorry, I was rushing. I've removed this and updated the tests.

vimleshtna (approved):
Looks good, thanks.

PR #2589, opened 2023-01-30
Add dependencies to FormField

techncl:
The previous code assumed:

1. Only Radio buttons questions can have dependencies - Any field could theoretically have a dependency
2. All Radio button questions have dependencies - That will probably not be true in the future


app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -95,17 +95,36 @@ class AddAdditionalMetadataController @Inject() (
       } yield result
   }
 
+  private def getDependenciesOfNonSelectedOptions(dependencies: Map[String, List[FormField]], nameOfOptionSelected: List[String]) =
+    dependencies.filterNot { case (name, _) => nameOfOptionSelected.contains(name) }.flatMap { case (_, fields) => fields.map(_.fieldId) }
+
+  // scalastyle:off cyclomatic.complexity
```
MancunianSam, line 101: We're not using scalastyle any more so you can get rid of this comment. There's a closing one on line 139 which can go as well.

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -117,6 +136,7 @@ class AddAdditionalMetadataController @Inject() (
       .saveMetadata(consignmentId, fileIds, request.token.bearerAccessToken, updateMetadataInputs)
       .map(_ => Redirect(routes.AdditionalMetadataSummaryController.getSelectedSummaryPage(consignmentId, metadataType, fileIds)))
   }
+  // scalastyle:on cyclomatic.complexity
```
MancunianSam, line 139: I know you didn't write this but it's going to need fixing. The call to `customMetadataService.deleteMetadata` on line 133 returns a Future but it isn't mapped to anything. If the http call completes before this has finished for whatever reason, we're going to get some weird behaviour.

I tried a couple of different bits to get it mapped with the if statement, this is one of them
```scala
    for {
      _ <- customMetadataService
        .saveMetadata(consignmentId, fileIds, request.token.bearerAccessToken, updateMetadataInputs)
      _ <-
        if (deleteMetadataNames.nonEmpty) {
          customMetadataService.deleteMetadata(fileIds, request.token.bearerAccessToken, deleteMetadataNames)
        } else {
          Future.successful()
        }
    } yield Redirect(routes.AdditionalMetadataSummaryController.getSelectedSummaryPage(consignmentId, metadataType, fileIds))
```
Something like that anyway.

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -95,17 +95,36 @@ class AddAdditionalMetadataController @Inject() (
...
+    val updateMetadataInputs: List[UpdateFileMetadataInput] = buildUpdateMetadataInput(updatedFormFields)
 
-    val deleteMetadataNames = updatedFormFields
+    val deleteMetadataNames: Set[String] = updatedFormFields
       .collect {
         case p if !updateMetadataInputs.exists(_.filePropertyName == p.fieldId) => p.fieldId :: Nil
-        case RadioButtonGroupField(_, _, _, _, _, _, selectedOption, _, _, dependencies, _) =>
-          dependencies.removed(selectedOption).flatMap { case (_, fields) => fields.map(_.fieldId) }
+        case RadioButtonGroupField(_, _, _, _, _, _, selectedOption, _, _, _, dependencies) if dependencies.nonEmpty =>
+          val nameOfOptionSelected: List[String] = List(selectedOption)
```
MancunianSam, line 112: I think this can be simplified a bit. The only variable part is how you get the selection option name so is it worth adding a method to the `FormField` abstract class like `selectedOptionNames` and putting the logic in there, then the case class can be simplified quite a bit?

I've put an example here https://github.com/nationalarchives/tdr-transfer-frontend/compare/addDependenciesPropertyToFormField...dependencies-example-branch?expand=1

Feel free to ignore this, I was finding the long case statement quite difficult to read but that might just be me.

app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -95,17 +95,36 @@ class AddAdditionalMetadataController @Inject() (
...
+        case TextField(_, _, _, _, nameAndValue, _, _, _, _, _, dependencies) if dependencies.nonEmpty =>
+          val nameOfOptionSelected: List[String] = List(nameAndValue.name)
+          getDependenciesOfNonSelectedOptions(dependencies, nameOfOptionSelected)
+        case TextAreaField(_, _, _, _, nameAndValue, _, _, _, _, _, dependencies) if dependencies.nonEmpty =>
+          val nameOfOptionSelected: List[String] = List(nameAndValue.name)
+          getDependenciesOfNonSelectedOptions(dependencies, nameOfOptionSelected)
+        case DropdownField(_, _, _, _, _, selectedOption, _, _, dependencies) if dependencies.nonEmpty && selectedOption.nonEmpty =>
+          val nameOfOptionSelected: List[String] = selectedOption.map(_.name).toList
+          getDependenciesOfNonSelectedOptions(dependencies, nameOfOptionSelected)
+        case MultiSelectField(_, _, _, _, _, _, selectedOptions, _, _, dependencies) if dependencies.nonEmpty && selectedOptions.nonEmpty =>
```
MancunianSam, line 123: Why is `DateField` missing off here?

MancunianSam (commented):
Looks good thanks. The simplified case statement is just a suggestion so feel free to tell me it won't work. The dangling Future does need sorting though, even though you didn't change it :smile:

app/controllers/AddAdditionalMetadataController.scala
techncl: IntelliJ uses it (We still have the scalastyle file in the main directory). I still find it kind of useful to have (as it highlights the warnings and makes you think of what your doing and if it could be simplified) but I might remove the Cyclomatic complexity rule (as well as the comment you mentioned) since that is the most annoying warning.

app/controllers/AddAdditionalMetadataController.scala
techncl: Good spot. Shouldn't it be the other way around though, `customMetadataService.deleteMetadata` before the `customMetadataService.saveMetadata` to ensure that we remove properties and dependencies before?

app/controllers/AddAdditionalMetadataController.scala
techncl: I think this was just an omission 🤦

app/controllers/AddAdditionalMetadataController.scala
techncl: Cool. This is a good idea! A lot cleaner, and would prevent omissions like the `DateField` one.

app/controllers/AddAdditionalMetadataController.scala
MancunianSam: It's a bit odd having two formatting tools going at once. I think the best bet is to remove the scalastyle xml file and I think you can then use Intellij to format using scalafmt.

app/controllers/AddAdditionalMetadataController.scala
MancunianSam: Good point yeah, the delete should be first :+1:

app/controllers/AddAdditionalMetadataController.scala
techncl: Cool, I've removed it

MancunianSam (approved):
:+1:

PR #2614, opened 2023-02-06
TDR-2811 summary description preserves spacing and line breaks

thanhz:
The description seems to already be saved correctly to the database. So this just uses the `<pre>` tag to preserve the line breaks and white spaces.

Screenshot from 2023-02-06 11-52-55

I have also changed the `textarea` field wrap to `soft` because a hard wrap inserts actual line breaks in the text at wrap points which was why the description had line breaks when copied to a text editor from the form.


TomJKing (approved):
Looks good. Thanks

PR #2615, opened 2023-02-06
TDR-2755 - Status tags for the file selection page

vimleshtna:
Required: https://github.com/nationalarchives/tdr-generated-graphql/pull/857


npm/css-src/sass/_card.scss
```diff
@@ -23,3 +23,69 @@
 .tdr-metadata-card .govuk-inset-text {
   color: $govuk-secondary-text-colour;
 }
+
+
+.tdr-tag {
```
TomJKing, line 28: Is this copied from the GDS?

npm/css-src/sass/_card.scss
```diff
@@ -23,3 +23,69 @@
...
+  outline: 2px solid transparent;
+  outline-offset: -2px;
+  padding-bottom: 2px;
+  padding-left: 8px;
+  padding-right: 8px;
+  padding-top: 3px;
+  text-decoration: none;
+  text-transform: uppercase;
+}
+
```
TomJKing, line 46: Unnecessary line break

npm/css-src/sass/_card.scss
```diff
@@ -23,3 +23,69 @@
...
+
+.tdr-tag--grey {
```
TomJKing, line 48: Are all these colours needed?

app/views/standard/additionalMetadataNavigation.scala.html
```diff
@@ -53,6 +53,22 @@ <h2 class="govuk-error-summary__title">There is a problem</h2>
     </div>
   }
 }
+@displayFileStatus(file: File, metadataType: String) = {
```
TomJKing, line 56: Is it possible to move this logic out of the template and into the controller? Its better to avoid putting too much logic into the templates. Maybe change the `File` case class to something like this:

```
case class StatusTag(text: String, colour: String)

case class File(id: UUID, name: String, fileType: Option[String], children: List[File], statusTag: StatusTag)
```
There needs to be something similar for the view transfers page, so it would be good to create a common object `StatusTag` that can be used for different tagging scenarios.

app/controllers/AdditionalMetadataNavigationController.scala
```diff
@@ -59,11 +60,4 @@ class AdditionalMetadataNavigationController @Inject() (
       Future(Redirect(routes.AddAdditionalMetadataController.addAdditionalMetadata(consignmentId, metadataType, fileIds)))
     }
   }
-
-  private def getCachedFiles(consignmentId: UUID, metadataType: String, request: Request[AnyContent]): Future[ConsignmentService.File] = {
```
TomJKing, line 63: Is there a reason for removing the caching?

TomJKing (commented):
Looks good. Just some minor comments

npm/css-src/sass/_card.scss
vimleshtna: Gareth shared these customized tags for the navigation component.
https://github.com/nationalarchives/tdr-prototype/commit/77b9e5460a88ad0c1e2e46b09f7e6cb5bd5a6f55

npm/css-src/sass/_card.scss
vimleshtna: Not required at the moment. I have removed them, thanks.

app/views/standard/additionalMetadataNavigation.scala.html
vimleshtna: Yeah, good point, thanks.

app/controllers/AdditionalMetadataNavigationController.scala
vimleshtna: Yeah, we had to remove this so that we can get the latest file status on the page load. I understand that we need to call again in case of an error (the user submits that page without selecting a file) but this is a very rare case.

TomJKing (approved):
Looks good. Thanks

PR #2622, opened 2023-02-07
Remove deprecated headers

MancunianSam:
The pen test has found the front end is returning a deprecated header
X-XSS-Protection

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

https://www.playframework.com/documentation/2.8.x/SecurityHeaders#Configuring-the-security-headers


TomJKing (approved):
Looks good. Thanks

PR #2629, opened 2023-02-09
Tdr 2323 more helpful error closure metadata form errors

thanhz:
Some of the errors were quite similar to each other for example 

> The date of the Advisory Council approval must be in the past

and 
> The closure start date must be in the past

so for simplicity and to avoid doing many if statements to check for the fieldName and output the different error messages I've just made the error messages generic to apply to both. So for the above case, the error for both will be

> The [Name of Field] must be in the past


app/controllers/util/FormField.scala
```diff
@@ -115,24 +118,22 @@ case class DateField(
...
+  val invalidDayError = "%s does not have %d days in it. Enter the day for the %s between 1 and %d"
+  val radioOptionNotSelectedError = "Select if the %s is sensitive to the public"
   val invalidRadioOptionSelectedError = "Option '%s' was not an option provided to the user."
   val tooLongInputError = "%s must be %s characters or less"
-
-  def inputModeToFieldType(inputMode: String): String = {
-    if (inputMode.equals("numeric")) "number" else "text"
-  }
+  val closurePeriodNotEnteredError = "Enter the number of years the record is closed from the closure start date"
+  val dependencyNotEntered = "Add an %s for this record"
```
MancunianSam, line 135: Will this always be `Add an` ? It works for field names beginning with a vowel but it sounds odd otherwise, e.g `Add an closure period for this record`

app/controllers/util/FormField.scala
```diff
@@ -210,14 +211,12 @@ object TextField {
...
-      val fieldName = if (textField.fieldId == "TitleAlternate") "alternative title" else textField.fieldName
-      Some(emptyValueError.format(fieldType, fieldName))
+      val fieldName = if (textField.fieldAlternativeName.isEmpty) textField.fieldName.toLowerCase else textField.fieldAlternativeName
+      if (textField.fieldId == "ClosurePeriod") Some(closurePeriodNotEnteredError.format(fieldName)) else Some(dependencyNotEntered.format(fieldName))
     } else if (textField.inputMode.equals("numeric")) {
-      val inputName = textField.nameAndValue.name
       text match {
-        case t if allCatch.opt(t.toInt).isEmpty => Some(numberError.format(inputName))
-        case t if t.toInt < 0                   => Some(negativeNumberError.format(inputName))
+        case t if allCatch.opt(t.toInt).isEmpty => Some(wholeNumberError.format(textField.fieldName.toLowerCase, "3, 15, 21"))
```
MancunianSam, line 218: The allCatch thing is nice, not seen that before :+1:

app/controllers/util/FormField.scala
```diff
@@ -255,12 +254,16 @@ object DateField {
...
-    var error = validateValue(day, "Day", invalidDayValidation)
-    error = if (error.isEmpty) validateValue(month, "Month", invalidMonthValidation) else error
-    error = if (error.isEmpty) validateValue(year, "Year", invalidYearValidation) else error
-    error = if (error.isEmpty) checkDayForTheMonthAndYear(day.toInt, month.toInt, year.toInt) else error
-    if (error.isEmpty) checkIfFutureDateIsAllowed(day.toInt, month.toInt, year.toInt, dateField) else error
+    val fieldName = if (dateField.fieldAlternativeName.isEmpty) dateField.fieldName.toLowerCase else dateField.fieldAlternativeName
+    if (day.isEmpty && month.isEmpty && year.isEmpty) {
+      Some(dateNotEnteredError.format(fieldName))
+    } else {
+      var error = validateValue(day, "Day", invalidDayValidation, "3, 15, 21", "31", fieldName)
```
MancunianSam, line 261: Can you replace the `var` with a series of vals please. We try not to use `var` except when it's unavoidable and I think you can do this without it. Something like:
```scala
      val dayError = validateValue(day, "Day", invalidDayValidation, "3, 15, 21", "31", fieldName)
      val monthError = if (dayError.isEmpty) validateValue(month, "Month", invalidMonthValidation, "3, 9, 12", "12", fieldName) else dayError
      val yearError = if (monthError.isEmpty) validateValue(year, "Year", invalidYearValidation, "1994, 2000, 2023", "", fieldName) else monthError
      val dayForMonthAndYearError = if (yearError.isEmpty) checkDayForTheMonthAndYear(day.toInt, month.toInt, year.toInt, fieldName) else yearError
      if (dayForMonthAndYearError.isEmpty) checkIfFutureDateIsAllowed(day.toInt, month.toInt, year.toInt, dateField) else dayForMonthAndYearError
```

test/controllers/util/FormFieldSpec.scala
```diff
@@ -23,20 +23,31 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
...
       updatedField shouldBe radioButtonGroupField.copy(selectedOption = "no")
     }
+
+    "validate should not return any error when the given value is valid" in {
+      List("yes", "no").foreach { validValue =>
+        RadioButtonGroupField.validate(validValue, Map.empty, radioButtonGroupField) shouldBe List()
+      }
+    }
+
+    "validate should an error when the given value is invalid" in {
```
MancunianSam, line 41: Think this should be `validate should return an error`

app/controllers/util/FormField.scala
MancunianSam: Could you put the "3, 15, 21" in a variable. It's in here a couple of times and it will make it easier to update.

MancunianSam (commented):
Looks good thanks, just a couple of things.

app/controllers/util/FormField.scala
thanhz: At the moment only 2 fields have `Add an`. Closure period has it's own error message. Example below
Screenshot from 2023-02-10 10-36-23

app/controllers/util/FormField.scala
thanhz: ah my bad, the var was there prior, your above suggestion works fine :+1:

app/controllers/util/FormField.scala
```diff
@@ -210,14 +212,12 @@ object TextField {
 
   def validate(text: String, textField: TextField): Option[String] =
     if (text == "") {
-      val fieldType: String = inputModeToFieldType(textField.inputMode)
-      val fieldName = if (textField.fieldId == "TitleAlternate") "alternative title" else textField.fieldName
-      Some(emptyValueError.format(fieldType, fieldName))
+      val fieldName = if (textField.fieldAlternativeName.isEmpty) textField.fieldName.toLowerCase else textField.fieldAlternativeName.toLowerCase
```
vimleshtna, line 215: Sorry, can we move this logic to `FormField`?

Maybe something like this:

```
abstract class FormField {
  val fieldId: String
  val fieldName: String
  val fieldAlternativeName: String
  val fieldDescription: String
  val multiValue: Boolean
  val isRequired: Boolean
  val hideInputs: Boolean = false
  val fieldErrors: List[String]
  val dependencies: Map[String, List[FormField]] = Map.empty
  def selectedOptionNames(): List[String]

  def getAlternativeName: String = {
    if (fieldAlternativeName.nonEmpty) fieldAlternativeName else fieldName
  }
}
```

We can use this new method to get the AlternativeName in all the `validate` methods.

app/controllers/util/FormField.scala
MancunianSam: I was more thinking for the future, if we update the database with new fields that need this. The rules are quite complicated around it so maybe we'll end up doing `Add a(n) xxx` or something. Guess we can leave it for now.

app/controllers/util/FormField.scala
thanhz: This sounds like a reasonable idea, since all fields can have an alternative name. I've made the changes as suggested

app/controllers/util/FormField.scala
thanhz: oh right I know what you mean. Julius mentioned something about having a separate table for errors but I guess we can think about it when we start to implement more fields.

MancunianSam (approved):
:+1:

PR #2641, opened 2023-02-13
Add missing legend to fieldset.

MancunianSam:
This came from the exploratory testing TDR-2845
There was an alert from Wave about a missing legend tag.
The fieldset it’s missing from isn’t from this ticket but
we have legends in other fieldsets to we should add it.


TomJKing (approved):
Looks good. Thanks

PR #2642, opened 2023-02-13
TDR-2747/2748 - Confirm delete metadata content changes

thanhz:


app/views/standard/confirmDeleteAdditionalMetadata.scala.html
```diff
@@ -1,27 +1,32 @@
...
 
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-two-thirds">
-                    <h1 class="govuk-heading-xl">
-                        Delete @metadataType metadata
-                    </h1>
-                    @fileNameCard(fileNames, Messages(s"deleteAdditionalMetadata.${metadataType}DeletionMessage"))
-                    @warningMessage(s"Once deleted $metadataType metadata cannot be recovered.")
-                    <p class="govuk-body">Are you sure you would like to proceed?</p>
+                    @form(routes.DeleteAdditionalMetadataController.deleteAdditionalMetadata(consignmentId, metadataType, fileIds), (Symbol("novalidate"), "")) {
```
TomJKing, line 12: Is there a reason for changing this to a form?

TomJKing (commented):
Looks good. Just one minor question. Thanks

app/views/standard/confirmDeleteAdditionalMetadata.scala.html
thanhz: The "delete and return to files" is now a button rather than a link so I had to define a route for the submit.

TomJKing (approved):
Looks good. Might be worthwhile testing the "delete" functionality still behaves correctly for closure and descriptive metadata.

PR #2650, opened 2023-02-15
Replace file list button with span.

MancunianSam:
The expander icon in the file list page was a button. If this was
clicked before the js loaded which changed its behaviour, it would
submit the form when clicked.

This has been replaced by a span which won't submit the form.


TomJKing (approved):
Looks good. Thanks

PR #2662, opened 2023-02-17
Tdr 2917 account for all file check statuses

TomJKing:


test/testUtils/ConsignmentStatusOptions.scala
```diff
@@ -26,40 +26,100 @@ object ConsignmentStatusesOptions {
...
+    ("client checks in progress", CurrentStatus(completed, completed, completed, inProgress, None, None, None, None, None), "/upload", "In Progress", "Resume transfer"),
+    ("client checks failed", CurrentStatus(completed, completed, completed, failed, None, None, None, None, None), "/upload", "Failed", "View errors"),
     (
       "client checks completed with issues",
       CurrentStatus(completed, completed, completed, completedWithIssues, None, None, None, None, None),
+      "/upload",
+      "Failed",
+      "View errors"
+    ),
+    ("client checks completed", CurrentStatus(completed, completed, completed, completed, None, None, None, None, None), "/file-checks", "In Progress", "Resume transfer"),
```
MancunianSam, line 38: This might make things more confusing so feel free to ignore it but there are a lot of common statuses here so is it worth putting then in a separate variable. You can then add Tables together. I got the tests passing like this https://github.com/nationalarchives/tdr-transfer-frontend/commit/7d0f6087fa265a80e7fcac45acb36f0ec165c9ae but I'm not sure if I've missed something.

The only other reason it might be worth it is that there will probably be more common statuses in the future

MancunianSam (commented):
Looks good, cheers. Just one suggestion for the tests but feel free to ignore it if it doesn't make sense.

test/testUtils/ConsignmentStatusOptions.scala
TomJKing: I did think of doing this, but I found it makes it more confusing.

PR #2671, opened 2023-02-20
Display correct column name for "TitleAlternate" on the summary page

thanhz:
The `fieldsWithAlternativeName` variable is for properties that we want to display the alternative name. There are cases whereby a property for example FoiExemptionAsserted will have an alternativeName but we still want to display the original name.


MancunianSam (approved):
Looks good, thanks.

PR #2672, opened 2023-02-20
TDR-2883 Add skip link to all pages.

MancunianSam:
I've added the GovUK skip to main content link to main.scala.html and
updated the tests that check the main.scala pages.

The rest is whitespace changes because the indentation was a bit off.


thanhz (approved):
Seems to work okay, Thanks

PR #2679, opened 2023-02-21
Make the metadata file selection work without js

MancunianSam:
I've set the file list to be expanded on default. This returns the
flicker problem where you see the expanded list and the js closes it but
I've raised a seperate ticket to deal with that.

I've changed here:
* The spans for the radio buttons and labels are now `<input>` and  `<label>` This allows them to be checked and allows them to be submitted in the form post.
* Removed the submit handler for this form from the js.
* Updated the controller to take the fileIds from the form. This is  the same behaviour whether js is enabled or not which keeps things consistent
* Moved the "Select the file.." text to the `<legend>` for a `fieldset` as the accessibility tools show warnings if these are missing.
* Updated the tests for the html


vimleshtna (approved):
Looks good, thanks. Now, it will be easy for me to implement view metadata functionality.

PR #2681, opened 2023-02-21
TDR-2772 Download metadata content updates

MancunianSam:
Content updates for the download metadata page.
CSS taken from the prototype so it should work.


thanhz (approved):
Looks good to me, page & button displaying correctly :+1:

PR #2693, opened 2023-02-22
TDR-2905 Add details to text area for description.

MancunianSam:
This ticket asks for the GDS details component to be added to the text area for description. As far as I know, there isn't anything suitable in DisplayProperties to deal with this so I've added two new rows for the description field,  DetailsSummary and DetailsText and used them to add the text to the form.

Matching DB changes. They can be deployed seperately but the details component won't display until the DB changes are in.
https://github.com/nationalarchives/tdr-consignment-api-data/pull/262


TomJKing (approved):
Looks good. Thanks

PR #2697, opened 2023-02-23
Add descriptive metadata card details.

MancunianSam:
This is another details component which describes what to do if there is
a sensitive description.

This relies on changes in https://github.com/nationalarchives/tdr-transfer-frontend/pull/2693


TomJKing (approved):
Looks good. Thanks

PR #2703, opened 2023-02-24
Get consignment replace current status

TomJKing:


app/controllers/ViewTransfersController.scala
```diff
@@ -16,40 +18,22 @@ import javax.inject.Inject
...
-    "fileChecksResults" -> routes.FileChecksResultsController.judgmentFileCheckResultsPage(consignmentId).url,
-    "export" -> routes.TransferCompleteController.judgmentTransferComplete(consignmentId).url
-  )
-
-  val statusColours: Map[String, String] = Map(transferStatusInProgress -> "yellow", transferStatusFailed -> "red", contactUs -> "red", transferStatusTransferred -> "green")
+  private val statusColours: Map[String, String] = Map(InProgress.value -> "yellow", Failed.value -> "red", ContactUs.value -> "red", Transferred.value -> "green")
+  private val formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm:ss")
+
+  implicit class ConsignmentStatusesHelper(statuses: List[ConsignmentStatuses]) {
+    def containsStatus(statusType: StatusType): Boolean = {
```
MancunianSam, line 25: This might make things more complicated but the case statement on line 77 is getting quite crowded. I know it's a complicated thing so it needs to be but I wonder if making this take multiple statuses would help. 
So change the method to something like
```scala
    def containsStatuses(statusTypes: StatusType*): Boolean = {
      statusTypes.foldLeft(false)((contains, statusType) => contains || statuses.exists(_.statusType == statusType.id))
    }
```
Then in the case statement, instead of having 
```scala
s.containsStatus(ServerAntivirusType) || s.containsStatus(ServerChecksumType) || s.containsStatus(ServerFFIDType)
``` 
you could do 
```scala
containsStatuses(ServerAntivirusType, ServerChecksumType, ServerFFIDType)
```
I'm not convinced this would help the readability at all and it might make it worse. What do you reckon?

app/controllers/ViewTransfersController.scala
```diff
@@ -62,111 +46,136 @@ class ViewTransfersController @Inject() (val consignmentService: ConsignmentServ
...
+    } else {
+      statuses
+    }
 
-      case s if s.confirmTransfer.isEmpty                                       => convertUploadAndFileCheckStatusesToUserActions(s, pageNameToUrlMap, consignmentRef)
-      case s if s.confirmTransfer.contains(completedStep) && s.`export`.isEmpty => UserAction(transferStatusInProgress, pageNameToUrlMap("fileChecksResults"), resumeTransfer)
-      case s if s.`export`.nonEmpty                                             => convertExportStatusesToUserActions(s, pageNameToUrlMap, consignmentRef, consignmentType)
+    statusesToCheck match {
+      case s if s.containsStatus(ExportType) => toExportAction(s.find(_.statusType == ExportType.id).get, judgmentType, consignmentId, consignmentRef)
+      case s if s.containsStatus(ConfirmTransferType) && s.statusValue(ConfirmTransferType).contains(CompletedValue.value) =>
```
MancunianSam, line 79: Again, not sure if this will make things more complicated to read but you don't need to check `containsStatus(ConfirmTransferType)` first. `s.statusValue(ConfirmTransferType).contains(CompletedValue.value)` will return false if the list doesn't have that status so you can remove it. 
Same applies to the Series check on line 86.

MancunianSam (commented):
Looks good, thanks. Just a couple of suggestions about making that case statement a bit shorter but it might make it worse so feel free to ignore me.

app/controllers/ViewTransfersController.scala
TomJKing: That does help

app/controllers/ViewTransfersController.scala
TomJKing: That's does help. Thanks

app/controllers/ViewTransfersController.scala
```diff
@@ -16,40 +18,23 @@ import javax.inject.Inject
...
-    "export" -> routes.TransferCompleteController.judgmentTransferComplete(consignmentId).url
-  )
-
-  val statusColours: Map[String, String] = Map(transferStatusInProgress -> "yellow", transferStatusFailed -> "red", contactUs -> "red", transferStatusTransferred -> "green")
+  private val statusColours: Map[String, String] = Map(InProgress.value -> "yellow", Failed.value -> "red", ContactUs.value -> "red", Transferred.value -> "green")
+  private val formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm:ss")
+
+  implicit class ConsignmentStatusesHelper(statuses: List[ConsignmentStatuses]) {
+    def containsStatuses(statusTypes: StatusType*): Boolean = {
+      // statuses.exists(_.statusType == statusType.id)
```
MancunianSam, line 26: I don't think you need the commented out line.

MancunianSam (approved):

PR #2715, opened 2023-02-28
TDR-2990 Add section breaks 

thanhz:
+ fix vertical spacing for textAreaFields

### DIFF
--- a/app/views/standard/addAdditionalMetadata.scala.html
+++ b/app/views/standard/addAdditionalMetadata.scala.html
@@ -1,18 +1,12 @@
 @* helper package contains inputRadioGroup therefore exclude from imports to prevent ambiguous import with view of the same name *@
-@import helper.{inputRadioGroup => _, _}
-@import java.util.UUID
-
-@import scala.language.postfixOps
-@import views.html.partials.{inputDate, inputText, errorSummary, inputTextArea}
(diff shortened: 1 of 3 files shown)


app/views/standard/addAdditionalMetadata.scala.html
```diff
@@ -1,18 +1,12 @@
 @* helper package contains inputRadioGroup therefore exclude from imports to prevent ambiguous import with view of the same name *@
+@import controllers.AddAdditionalMetadataController.File
+@import controllers.util.MetadataProperty.{description, end_date, foiExemptionCode}
+@import controllers.util._
 @import helper.{inputRadioGroup => _, _}
-@import java.util.UUID
+@import views.html.partials.{inputDate, inputText, _}
```
MancunianSam, line 6: Very minor thing but can you have this import as `@import views.html.partials._`

thanhz: There was some ambiguous imports due to the way we imported the helper package. So I've imported the helper package more explicitly, everything seems to still work.

MancunianSam: Cool, thanks. You've got some merge conflicts, I'll approve it once those are gone.

PR #2716, opened 2023-02-28
Remove maxlength character limit from textarea.

vimleshtna:


app/views/standard/addAdditionalMetadata.scala.html
```diff
@@ -75,7 +75,7 @@ <h1 class="govuk-heading-l">Add or edit metadata</h1>
                   inputMode,
                   addSuffixText
                 )
-                case TextAreaField(fieldId, fieldName, _, fieldDescription, _, _, inputNameAndValue,  _, fieldErrors, rows, wrap, characterLimit, details, _) => inputTextArea(
+                case TextAreaField(fieldId, fieldName, _, fieldDescription, _, _, inputNameAndValue,  _, fieldErrors, rows, wrap, _, details, _) => inputTextArea(
```
MancunianSam, line 78: Can we get rid of this field from the `TextAreaField` case class if we're not using it any more?

vimleshtna: We are still using it to validate the character limit in the backend - https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/controllers/util/FormField.scala#L253

MancunianSam (approved):

PR #2727, opened 2023-03-03
Content and page updates for public beta

TomJKing:
Only warning text behind the FAB as other changes still make sense for the current private beta version

Renaming of classes will be looked at separately

These changes will break the e2e tests. There is a draft PR to account for these changes: https://github.com/nationalarchives/tdr-e2e-tests/pull/446


MancunianSam (approved):
Looks good, thanks.

PR #2736, opened 2023-03-06
Rename transfer agreement classes

TomJKing:


MancunianSam (approved):
Looks good, thanks. Now imagine doing this in Python :smile:

PR #2737, opened 2023-03-06
TDR-2687 - Add date & timestamp to download report

thanhz:
I've used the existing datetime formatter on line 45 `DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss")` this results in a filename like this `TDR-2023-GW3K-2023-03-06T10 19 41-metadata.csv` any suggestions to what the format should be, would be appreciated.


TomJKing:
> like this `TDR-2023-GW3K-2023-03-06T10 19 41-metadata.csv` any suggestions to what the format should be

I would go for something without any spaces in it, given this will be a file name on the users file system. Maybe something like this: `TDR-2023-GW3K-2023-03-06T10-19-41`

PR #2739, opened 2023-03-06
TDR-3032 Disabled button accessibility changes.

MancunianSam:
As per Gareth's recommendations on the ticket, I've changed disabled to
aria-disabled and added a "reason-disabled" paragraph.

I've updated the js to set aria-disabled to true/false and to remove the
reason-disabled paragraph when the checks are complete.


TomJKing (approved):
Looks good. Thanks

PR #2740, opened 2023-03-06
Add govuk-template to html element.

MancunianSam:
As per GDS design template. I've also added in a line to check it in the
tests.


TomJKing (approved):
Looks good. Thanks

PR #2752, opened 2023-03-08
TDR 3003 additional metadata validation on submit

MancunianSam:
This adds a validate endpoint which checks all file statuses for an Incomplete status. If it finds any, it displays the errors on the page.


app/controllers/AdditionalMetadataController.scala
```diff
@@ -17,28 +20,66 @@ class AdditionalMetadataController @Inject() (
...
+        Redirect(routes.DownloadMetadataController.downloadMetadataPage(consignmentId))
+      }
+    }).recover(err => {
+      logger.error(err.getMessage, err)
+      Forbidden(views.html.forbiddenError(request.token.name, isLoggedIn = true, isJudgmentUser = false))
+    })
+  }
+
+  private def getErrors(statuses: GetConsignmentFiles.getConsignmentFiles.GetConsignment, metadataType: String, statusType: StatusType): Seq[(String, List[String])] = {
+    val incompleteCount = statuses.files.flatMap(_.fileStatuses).filter(_.statusType == statusType.id).count(_.statusValue == "Incomplete")
```
TomJKing, line 77: Could you use the case object value type instead of hardcoding "Incomplete" value, ie `IncompleteValue.value`

app/controllers/AdditionalMetadataController.scala
```diff
@@ -17,28 +20,66 @@ class AdditionalMetadataController @Inject() (
...
+      } else {
+        Redirect(routes.DownloadMetadataController.downloadMetadataPage(consignmentId))
+      }
+    }).recover(err => {
+      logger.error(err.getMessage, err)
+      Forbidden(views.html.forbiddenError(request.token.name, isLoggedIn = true, isJudgmentUser = false))
+    })
+  }
+
+  private def getErrors(statuses: GetConsignmentFiles.getConsignmentFiles.GetConsignment, metadataType: String, statusType: StatusType): Seq[(String, List[String])] = {
```
TomJKing, line 76: Could this method take a list of `metadataType`s so the retrieval of errors could be done in a single call?

TomJKing (commented):

TomJKing (approved):
Looks good. Thanks

PR #2757, opened 2023-03-10
Redirect users to the 'view transfers' page

TomJKing:
Temporary fix for https://national-archives.atlassian.net/browse/TDR-2265 was to redirect to the users to the homepage

Now the 'view transfers' page implemented users should be redirected there if signed out and back in with loss of state


MancunianSam (approved):
:+1:

PR #2763, opened 2023-03-10
TDR-1596 - Display include a top level Folder checkbox on the upload …

vimleshtna:
…page and pass its value to the API


app/views/standard/upload.scala.html
```diff
@@ -85,7 +85,18 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
                                 </div>
                             </div>
                             <p class="govuk-body">For information on what metadata will be captured during upload, visit the <a href="@routes.FaqController.faq()#metadata-captured" target="_blank" rel="noopener noreferrer" class="govuk-link">FAQ (opens in new tab)</a>.</p>
-
+                            <div class="govuk-form-group">
+                                <fieldset class="govuk-fieldset" aria-describedby="waste-hint">
+                                    <div class="govuk-checkboxes">
+                                        <div class="govuk-checkboxes__item">
+                                            <input class="govuk-checkboxes__input" id="includeTopLevelFolder" name="includeTopLevelFolder" type="checkbox" value="mines">
```
TomJKing, line 92: Should the `@value` be set to `mines`?

app/views/standard/upload.scala.html
```diff
@@ -85,7 +85,18 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
                                 </div>
                             </div>
                             <p class="govuk-body">For information on what metadata will be captured during upload, visit the <a href="@routes.FaqController.faq()#metadata-captured" target="_blank" rel="noopener noreferrer" class="govuk-link">FAQ (opens in new tab)</a>.</p>
-
+                            <div class="govuk-form-group">
+                                <fieldset class="govuk-fieldset" aria-describedby="waste-hint">
```
TomJKing, line 89: Should the `@aria-describedby` be set to `waste-hint`?

build.sbt
```diff
@@ -33,7 +33,7 @@ libraryDependencies ++= Seq(
   "com.softwaremill.sttp.client" %% "async-http-client-backend-future" % sttpVersion,
   "uk.gov.nationalarchives" %% "tdr-graphql-client" % "0.0.94",
   "uk.gov.nationalarchives" %% "tdr-auth-utils" % "0.0.120",
-  "uk.gov.nationalarchives" %% "tdr-generated-graphql" % "0.0.313",
+  "uk.gov.nationalarchives" %% "tdr-generated-graphql" % "0.0.314",
```
TomJKing, line 36: Is that the latest version of the graphql?

npm/src/upload/form/upload-form.ts
```diff
@@ -63,6 +64,10 @@ export class UploadForm {
     return value
   }
 
+  includeTopLevelFolder: () => boolean = () => {
+    return this.formElement.includeTopLevelFolder?.checked
```
TomJKing, line 68: Will this return `false` if the box is not checked?

TomJKing (approved):
Looks good. Just some minor comments. Thanks

build.sbt
vimleshtna: Updated with the latest version.

npm/src/upload/form/upload-form.ts
vimleshtna: Yes, I tested it and it's working as expected.

PR #2777, opened 2023-03-15
TDR-2818 Implement GDS Character count for TextAreaField

thanhz:
Right image shows user going over the character count as well as validation on submit (Character count set to 5 for testing purposes) 
CharacterCount


MancunianSam (approved):
Looks good, thanks.

PR #2801, opened 2023-03-21
Fix descriptive metadata form for new fields

vimleshtna:
A minor fix related to the generateTextField method to work based on the data type and also sort by name if uiOrdinal is not valid


test/controllers/util/DisplayPropertiesUtilsSpec.scala
```diff
@@ -388,24 +428,42 @@ class DisplayPropertiesUtilsSpec extends AnyFlatSpec with MockitoSugar with Befo
     displayPropertiesUtils.CustomMetadataHelper(None).defaultInput should equal(None)
   }
 
-  "CustomMetadataHelper" should "return the correct 'defined inputs' for custom metadata" in {
+  "CustomMetadataHelper.definedInputs" should "should sort by uiOrdinal when uiOrdinal is valid" in {
```
thanhz, line 431: You have the word "should" twice here in the test name.

PR #2818, opened 2023-03-24
Update page titles.

MancunianSam:
The accessibility audit has recommended that we change each of the page
titles to match the h1 of the page they're on.

I've updated each of these and added the - Transfer Digital Records - GOV.Uk
suffix they've recommended as well.

I've added a `defining` block to define the title and used this for both
the title and the h1 text so it'll be difficult to get them out of sync
accidentally.

There's a huge amount of whitespace changes because the `defining` block
needs everything else shifted forward. If you ignore whitespaces on the
PR, that might help.

On the subject of whitespace, some of the tests failed because of
whitespace changes. I've simplified/deleted these checks because I'm fed
up of trying to change whitespace in the tests.


vimleshtna (commented):
Looks good, thanks for doing this.  Just have a question related to the view metadata page.

vimleshtna (approved):
Looks good, thanks

PR #2822, opened 2023-03-27
Add accessibility statement page.

MancunianSam:
I've added the PDF with the original name so if anyone downloads it, it
will keep the original name.


app/views/accessibilityStatement.scala.html
```diff
@@ -0,0 +1,236 @@
...
+    </ul>
+
+
+    <p class="govuk-body">
+      As part of ongoing development, design iterations incorporate a progressive enhancement approach
+      and TDR is continuously tested to ensure that it meets WCAG 2.1 AA standards wherever possible.
+    </p>
+    <h2 class="govuk-heading-m">Feedback and contact information</h2>
+    <p class="govuk-body">If you need information on this website in a different format:</p>
+    <p class="govuk-body">Email: <a href="mailto:tdr@@nationalarchives.gov.uk">tdr@@nationalarchives.gov.uk</a></p>
```
TomJKing, line 38: Could you use the `Messages` for the email address to keep this consistent with the other pages: `nationalArchives.email` 

Also mean updating is easier should the email address change in the future

TomJKing (commented):
Looks good. Just one minor comment. Also looks like the unit tests are failing

app/views/accessibilityStatement.scala.html
MancunianSam: Good spot, I've swapped it out for the `@Messages` one.

MancunianSam:
The unit tests are OK now, They were checking for changes based on my other PR which hadn't been merged yet, it's in now so they're ok.

PR #2832, opened 2023-03-29
Update choose file page title.

MancunianSam:
This was changed so that the title matched the `<h1>` but the h1 is “Choose a file” for both pages so we’ve got the same problem.

I’ve checked with Gareth. He wants to keep the heading the same and change the title. I’ve udated the title to Descriptive metadata choose a file and Closure metadata choose a file


MancunianSam:
The h1 is the same and I checked the e2e tests and we don't seem to be checking the title anywhere so I think we're ok.

PR #2836, opened 2023-03-30
TDR-3115 Spacing between the details attributes on the metadata start page

thanhz:


thanhz:
> Looks good. Thanks. Is this going to effect the e2e tests?

It shouldn't break the e2e tests

PR #2837, opened 2023-03-30
Judgment guide updates.

MancunianSam:
I've tried to stick as closely to the Google doc as I can in HTML.
The only bit I've changed that I'm not sure of is in the
"Step 2: Upload the judgment" section, there doesn't seem to be any
place for the first file upload image. The new text seems to suggest
that the error image is the next one so I've removed the first image.

I've also update the two MFA images to show the new page headings on the
MFA pages. These images are shared with the standard guide so they
should be updated there as well.


thanhz (approved):
Thanks, Looks good to me

PR #2898, opened 2023-04-21
TDR-3109 - Show only "end date" if it is present in the DB else show …

vimleshtna:
…"last modified date" for the closure start date field

### DIFF
--- a/test/controllers/AddAdditionalMetadataControllerSpec.scala
+++ b/test/controllers/AddAdditionalMetadataControllerSpec.scala
@@ -23,6 +23,7 @@ import org.pac4j.play.scala.SecurityComponents
 import org.scalatest.matchers.should.Matchers._
 import play.api.Play.materializer
 import play.api.cache.AsyncCacheApi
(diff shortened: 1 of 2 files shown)


test/controllers/AddAdditionalMetadataControllerSpec.scala
```diff
@@ -778,7 +778,7 @@ class AddAdditionalMetadataControllerSpec extends FrontEndTestHelper {
       actualFormField should equal(expectedField)
     }
 
-    "add the 'end_date' and 'lastModifiedDate' when present to the 'closureStartDate' field" in {
+    "add the 'end_date' only when it is available for the 'closureStartDate' field" in {
```
TomJKing, line 781: Should there be a test to for showing the "last modified date" where there is no "end date"?

TomJKing (commented):
Looks good. One minor question. Can you also check that this change doesn't break the e2e tests. Thanks.

TomJKing (approved):
Looks good. Thanks

PR #2900, opened 2023-04-21
TDR-3151 Tree navigation improvements

thanhz:


TomJKing (commented):
Looks good. Thanks. Is there a reason we're changing the images from `svg` to `png`?

thanhz:
> Looks good. Thanks. Is there a reason we're changing the images from `svg` to `png`?

Reply from Gareth: 
_It’s a convoluted reason but mostly it was convenience. Because of the way the HTML is structured it became a real challenge to get the grey hover background working through CSS alone. I therefore resorted to embedding the background into the image and swapping the image out. I think mostly through force of habit I used a PNG because of the habit of using those in CSS background-images & also base64 encoding them.
There is no reason why they can’t be an SVG._

I think it's fine to merge this since the icons don't change with the screen size.

PR #2928, opened 2023-05-02
TDR-3160 - HTML and content changes for the metadata view/review page

vimleshtna:


app/views/partials/backLink.scala.html
```diff
@@ -1,2 +1,2 @@
-@(url: String, progress: String, pageName: String)(implicit messages: Messages)
-<a href="@url" class="govuk-back-link">Step @Messages(progress): @pageName</a>
+@(url: String, name: String)
```
thanhz, line 1: Probably call this parameter `text` instead of `name`

app/views/standard/additionalMetadataView.scala.html
```diff
@@ -3,6 +3,7 @@
 
 @import graphql.codegen.GetConsignmentFilesMetadata.getConsignmentFilesMetadata.GetConsignment.Files.FileMetadata
 
+@import views.html.partials.backLink
```
thanhz, line 6: Can add this to the existing import above

`@import views.html.partials.{transferReference, metadata, backLink}`

thanhz (approved):

PR #2932, opened 2023-05-03
Add Ids to find the element on the view and review metadata pages for…

vimleshtna:
… the e2e tests

### DIFF
--- a/app/views/standard/additionalMetadataReview.scala.html
+++ b/app/views/standard/additionalMetadataReview.scala.html
@@ -23,10 +23,10 @@
         @metadata(fileNames, metadataList)
 
         <div class="govuk-button-group">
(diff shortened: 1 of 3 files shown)


TomJKing (commented):
Do ids need to be added? Can the e2e tests not use the button text to locate the correct element?

Doesn't make sense to be changing the frontend to get the e2e tests to work.

vimleshtna:
> Do ids need to be added? Can the e2e tests not use the button text to locate the correct element?
> 
> Doesn't make sense to be changing the frontend to get the e2e tests to work.

We can fix it but the id is the best option to locate an element and it is a bit faster. Without Id, I can use only use the XPath to locate the element. I don't think any harm in adding the ID to the element.

What do you think?

TomJKing (approved):
Looks good. Thanks

PR #2993, opened 2023-05-22
TDR-3118 View transfers Incorrect status for export inProgress 

thanhz:
If export status is in-progress. The view transfer should still say "Transferred" because there's nothing further the user can do. Status can still change to failed if export fails. Might need to inform users if it does. But that will be covered in another story.


app/controllers/ViewTransfersController.scala
```diff
@@ -109,10 +109,10 @@ class ViewTransfersController @Inject() (
...
       else routes.DownloadMetadataController.downloadMetadataCsv(consignmentId).url
     status.value match {
-      case s if s == InProgressValue.value => UserAction(InProgress.value, url, actionText)
-      case s if s == CompletedValue.value  => UserAction(Transferred.value, url, actionText)
-      case s if s == FailedValue.value     => UserAction(Failed.value, s"""mailto:%s?subject=Ref: $consignmentRef - Export failure""", ContactUs.value)
-      case _                               => toContactUsAction(consignmentRef)
+      // Even though export is InProgress once a user clicks export there is nothing else they can do, hence setting the status to transferred
+      case s if s == InProgressValue.value || s == CompletedValue.value => UserAction(Transferred.value, url, actionText)
+      case s if s == FailedValue.value                                  => UserAction(Failed.value, s"""mailto:%s?subject=Ref: $consignmentRef - Export failure""", ContactUs.value)
+      case _                                                            => toContactUsAction(consignmentRef)
```
vimleshtna, line 115: Just to simplify a bit, you may remove the `if` from the `case` statement:
```

    status.value match {
      // Even though export is InProgress once a user clicks export there is nothing else they can do, hence setting the status to transferred
      case InProgressValue.value | CompletedValue.value => UserAction(Transferred.value, url, actionText)
      case FailedValue.value                            => UserAction(Failed.value, s"""mailto:%s?subject=Ref: $consignmentRef - Export failure""", ContactUs.value)
      case _                                            => toContactUsAction(consignmentRef)
    }
```

vimleshtna (commented):
Looks good, thanks.
Just a minor comment, feel free to ignore it.

app/controllers/ViewTransfersController.scala
thanhz: Yeah that's much simpler :+1:

PR #3008, opened 2023-05-25
TDR-3116 CSS for <address> attribute

thanhz:


vimleshtna (commented):
Looks good. We are using `<br />` tags in other places too, we need to remove them. Search with "`<address>`" in IntelliJ Idea, you will find all the references.

vimleshtna (approved):
Looks good, thanks

PR #3039, opened 2023-06-06
Fix disabled button issue on checking your record page

vimleshtna:
Use the `button-disabled` tdr component on checking your record page to prevent any button with attribute `data-tdr-module="button-disabled"` from being submitted


TomJKing (approved):
Looks good. Thanks

PR #3047, opened 2023-06-07
Fix screen reader issue

vimleshtna:
Replace `hidden="true"` with `govuk-visually-hidden` class so that the screen reader can read the uploaded folder


npm/src/upload/form/update-and-display-success-message.ts
```diff
@@ -47,6 +47,6 @@ export const displaySelectionSuccessMessage = (
   )
 
   successMessageContainer?.removeAttribute("hidden")
-  successMessage?.removeAttribute("hidden")
+  successMessage?.classList.remove("govuk-visually-hidden")
```
thanhz, line 50: Judgment upload page still uses the hidden attribute. I think you can just add `govuk-visually-hidden` to the judgment upload view and remove "hidden"

thanhz (commented):
Looks, good some a few comments

PR #3053, opened 2023-06-08
Tdr 3245 improve user error messaging

TomJKing:


app/views/partials/loggedOutErrorMessage.scala.html
```diff
@@ -5,7 +5,7 @@ <h2 class="govuk-error-summary__title logged-out-title" id="logged-out-error-tit
     Your session has timed out
   </h2>
   <div class="govuk-error-summary__body logged-out-body">
-    <p>Your session has ended because you were inactive for longer than 30 minutes. Please sign in again to continue from where you left off.</p>
+    <p>Your session has ended because you were inactive for longer than 60 minutes. Please sign in again to continue from where you left off.</p>
```
vimleshtna, line 8: Instead of hardcoding the session timeout, can we get it from the token?
I think you need to update the `tdr-auth-utils` to get the session timeout.

vimleshtna (commented):
Looks good, just a suggestion if we can dynamically fetch the session timeout. If it is not a quick one then we can leave it as hardcoded at the moment.

app/views/partials/loggedOutErrorMessage.scala.html
TomJKing: Not a straight forward way to do this. It would need to be set in the token, which would mean adding it to the user's profile. Not sure how that would be done. Or whether we should. Maybe something to look at in the future.

vimleshtna (approved):

PR #3057, opened 2023-06-09
Change 'label' to 'span'

TomJKing:
Following Gareth's investigation changed the 'label' element to 'span'.

See comment here: https://national-archives.atlassian.net/browse/TDR-3241?focusedCommentId=49859


vimleshtna (approved):
Looks good, thanks

PR #3073, opened 2023-06-13
TDR-3178 - Show back link and "Start a new transfer" link if the user doesn't have any consignments

vimleshtna:


thanhz (commented):
Looks good thanks. Just a question. Does the `start a new transfer` link need to work for judgment users too?

vimleshtna:
> need to work for judgment users too

Thanks for raising this point. I forgot to do it for the judgment user. I have added it now, thanks.

thanhz (approved):
Looks good thanks

PR #3078, opened 2023-06-14
TDR-3243 View Transfers showing Incorrect status for series

thanhz:
Closure&Descriptive statuses are inserted at the point of consignment creation. I've added to the pattern match this scenario so if a user cancels on the series select page. The view transfer page will show  the consignment as _In Progress_


vimleshtna (commented):
Looks good, just have a suggestion, I think we should use series status instead of the descriptive or closure metadata statuses for the series page.

PR #3083, opened 2023-06-16
TDR-3244 View transfer correct series status

thanhz:
As suggested in the previous [PR](https://github.com/nationalarchives/tdr-transfer-frontend/pull/3078#discussion_r1231912019). This will now set the series to "InProgress' on first load. This also fixes the View transfer displaying _Contact Us_ instead of _In progress_ when users goes to select a series then clicks cancel.

Related api change pr: https://github.com/nationalarchives/tdr-consignment-api/pull/632


app/controllers/SeriesDetailsController.scala
```diff
@@ -79,6 +79,9 @@ class SeriesDetailsController @Inject() (
               .uncache()
           )
         case _ =>
+          if (!seriesStatus.contains(InProgressValue.value)) {
```
vimleshtna, line 82: Looks good, can you please add a unit test to verify if the `seriesStatus` is `None` then it should add a consignment status and if it is in progress then it shouldn't.

vimleshtna (commented):
Thanks for doing this. Looks good, just asked to add a couple of unit tests, thanks.

app/controllers/ViewTransfersController.scala
```diff
@@ -96,8 +96,8 @@ class ViewTransfersController @Inject() (
       case s if s.containsStatuses(ClientChecksType, UploadType) => toClientSideChecksAction(statuses, consignmentId, judgmentType)
       case s if s.containsStatuses(TransferAgreementType) =>
         toTransferAgreementAction(s.find(_.statusType == TransferAgreementType.id).get, consignmentId)
-      case s if s.statusValue(SeriesType).contains(CompletedValue.value) =>
-        UserAction(InProgress.value, routes.TransferAgreementPart1Controller.transferAgreement(consignmentId).url, Resume.value)
+      case s if s.containsStatuses(SeriesType) =>
```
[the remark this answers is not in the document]
thanhz: Thanks, I forgot about this

app/controllers/SeriesDetailsController.scala
thanhz: As suggested I've added unit tests to cover those scenarios. I had to move the logic into its own function to make the testing a bit more manageable.

app/controllers/SeriesDetailsController.scala
```diff
@@ -97,6 +95,12 @@ class SeriesDetailsController @Inject() (
     }
     DropdownField(form("series").id, "", "", "", Nil, multiValue = false, options, None, isRequired = true, errors.toList)
   }
+
+  def addSeriesStatus(seriesStatus: Option[String], consignmentId: UUID, request: Request[AnyContent]) = {
```
vimleshtna, line 99: Sorry, just a small thing, actually, if we are not using any method outside of the class then we should make it private but I think you can still test it with the main method.

vimleshtna (commented):
Looks good to me, just a small suggestion.

app/controllers/SeriesDetailsController.scala
thanhz: No worries. I've moved the method back into the main method and made changes to the tests.

vimleshtna (approved):
Looks good, thanks

PR #3106, opened 2023-06-23
Remove Back to homepage button from the view transfer page

vimleshtna:


TomJKing (commented):
Looks good, thanks.

Think this change will break the e2e tests: https://github.com/nationalarchives/tdr-e2e-tests/blob/master/src/test/resources/features/ViewTransfers.feature#L8

If so, could you create a PR for the e2e tests first, before merging these changes.

vimleshtna:
> s change will break the e2e

Thanks, updated the existing one add a new test to test the new behaviour https://github.com/nationalarchives/tdr-e2e-tests/pull/508

TomJKing (approved):
Looks good. Thanks.

PR #3148, opened 2023-07-05
Tdr 3248 delete metadata consignmentid required

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-generated-graphql/pull/1002

This is a breaking change so needs reviewing first before updating so unit tests pass.


tna-aford (approved):
lgtm, let me know when you need re-review

tna-aford (approved):

PR #3239, opened 2023-08-10
Show an error if the alternate title field is empty

vimleshtna:
Show an error if the alternate title field is empty only when user is selected "yes" for the TitleClosed field


app/controllers/util/FormField.scala
```diff
@@ -164,7 +164,7 @@ object RadioButtonGroupField {
     def dependenciesError: Option[String] = radioButtonGroupField.dependencies
       .get(option)
       .flatMap(_.flatMap {
-        case textField: TextField         => TextField.validate(dependencies(textField.fieldId), textField)
+        case textField: TextField         => TextField.validate(dependencies(textField.fieldId), textField.copy(isRequired = true))
```
TomJKing, line 167: Does this impact other text fields that are not required and left empty?

TomJKing (commented):

app/controllers/util/FormField.scala
vimleshtna: I don't think so as it's only required for the radio button field. In this case, if the user selects "yes" then a dependent field is required. For the other TextField, we are directly calling the `TextField.validate` method.

TomJKing (approved):
Looks good. Thanks

PR #3253, opened 2023-08-16
TDR-3161 - Use TNA header from TDR Components library

vimleshtna:


thanhz (commented):
Looks good. There are some scss that probably needs removing from what I can see on Gareth's [pull request](https://github.com/nationalarchives/tdr-components/pull/48/files)

vimleshtna:
> Looks good. There are some scss that probably needs removing from what I can see on Gareth's [pull request](https://github.com/nationalarchives/tdr-components/pull/48/files)

Good point I have removed them, thanks

PR #3293, opened 2023-08-31
Fix screen reader issue (announcing multiple times) by modifying the dom once instead of twice

vimleshtna:


TomJKing:
Looks like there are some linting issues with the PR which is why the tests are failing.

vimleshtna:
> Looks like there are some linting issues with the PR which is why the tests are failing.

It was a JS formatting issue. I fixed that.

TomJKing (approved):
Looks good. Thanks

PR #3329, opened 2023-09-13
TDR-3225 - Show specific error message when user uploads an empty folder

vimleshtna:


TomJKing (commented):
Looks good. Thanks

Does the user still see the single folder error if they try to drag more than one folder?

vimleshtna:
> ee the single folder error if they try to drag more than one folder?

Yes, it will show this when a user tries to upload multiple folders - "You can only upload one top-level folder per consignment. However, that folder can contain multiple files and sub folders."

TomJKing (approved):
Looks good. Thanks

PR #3433, opened 2023-10-27
Use validation library to validate closure and descriptive metadata

vimleshtna:


app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -129,6 +129,33 @@ class AddAdditionalMetadataController @Inject() (
     consignmentService.getConsignmentFileMetadata(consignmentId, request.token.bearerAccessToken, Some(metadataType), Some(fileIds), additionalProperties)
   }
 
+  private def getMetadataValidation(consignmentId: UUID)(implicit request: Request[AnyContent]): Future[MetadataValidation] = {
```
TomJKing, line 132: Would it make more sense to have this method in the `MetadataValidationUtils.scala`?

TomJKing (commented):
Looks good. Thanks. One minor comment. 

Also could you update the exploratory test Jira task with test cases to cover all the scenarios for adding form data: https://national-archives.atlassian.net/browse/TDR-3420

app/controllers/AddAdditionalMetadataController.scala
vimleshtna: I created this `MetadataValidationUtils.scala` to process data. I don't think it's a good idea to call services from the utility class.

TomJKing: Fair enough

test/controllers/util/FormFieldSpec.scala
```diff
@@ -513,6 +562,28 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
       })
     }
 
+    "updateError should set an error message as per the given error code" in {
```
TomJKing, line 565: Can't see a check for the max character limit error in this test?

TomJKing (commented):

test/controllers/util/FormFieldSpec.scala
vimleshtna: Actually, the date field will never have a max character limit error. It is only applicable for `TextArea` and `RadioButtonGroupField` (dependency).

TomJKing: Fair enough

TomJKing (approved):
Looks good. Thanks.

TomJKing (approved):

PR #3438, opened 2023-10-30
update multiselect error message

thanhz:
image


test/controllers/AddAdditionalMetadataControllerSpec.scala
```diff
@@ -296,7 +296,7 @@ class AddAdditionalMetadataControllerSpec extends FrontEndTestHelper {
       )
       addAdditionalMetadataPageAsString should include("""    <p class="govuk-error-message" id="error-FoiExemptionCode">
           |        <span class="govuk-visually-hidden">Error:</span>
-          |        Search for and select at least one FOI exemption code(s)
+          |        Select at least one FOI exemption code(s)
```
TomJKing, line 299: Is the `...(s)` needed on the word `code`. Make sense without it

TomJKing (commented):

test/controllers/AddAdditionalMetadataControllerSpec.scala
thanhz: I think this was based on an old copy change. But makes sense to remove it.

TomJKing (approved):
Looks good. Thanks

TomJKing (approved):
Looks good. Thanks

PR #3499, opened 2023-11-28
TDR-3376 - Check closure status on closure form page 

thanhz:
To prevent from accessing the closure form URL without confirming the closure status, we need to check the closure status on page load. If the ClosureType is Open then redirect to the closure status page else load the closure metadata page successfully

### DIFF
--- a/app/controllers/AddAdditionalMetadataController.scala
+++ b/app/controllers/AddAdditionalMetadataController.scala
@@ -43,32 +43,14 @@ class AddAdditionalMetadataController @Inject() (
     implicit request: Request[AnyContent] =>
       for {
         consignment <- getConsignmentFileMetadata(consignmentId, metadataType, fileIds)
-        isOpen = consignment.files.flatMap(files => files.fileMetadata.find(fmetadata => fmetadata.name == closureType.name && fmetadata.value == "Open")).nonEmpty
+        isClosureOpen = consignment.files.flatMap(files => files.fileMetadata.find(fmetadata => fmetadata.name == closureType.name && fmetadata.value == "Open")).nonEmpty
         result <-
-          if (isOpen) {
+          if (isClosureOpen) {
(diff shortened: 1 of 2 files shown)


app/controllers/AddAdditionalMetadataController.scala
```diff
@@ -43,20 +43,14 @@ class AddAdditionalMetadataController @Inject() (
...
-          val fileExtension = consignment.files.getFileExtension
-          Future.successful(updateFormFields(formFields, metadataMap, fileExtension))
-        }
-      } yield {
-        Ok(
-          views.html.standard
-            .addAdditionalMetadata(consignmentId, consignment.consignmentReference, metadataType, updatedFormFields, request.token.name, consignment.files.toFiles)
-        ).uncache()
-      }
+        isClosureOpen = consignment.files.flatMap(files => files.fileMetadata.find(fmetadata => fmetadata.name == closureType.name && fmetadata.value == "Open")).nonEmpty
```
TomJKing, line 46: Maybe rename this to: `closureStatusOpen`

test/controllers/AddAdditionalMetadataControllerSpec.scala
```diff
@@ -245,6 +245,21 @@ class AddAdditionalMetadataControllerSpec extends FrontEndTestHelper {
       playStatus(addAdditionalMetadataPage) mustBe FOUND
     }
 
+    "return a redirect to the additionalMetadataClosureStatus page if the closure status isn't closed" in {
```
TomJKing, line 248: Could you add another specific test for the correct redirect when a file is `closed`. It just makes sure that this is tested independently, instead of being indirectly checked in other unit tests.

TomJKing (commented):
Looks good. Just a couple of minor comments. Thanks.

TomJKing (approved):
Looks good. Thanks

PR #3511, opened 2023-12-01
TDR-3587 - Return an error if integer is 0

thanhz:
Closure period cannot be 0.
Requires https://github.com/nationalarchives/tdr-metadata-validation/pull/22


TomJKing (approved):
Looks good. Thanks

PR #3534, opened 2023-12-12
User session timeout temporarily increased

TomJKing:
This is to accomodate a larger consignment from the Welsh Government which is taking a long to time to complete its upload due to network issues


Tom-Hallett (approved):
All good.

PR #3583, opened 2024-01-05
Display a character count status for the alternative description field

vimleshtna:


thanhz (commented):
Looks good. Feel free to ignore if it's too much of a hassle but I was wondering if it was possible to reuse the partial template `inputTextArea.scala.html`?

vimleshtna:
> Looks good. Feel free to ignore if it's too much of a hassle but I was wondering if it was possible to reuse the partial template `inputTextArea.scala.html`?

Unfortunately, we can't use that template as we need to show the text area based on the radio button selection.

thanhz (approved):

PR #3588, opened 2024-01-12
TDR-3342 : Fix error messages for invalid date entries

annielh:
- Use `java.time` functions instead of hand rolled logic in `checkDayForTheMonthAndYear` to ensure all edge cases handled out of the box.
- Refactor to enable use of single `checkDayForTheMonthAndYear` method as opposed to duplication of logic.
- Split date validation into (first) basic format validation then range validation for specific fields so that year and month can be validated upfront as they are dependencies for constructing a full and descriptive error message for incorrect date entries
- **NB** the order of precedence for surfacing error messages has been changed from day -> month -> year to year -> month -> day. I believe this is necessary to meet the requirements of the ticket, as we need a valid month/year pair to properly generate the accurates day error messages requested. I have updated the tests to try to make this change in logic explicit- but do flag if anyone can think of a particular design reason for the previous order.

On entering 32/11/2022:
|Before|After|
|---|---|
|Screenshot from 2024-01-15 11-30-52|Screenshot from 2024-01-15 11-26-09|
---
[Jira Ticket](https://national-archives.atlassian.net/jira/software/projects/TDR/boards/6?assignee=712020%3A8ca26fc0-b803-477b-b243-2ed437bebc1f&selectedIssue=TDR-3342) | Depends on this [metadata library ticket](https://github.com/nationalarchives/tdr-metadata-validation/pull/26) - will update sbt build file with new version once released


thanhz (commented):
Hi Emma, have you tried updating the build.sbt validation library to the version with your changes? I get an error when trying to submit a form.

thanhz (commented):
Looks good, Thanks! Just need to update the build.sbt

thanhz (approved):

PR #3620, opened 2024-01-24
TDR-3680 - Clean up old validation code

vimleshtna:
Clean up old validation code as we will be using only metadata validation library now


test/controllers/util/FormFieldSpec.scala
```diff
@@ -75,52 +75,6 @@ class FormFieldSpec extends AnyWordSpec with MockitoSugar with BeforeAndAfterEac
         .fieldErrors shouldBe List("alternativeDesc must be 8000 characters or less")
       RadioButtonGroupField.updateError(radioButtonGroupField, CLOSURE_STATUS_IS_MISSING).fieldErrors shouldBe List("CLOSURE_STATUS_IS_MISSING")
     }
-
```
thanhz, line 78: Unused dependency` org.apache.commons.lang3.NotImplementedException` and method ` def getDate(dateTime: LocalDateTime)`

test/controllers/util/DynamicFormUtilsSpec.scala
```diff
@@ -77,318 +77,6 @@ class DynamicFormUtilsSpec extends AnyFlatSpec with MockitoSugar with BeforeAndA
     thrownException.getMessage should equal(s"$fieldTypeWithIncorrectPrefix is not a supported field type.")
   }
```
thanhz, line 79: Unused dependency `import java.time.LocalDateTime`

thanhz (commented):
Thanks, looks good. Just some minor comments and a question regarding the caching.

If we add/remove fields from the form will the users see these changes since the cache never expires?

vimleshtna:
> Thanks, looks good. Just some minor comments and a question regarding the caching.
> 
> If we add/remove fields from the form will the users see these changes since the cache never expires?

Yeah, it should work because I have done the caching over the custom metadata, not the user data so it should work.

thanhz (approved):
Looks good, thanks

PR #3685, opened 2024-02-14
Trigger metadata validation

TomJKing:


app/services/ValidateMetadataService.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.util.UUID
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class ValidateMetadataService @Inject() (val ws: WSClient, val configuration: Configuration, graphQLConfiguration: GraphQLConfiguration)(implicit
+    val executionContext: ExecutionContext
+) extends Logging {
+
+  def triggerMetadataValidation(consignmentId: UUID, token: String): Future[Boolean] = {
+    val url = s"${configuration.get[String]("metadatavalidation.baseUrl")}/draft-metadata/validate/$consignmentId"
```
vimleshtna, line 16: Can you use the `ApplicationConfig` to get the `metadatavalidation.baseUrl`? You need to add a variable for that:

  `val metadataValidationBaseUrl: String = configuration.get[String]("metadatavalidation.baseUrl")
`

app/services/ValidateMetadataService.scala
```diff
@@ -0,0 +1,29 @@
...
+    ws.url(url)
+      .addHttpHeaders(("Authorization", token), ("Content-Type", "application/json"))
+      .post("{}")
+      .flatMap(r =>
+        r.status match {
+          case 200 => Future(true)
```
vimleshtna, line 22: Can you use `Future.successful(true)` instead of `Future(true)` because it is much faster and reliable?

https://blog.colinbreck.com/future-successful-blocking-your-future-success/

vimleshtna (commented):
Looks good. Just a few minor comments. Thanks

vimleshtna (approved):
Looks good. Thanks

PR #3763, opened 2024-03-05
TDRD-6 Results of metadata upload page

thanhz:


app/controllers/DraftMetadataChecksResultsController.scala
```diff
@@ -18,7 +21,8 @@ class DraftMetadataChecksResultsController @Inject() (
     val keycloakConfiguration: KeycloakConfiguration,
     val frontEndInfoConfiguration: ApplicationConfig,
```
vimleshtna, line 22: We are not using `frontEndInfoConfiguration` you can remove it from the controller and html page as well

app/controllers/DraftMetadataChecksResultsController.scala
```diff
@@ -27,12 +31,31 @@ class DraftMetadataChecksResultsController @Inject() (
...
+      case FailedValue.value              => failed
+      case CompletedWithIssuesValue.value => DraftMetadataProgress("ERRORS", "red")
+      case CompletedValue.value           => DraftMetadataProgress("IMPORTED", "blue")
+      case _                              => failed
+    } getOrElse failed
+  }
+}
+
+object DraftMetadataChecksResultsController {
+  case class DraftMetadataProgress(value: String, colour: String)
```
vimleshtna, line 60: You can simply create this `DraftMetadataProgress` case class without the `object DraftMetadataChecksResultsController`.

test/controllers/DraftMetadataChecksResultsControllerSpec.scala
```diff
@@ -34,22 +38,6 @@ class DraftMetadataChecksResultsControllerSpec extends FrontEndTestHelper {
   }
 
   "DraftMetadataChecksResultsControllerSpec GET" should {
```
vimleshtna, line 40: I know you didn't write this but can you please change "DraftMetadataChecksResultsControllerSpec GET" to `DraftMetadataChecksResultsController GET`?

vimleshtna (commented):
Looks good. Just a few minor comments. Thanks
Also, was checking the functionality section of the ticket https://national-archives.atlassian.net/browse/TDRD-6. I think the implementation of the last 3 points is missing in this PR. Can you please check them with Tom if you have to implement them?

TomJKing:
> _Also, was checking the functionality section of the ticket https://national-archives.atlassian.net/browse/TDRD-6. I think the implementation of the last 3 points is missing in this PR. Can you please check them with Tom if you have to implement them?_

@vimleshtna / @thanhz I discussed this with Thanh. The prototype designs have changed since I originally wrote the stories. They are now based on the assumption that what the users uploads will be valid, so there would be no need for the user to download a version of the CSV with the errors in it. Should raise with Emma / Gareth whether they want that functionality implemented now. If so can raise another PR with the necessary changes.

PR #3764, opened 2024-03-06
Tdr 2859 file check validation

Tom-Hallett:


app/controllers/ConfirmTransferController.scala
```diff
@@ -49,19 +50,33 @@ class ConfirmTransferController @Inject() (
       implicit request: Request[AnyContent]
   ): Future[Result] = {
     consignmentStatusService.getConsignmentStatuses(consignmentId, request.token.bearerAccessToken).flatMap { consignmentStatuses =>
+      val consignmentStatusValues: Map[Statuses.StatusType, Option[String]] =
+        consignmentStatusService.getStatusValues(consignmentStatuses, SeriesType, TransferAgreementType, UploadType, ClientChecksType)
       val exportTransferStatus = consignmentStatusService.getStatusValues(consignmentStatuses, ExportType).values.headOption.flatten
```
TomJKing, line 55: Do you need to make the call to get the export transfer status separately? The `consignmentStatusService.getStatusValues` is called already. 

Could the export status value be included in this first call, and then retrieved from the resulting map? Would save a call to the DB.

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -549,7 +653,13 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
       forAll(consignmentStatuses) { consignmentStatus =>
         s"render the confirm transfer 'already confirmed' page with an authenticated $userType user if export status is '$consignmentStatus'" in {
           val controller = instantiateConfirmTransferController(getAuthorisedSecurityComponents)
-          val consignmentStatuses = List(ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "Export", consignmentStatus, someDateTime, None))
+          val consignmentStatuses = List(
```
TomJKing, line 656: I'd be tempted to abstract out this list of consignment statuses, as its duplicated code, into something like the below. It would reduce some of the clutter in the tests:

```
private def mockConsignmentStatuses(exportStatus: String): List[ConsignmentStatuses] = {
    List(
      ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "Series", "Completed", someDateTime, None),
      ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "TransferAgreement", "Completed", someDateTime, None),
      ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "Upload", "Completed", someDateTime, None),
      ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "ClientChecks", "Completed", someDateTime, None),
      ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "Export", exportStatus, someDateTime, None)
    )
  }
```

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -590,7 +706,13 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
           s"confirmTransfer after previously submitting an incorrect form and the export status is '$consignmentStatus'" in {
             val controller = instantiateConfirmTransferController(getAuthorisedSecurityComponents)
             val incompleteTransferConfirmationForm = Seq()
-            val consignmentStatuses = List(ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), "Export", consignmentStatus, someDateTime, None))
+            val consignmentStatuses = List(
```
TomJKing, line 709: Duplicated code. See comment for line 656

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -76,6 +78,102 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
...
+
+      val confirmTransferPage = controller
+        .confirmTransfer(consignmentId)
+        .apply(FakeRequest(GET, s"/consignment/$consignmentId/confirm-transfer").withCSRFToken)
+
+      playStatus(confirmTransferPage) mustBe SEE_OTHER
+      redirectLocation(confirmTransferPage).get must equal(s"/consignment/$consignmentId/file-checks?uploadFailed=false")
+    }
+
+    "render the confirm transfer page with an authenticated user" in {
```
TomJKing, line 161: Maybe reword the test slightly to: `render the confirm transfer page with an authenticated user when all relevant statuses are 'Completed'`

Just makes it a little clearer what is meant to happen.

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -76,6 +78,102 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
...
+
+      val confirmTransferPage = controller
+        .confirmTransfer(consignmentId)
+        .apply(FakeRequest(GET, s"/consignment/$consignmentId/confirm-transfer").withCSRFToken)
+
+      playStatus(confirmTransferPage) mustBe SEE_OTHER
+      redirectLocation(confirmTransferPage).get must equal(s"/consignment/$consignmentId/upload")
+    }
+
+    "redirect to the file checks page if the file checks status is not 'Completed'" in {
```
TomJKing, line 136: Maybe change the test title slightly to: `... checks page with an authenticated user if the file ...`

test/controllers/ConfirmTransferControllerSpec.scala
```diff
@@ -76,6 +78,102 @@ class ConfirmTransferControllerSpec extends FrontEndTestHelper {
...
+
+      val confirmTransferPage = controller
+        .confirmTransfer(consignmentId)
+        .apply(FakeRequest(GET, s"/consignment/$consignmentId/confirm-transfer").withCSRFToken)
+
+      playStatus(confirmTransferPage) mustBe SEE_OTHER
+      redirectLocation(confirmTransferPage).get must equal(s"/consignment/$consignmentId/transfer-agreement")
+    }
+
+    "redirect to the upload page if the upload status is not 'Completed'" in {
```
TomJKing, line 112: Maybe change the test title slightly to: `... upload page with an authenticated user if the upload ...`

TomJKing (commented):
Looks good. Just some very minor comments about cosmetic things. Thanks

I think this task might warrant a separate exploratory testing Jira task to test the permutations.

app/controllers/ConfirmTransferController.scala
```diff
@@ -49,19 +50,33 @@ class ConfirmTransferController @Inject() (
...
-      exportTransferStatus match {
-        case Some(InProgressValue.value) | Some(CompletedValue.value) | Some(FailedValue.value) =>
-          consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken).map { consignmentRef =>
-            Ok(views.html.transferAlreadyCompleted(consignmentId, consignmentRef, request.token.name)).uncache()
-          }
-        case None =>
-          getConsignmentSummary(request, consignmentId).map { consignmentSummary =>
-            httpStatus(views.html.standard.confirmTransfer(consignmentId, consignmentSummary, finalTransferForm, request.token.name)).uncache()
+      val incompleteStatuses = consignmentStatusValues.filter(s => !s._2.contains(CompletedValue.value) || s._2.contains(None))
```
annielh, line 56: I reckon the second half of this disjunction is redundant, so we can just have
```
consignmentStatusValues.filter(s => !s._2.contains(CompletedValue.value))
```
`s._2` here is an `Option[String]`, and `contains` will evaluate as true if `s._2` equals `Some` of the argument you have given to `contains` and will evaluate as false if `s._2` ia `None`. It will never contain `None` as we're not dealing with an an `Option[Option[T]`.

I also have a slight personal preference for using `case` syntax on map tuples for readability so you can name things instead of using `._*`, so here something like 
```
      val incompleteStatuses = consignmentStatusValues.filter { case (_, statusValue) => statusValue.contains(CompletedValue.value) }
```
But your call!

annielh (commented):
Nice! One little comment on a redundant disjunct + optional stylistic note.

annielh (approved):

PR #3939, opened 2024-06-05
[TDRD 223] Add holding page for TNA user

annielh:


test/controllers/HomepageControllerSpec.scala
```diff
@@ -92,6 +95,19 @@ class HomepageControllerSpec extends FrontEndTestHelper {
       homepagePageAsString must not include viewTransferButton
     }
 
+    "render the DTA review homepage page with an authenticated tna user" in {
```
TomJKing, line 98: Might want a test here to make sure a user with the incorrect authorisation can't access the TNA user homepage content.

TomJKing (commented):
Looks good.

Something to consider, potentially will have different "types" of `TNAUser`, with different access requirements in the future, for example:

- **DTA Reviewer:** access to the metadata of all consignments from all users;
- **Product Owner:** access to reporting functionality
- **SomeOtherRole:** other access requirements

Would this be modelled in Keycloak with each TNA role have a group with a user assigned to multiple groups, eg:

- DTA Reviewer: `TNAUser` group + `TNADTAReviewer` group
- Product Manager: `TNAUser` group + `TNAProductManager` group

Also, the metadata will be stored in the database at this stage. This will mean the DTA reviewer user will need to request the data via the API, so there needs to be some consideration of the access requirements in the API.

See here how the API authorisation works via `tags`: https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala 

One options is to use the same method as for "reporting" / "export" access via a new Keycloak client. Or potentially check the individual user Token to see if they're a `TNAUser` considering the above point about different levels of access for different types of `TNAUsers`.

test/controllers/HomepageControllerSpec.scala
annielh: I've added lines in existing tests to check that the transfers for review button is not present for judgment or standard users, but will refactor out some sort of content method later on once the page becomes more substantial.

For no authorisation at all, we already seem to have a check to ensure redirects to the auth server from each of the paths we are using (`/judgment/homepage` and `/homepage`, which is currently used for the TNA homepage as well, though we will want to implement a dedicated route for TNA pages as they get added in further down the line).

annielh:
> Looks good.
> 
> Something to consider, potentially will have different "types" of `TNAUser`, with different access requirements in the future, for example:
> 
>     * **DTA Reviewer:** access to the metadata of all consignments from all users;
> 
>     * **Product Owner:** access to reporting functionality
> 
>     * **SomeOtherRole:** other access requirements
> 
> 
> Would this be modelled in Keycloak with each TNA role have a group with a user assigned to multiple groups, eg:
> 
>     * DTA Reviewer: `TNAUser` group + `TNADTAReviewer` group
> 
>     * Product Manager: `TNAUser` group + `TNAProductManager` group
> 
> 
> Also, the metadata will be stored in the database at this stage. This will mean the DTA reviewer user will need to request the data via the API, so there needs to be some consideration of the access requirements in the API.
> 
> See here how the API authorisation works via `tags`: https://github.com/nationalarchives/tdr-consignment-api/blob/master/src/main/scala/uk/gov/nationalarchives/tdr/api/auth/AuthorisationTags.scala
> 
> One options is to use the same method as for "reporting" / "export" access via a new Keycloak client. Or potentially check the individual user Token to see if they're a `TNAUser` considering the above point about different levels of access for different types of `TNAUsers`.

Thanks, this is really useful. Am hoping to flesh this all out as part of `TDRD 300`- just working on a slightly better card description at the moment.

TomJKing (approved):
Looks good. Thanks

Tom-Hallett (approved):
All good

PR #3941, opened 2024-06-05
Tdrd 194 remove all old code when automated judgment accepted 

ian-hoyle:


test/testUtils/TransferMockHelper.scala
```diff
@@ -0,0 +1,74 @@
...
+import io.circe.Printer
+import io.circe.generic.auto._
+import io.circe.syntax._
+import play.api.Application
+import uk.gov.nationalarchives.tdr.GraphQLClient
+
+import java.util.UUID
+import scala.concurrent.ExecutionContext
+
+object TransferMockHelper {
```
vimleshtna, line 17: We already have `FrontEndTestHelper` so you can move these stubs to this class.

vimleshtna (commented):
Looks good. Just a minor comment. Thanks

test/testUtils/TransferMockHelper.scala
ian-hoyle: moved

vimleshtna (approved):
Looks good. Thanks

PR #3972, opened 2024-06-20
TDRD-316 and TDRD-341 Updating tdr-components and fixing Excel bug

Tom-Hallett:


app/controllers/util/ExcelUtils.scala
```diff
@@ -26,7 +26,7 @@ object ExcelUtils {
       }
     }
 
-    ws.range(0, 0, 0, 100).style().bold().set()
+    ws.range(0, 0, 0, rows.head.length).style().bold().set()
```
annielh, line 29: Will this give us an off by one? Ie, will we get one extra column. Not totally sure from looking at the library docs, but worth sanity checking whether we want `.length` or `.length - 1` here.

Tom-Hallett: Well spotted

annielh (approved):

PR #3987, opened 2024-06-25
TDRD-309 - Remove dependency on client side to trigger the export

vimleshtna:


test/controllers/FileChecksControllerSpec.scala
```diff
@@ -99,9 +100,13 @@ class FileChecksControllerSpec extends FrontEndTestHelper with TableDrivenProper
         val filesProcessedWithChecksum = 12
         val filesProcessedWithFFID = 8
         val dataString: String = progressData(filesProcessedWithAntivirus, filesProcessedWithChecksum, filesProcessedWithFFID, allChecksSucceeded = false)
-
         val uploadStatus = List(ConsignmentStatuses(UUID.randomUUID(), UUID.randomUUID(), UploadType.id, CompletedValue.value, someDateTime, None))
-        mockGetFileCheckProgress(dataString, userType)
+        if (userType == "judgment") {
+          val dataString2: String = progressData(40, 40, 40, allChecksSucceeded = true)
```
thanhz, line 105: Can you name the variables so that it easy to know what the numbers mean?
`val dataString2: String = progressData(filesProcessedWithAntivirus = 40, filesProcessedWithChecksum = 40, filesProcessedWithFFID = 40, allChecksSucceeded = true)`

app/controllers/FileChecksController.scala
```diff
@@ -119,16 +121,45 @@ class FileChecksController @Inject() (
         ).uncache()
       } else {
         if (isJudgmentUser) {
-          Ok(views.html.judgment.judgmentFileChecksProgress(consignmentId, reference, applicationConfig.frontEndInfo, request.token.name)).uncache()
+          if (applicationConfig.blockAutomateJudgmentTransfers) {
+            Ok(views.html.judgment.judgmentFileChecksProgress(consignmentId, reference, applicationConfig.frontEndInfo, request.token.name)).uncache()
+          } else {
+            val totalSleepTime = 3 * 60 * 1000 // Total sleep time in milliseconds (5 minutes)
```
thanhz, line 127: Might be unnecessary but would it be useful to have these as an environmental variable?

app/controllers/FileChecksController.scala
```diff
@@ -119,16 +121,45 @@ class FileChecksController @Inject() (
         ).uncache()
       } else {
         if (isJudgmentUser) {
-          Ok(views.html.judgment.judgmentFileChecksProgress(consignmentId, reference, applicationConfig.frontEndInfo, request.token.name)).uncache()
+          if (applicationConfig.blockAutomateJudgmentTransfers) {
```
thanhz, line 124: Should ensure that the [feature block](https://github.com/nationalarchives/tdr-terraform-environments/pull/465) isn't merged or this may break.

app/controllers/FileChecksController.scala
```diff
@@ -119,16 +121,45 @@ class FileChecksController @Inject() (
...
+    val intervals = totalSleepTime / interval
+
+    def fileCheckProgress: Future[Boolean] = consignmentService
+      .fileCheckProgress(consignmentId, request.token.bearerAccessToken)
+      .map(consignment =>
+        consignment.fileChecks.checksumProgress.filesProcessed == consignment.totalFiles &&
+          consignment.fileChecks.ffidProgress.filesProcessed == consignment.totalFiles &&
+          consignment.fileChecks.antivirusProgress.filesProcessed == consignment.totalFiles
+      )
+    breakable {
```
thanhz, line 152: Feel free to ignore this as it may be a bit more unreadable. But to avoid using blocking with `await` this uses chaining with flatMap. It also introduces `future` with `blocking` for the sleep so that the execution context can manage thread pool resources more efficiently.

```    
def loop(remainingIntervals: Int): Future[Unit] = {
      if (remainingIntervals <= 0) {
        Future.unit
      } else {
        fileCheckProgress.flatMap { isCompleted =>
          if (isCompleted) {
            Future.unit
          } else {
            Future {
              blocking(Thread.sleep(interval))
            }.flatMap(_ => loop(remainingIntervals - 1))
          }
        }
      }
    }

loop(intervals)
```

thanhz (commented):
Looks good, just a few comments

app/controllers/FileChecksController.scala
vimleshtna: Yeah, that makes sense. Create a environment variable.

app/controllers/FileChecksController.scala
vimleshtna: Yeah, I have removed this block

app/controllers/FileChecksController.scala
vimleshtna: That's a good approach. I have used this. Thanks

vimleshtna:
> Looks good, just a few comments

@thanhz Just to let you know, the file checks page will still be there for judgment. On the file checks page, It will make a call to the server to continue the transfer, and all the file check progress calls and transfers will be happening on the server side.

app/services/ConsignmentService.scala
```diff
@@ -36,7 +36,6 @@ class ConsignmentService @Inject() (val graphqlConfiguration: GraphQLConfigurati
   private val getConsignmentClient = graphqlConfiguration.getClient[getConsignment.Data, getConsignment.Variables]()
```
thanhz, line 36: unused import `GetFileCheckProgress`

thanhz:
Can you check this doesn't break e2e tests

app/services/ConsignmentService.scala
vimleshtna: Removed it. Thanks

vimleshtna:
> Can you check this doesn't break e2e tests

Yeah, it won't break the frontend as there is no change in the UI

thanhz (approved):
Looks good thanks

PR #4039, opened 2024-07-19
Tdrd 226 dta review page

Tom-Hallett:


app/controllers/MetadataReviewStatusController.scala
```diff
@@ -0,0 +1,63 @@
...
+    val consignmentStatusService: ConsignmentStatusService,
+    val keycloakConfiguration: KeycloakConfiguration,
+    val applicationConfig: ApplicationConfig
+) extends TokenSecurity {
+
+  def metadataReviewStatusPage(consignmentId: UUID): Action[AnyContent] = standardTypeAction(consignmentId) { implicit request: Request[AnyContent] =>
+    if (applicationConfig.blockMetadataReview) {
+      Future(Ok(views.html.notFoundError(name = request.token.name, isLoggedIn = true, isJudgmentUser = false)))
+    } else {
+      for {
```
annielh, line 27: I think we can make this a little tighter by just putting view returning logic in the `yield` rather than constructing artificial `Future`s in the `for` loop:
```      for {
        consignmentStatus <- consignmentStatusService.consignmentStatusSeries(consignmentId, request.token.bearerAccessToken)
        reviewStatus = consignmentStatus.flatMap(s => consignmentStatusService.getStatusValues(s.consignmentStatuses, MetadataReviewType).values.headOption.flatten)
        reference <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
      } yield reviewStatus
        .map(status => Ok(views.html.standard.metadataReviewStatus(consignmentId, reference, request.token.name, request.token.email, status)))
        .getOrElse(Ok(views.html.notFoundError(name = request.token.name, isLoggedIn = true, isJudgmentUser = false)))
    }
```
This does remove the `IllegalStateException` but it looks like you handle an unexpected status in the frontend view logic anyway.

PR #4108, opened 2024-08-22
Add user as well as consignment type check to all transferring body pages

annielh:
Currently, all our transferring body actions on the frontend are wrapped in a check that the consignment in question is of the expected type. This, coupled with access to a consignment only being granted to the user who uploaded it, was previously sufficient to stop unwanted access to other users' consignments by URL.

As stands, however, the new TNA user, who has access to all consignments (currently- this will be restricted slightly), can access by url all transferring body facing pages for a consignment to which they are granted access.

To avoid this we need not only to check that the consignment type being fetched is as expected, but also that the user is of the correct type. This will stop the TNA user having access to judgment and consignment pages.


TomJKing (dismissed):
Looks good. Thanks for doing this. I'm assuming the action will be adjusted accordingly once the "TNA User" permissions are split out into the two existing types.

TomJKing (approved):
Re-approving

TomJKing (approved):
Assuming this is just you re-opening the same PR to get Snyk to work

PR #4146, opened 2024-09-10
TDRD 413 Redirect from draft metadata pages to metadata review status page if consignment is under review

annielh:


thanhz (dismissed):
Looks good to me. I would be inclined to move the `AdditionalMetadataController.redirectIfReviewInProgress(consignmentId, consignmentStatuses)(result)` logic to `controllers/utils/pageRedirectUtils` name can be whatever makes sense or you could leave it where it is as it is all related to the additional metadata anyway.

thanhz (approved):

PR #4151, opened 2024-09-12
Fix download metadata and update status issues for DTA

vimleshtna:


annielh (approved):
Great- thanks for this! LGTM

PR #4205, opened 2024-10-08
Added tests for messages from properties file

ian-hoyle:
Added properties for UTF_8 failure details and action.
Also added tests to check messages in properties files are correct


Tom-Hallett (approved):
Looks fine, thanks

PR #4229, opened 2024-10-16
TDRD-367: reload file checks page

MarkKingsbury:
- Reload page every 5 minutes to ensure credentials do not expire during a large transfer with a long back-end checks process
- Changes to `window.location` in test stop lots of console errors as seen here https://github.com/nationalarchives/tdr-transfer-frontend/actions/runs/11365411499/job/31613540381 (even though all tests pass).
- derived page reload frequency from the token rather than hard-coded so that will stay in sync with any change to keycloak config


TomJKing (approved):
Looks good. Thanks.

Did you speak to Lucien about this in the end?

PR #4271, opened 2024-11-06
unknown error details and action message

ian-hoyle:
Just message text


conf/messages
```diff
@@ -48,3 +48,5 @@ draftMetadata.validation.details.SCHEMA_REQUIRED=There was at least one missing
 draftMetadata.validation.action.SCHEMA_REQUIRED=Add the following column headers to your metadata file and re-load.
 draftMetadata.validation.details.DUPLICATE_HEADER=There was at least one duplicate column in your metadata file.
 draftMetadata.validation.action.DUPLICATE_HEADER=Ensure there is only one of the following duplicated columns.
+draftMetadata.validation.details.UNKNOWN=An unknown error was identified.
+draftMetadata.validation.action.UNKNOWN=Please contact your Digital Transfer Advisor on <a href="mailto:tdr@nationalarchives.gov.uk">tdr@nationalarchives.gov.uk</a> quoting the Consignment Reference.
```
Tom-Hallett, line 52: Can we try replacing this hardcoded email address with parameters (as above on line 28 and others) and then pass in the value using the `@Messages("nationalArchives.email")` value? Will keep things consistent.

Tom-Hallett (changes requested):
Comments about hardcoded email values

conf/messages
ian-hoyle: The action and details messages are retrieved 
`private def actionMessage(fileError: FileError.FileError): String = {
    val key = s"draftMetadata.validation.action.$fileError"
    if (messages.isDefinedAt(key))
      messages(key)
    else
      s"Require action message for $key"
  }`
To pass the nationalArchives.email  value as a parameter the calling code would need to get the value from the properties file first (this is not done in line 28) and ensure parameter position was at 0. The code would need to know if the fileError message needed to pass a particular parameter.  This would make un-generic code

ian-hoyle: Can you give an example of how this could be consistent please

test/controllers/DraftMetadataChecksResultsControllerSpec.scala
```diff
@@ -181,8 +181,8 @@ class DraftMetadataChecksResultsControllerSpec extends FrontEndTestHelper {
       (
         FailedValue.value,
         FileError.UNKNOWN,
-        "Require details message for draftMetadata.validation.details.",
-        "Require action message for draftMetadata.validation.action.",
+        "An unknown error was identified.",
+        "Please contact your Digital Transfer Advisor on <a href=\"mailto:tdr@nationalarchives.gov.uk\">tdr@nationalarchives.gov.uk</a> quoting the Consignment Reference.",
```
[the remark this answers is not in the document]
ian-hoyle: This spec is used to check the message has been correctly retrieved from the properties file so I would usually use a hardcoded string comparison. How would you change this even if there were parameterised message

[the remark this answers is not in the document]
Tom-Hallett: In other spec tests, the messages value is referenced directly - e.g. line 216 on UploadControllerSpec
`"""To transfer larger files, contact <a href="mailto:nationalArchives.email">nationalArchives.email</a>. """`

conf/messages
Tom-Hallett: By consistent I mean that the code is obviously designed to avoid use of hardcoded values and uses the messages file to do this and allow easy maintenance in the future should we ever change the email address we want to use. I can see the issue is not straightforward because there's a conflict between keeping code generic and avoiding hardcoded values - and there is some latitude in the fact that the hardcoding is contained in the messages file rather than directly into a controller or html.

But it still would be possible to use a parameter in the messages and just pass the email through. It would have to work on the assumption that the draftmetadata.validation messages only ever treated parameter 0 as the tdr email which is a constraint - I feel this would be acceptable though because if more parameters or dynamic values are needed for error messages then the approach would have to be reassessed anyway (perhaps using dynamic runtime interpolation with judicious use of string.replace). Message strings that don't have a parameter fail gracefully if a parameter has been passed so that is not an issue.

TomJKing: @ian-hoyle / @Tom-Hallett This was something I briefly discussed with Annie in another PR. The best way around this would to reference one property in the other, like this pseudo code:
```
key1= My name is
key2= ${key1} AName !
```
Though I don't think this is possible. So I would have a dig around to see if it is possible, if not leave it with the hardcoded values.

test/controllers/DraftMetadataChecksResultsControllerSpec.scala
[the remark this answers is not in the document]
ian-hoyle: Isn't this the same  as the test is running
`pageAsString must include("""Please contact your Digital Transfer Advisor on <a href="mailto:tdr@nationalarchives.gov.uk">tdr@nationalarchives.gov.uk</a> quoting the Consignment Reference.""")`

TomJKing (approved):
@ian-hoyle / @Tom-Hallett Approving this. The additional complexity is not worth any benefits that may provide

PR #4287, opened 2024-11-13
TDRD-591 - Set DraftMetadata status to InProgress while uploading draft metadata

vimleshtna:


app/controllers/DraftMetadataUploadController.scala
```diff
@@ -54,8 +57,10 @@ class DraftMetadataUploadController @Inject() (
       val uploadFileName = applicationConfig.draftMetadataFileName
       val uploadKey = s"$consignmentId/$uploadFileName"
       val noDraftMetadataFileUploaded: String = "No meta data file provided"
+      val consignmentStatusInput = ConsignmentStatusInput(consignmentId, DraftMetadataType.id, Some(InProgressValue.value))
 
       def uploadDraftMetadata: IO[Result] = for {
+        _ <- IO.fromFuture(IO(consignmentStatusService.updateConsignmentStatus(consignmentStatusInput, token.bearerAccessToken)))
```
thanhz, line 63: Does the `draftMetadata` status already exist when the user first does a csv upload? If it doesn't doesn't exist will need logic to use the `addConsignmentStatus` and `updateConsignmentStatus` for when it already exists

thanhz (commented):
Looks okay, just a question:

app/controllers/DraftMetadataUploadController.scala
thanhz: Ah seems it created at `AdditionalMetadataEntryMethodController` so should be okay

thanhz (approved):

PR #4292, opened 2024-11-18
[TDRD 87] Row validation error report download compatibility tweaks

annielh:
- Rename Field -> Column as requested by @TNAemmaPea 
- Put row validation errors at the top of error report
- The asset id will now be the file path by default so we can use it directly for the identifier (Filepath) column.

### DIFF
--- a/app/controllers/DraftMetadataChecksResultsController.scala
+++ b/app/controllers/DraftMetadataChecksResultsController.scala
@@ -4,11 +4,10 @@ import auth.TokenSecurity
 import configuration.{ApplicationConfig, KeycloakConfiguration}
 import controllers.util.ExcelUtils
 import controllers.util.MetadataProperty.filePath
-import graphql.codegen.GetConsignmentStatus.getConsignmentStatus.GetConsignment
 import org.pac4j.play.scala.SecurityComponents
 import play.api.i18n.{I18nSupport, Lang, MessagesApi}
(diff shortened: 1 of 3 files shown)


app/controllers/DraftMetadataChecksResultsController.scala
```diff
@@ -129,28 +128,33 @@ class DraftMetadataChecksResultsController @Inject() (
...
-            validationErrors.errors.map(error => List(data(filePath), error.property, data(error.property), error.message))
-          }
-        case _ => Nil
+  def downloadErrorReport(consignmentId: UUID, priorityErrorType: FileError = ROW_VALIDATION): Action[AnyContent] = standardUserAndTypeAction(consignmentId) {
+    implicit request: Request[AnyContent] =>
+      for {
+        reference <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
+        errorReport <- draftMetadataService.getErrorReport(consignmentId)
+      } yield {
+        val orderingPrioritisingRowErrors: Ordering[Error] =
```
TomJKing, line 137: Should rename this val to something less specific given it might not always be row errors with priority

TomJKing (commented):
Sorry just one very small additional comment

TomJKing (approved):
Looks good. Thanks.

PR #4413, opened 2025-02-03
Roll back version upgrade

TomJKing:
TDR seen a large increase in ClodWatch costs related to the 'OTEL' namespace occuring around the time of the update of the versions

Rolling back version to see if this affects the costs


vimleshtna (approved):
Looks good. Thanks

PR #4530, opened 2025-04-01
TRDRD-350/717- Results of your metadata checks content changes

thanhz:
Result page (no error download)
image
Results page with error download
image

### DIFF
--- a/app/views/draftmetadata/draftMetadataChecksWithErrorDownload.scala.html
+++ b/app/views/draftmetadata/draftMetadataChecksWithErrorDownload.scala.html
@@ -8,10 +8,12 @@
         <div class="govuk-grid-column-two-thirds">
             @draftMetadataChecksActionProcess(actionMessage, detailsMessage)
 
-            <p class="govuk-body">The report below contains details about issues found.</p>
+            <p class="govuk-body">The report below contains detailed errors with reference to the file path and column title, which caused the error.</p>
 
             @downloadMetadataLink(consignmentId, "Download report", routes.DraftMetadataChecksResultsController.downloadErrorReport(consignmentId).url)
-            <p class="govuk-body">Once you have addressed this issue upload a revised metadata file.</p>
+            <p class="govuk-body">Once you have addressed the errors, you can re-upload the metadata file.</p>
+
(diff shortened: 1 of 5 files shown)


app/views/partials/daAlert.scala.html
```diff
@@ -0,0 +1,12 @@
+@import java.util.UUID
+@(heading: String, content: Html)
+<div class="da-alert da-alert--default">
```
TomJKing, line 3: Is this an existing component in the design system?

app/views/draftmetadata/draftMetadataChecksWithErrorDownload.scala.html
```diff
@@ -8,10 +8,12 @@
...
             @draftMetadataChecksActionProcess(actionMessage, detailsMessage)
 
-            <p class="govuk-body">The report below contains details about issues found.</p>
+            <p class="govuk-body">The report below contains detailed errors with reference to the file path and column title, which caused the error.</p>
 
             @downloadMetadataLink(consignmentId, "Download report", routes.DraftMetadataChecksResultsController.downloadErrorReport(consignmentId).url)
-            <p class="govuk-body">Once you have addressed this issue upload a revised metadata file.</p>
+            <p class="govuk-body">Once you have addressed the errors, you can re-upload the metadata file.</p>
+
+            @daAlert(heading = "Leaving and returning to this transfer", content = Html(s"You can sign out and return to continue working on this transfer at any time from <a class='govuk-link' href='${routes.ViewTransfersController.viewConsignments(1)}'>View transfers</a>."))
```
TomJKing, line 16: Is that the correct input for the route with the `1`, ie: `... ViewTransfersController.viewConsignments(1)`

TomJKing (commented):
Looks good. Just a couple of minor comments. Thanks

app/views/draftmetadata/draftMetadataChecksWithErrorDownload.scala.html
thanhz: Yep, that would be the first page of the view transfers page. The same is done in the other views that also have a view transfer link.

app/views/partials/daAlert.scala.html
thanhz: I copied it of the prototype. I had a look at the govuk design system but there doesn't seem to be a component like this. Added as a partial as it is used in metadata results error and non-error report pages

TomJKing (approved):
Looks good. Thanks.

PR #4531, opened 2025-04-02
Provide wiz_project_id

thanhz:


TomJKing (commented):
Looks good. The actual secret will need to be added to the repo secrets though. 

For the other "Wiz" secrets these were added as organisational secrets by either Paul S or Kurtis. I think you need to ask them about this.

PR #4533, opened 2025-04-03
TDRD-717 Fix spacing

thanhz:
Fix some spacing to more closely match the prototype 
Screenshot from 2025-04-03 12-13-37
I removed the `govuk-!-margin-bottom-8`  from the partial view of the download button. Not sure why there is this additional margin-bottom, but I had a look at the pages that uses the partial and it still looks okay after removing it.


TomJKing (approved):
Looks good. Thanks

PR #4559, opened 2025-04-16
Tdrd 718 display draft metadata file name

TomJKing:
### Display for Error Page
TDRD-718-filename-error-page

### Display for Success Page
TDRD-718-filename-success-page

### DIFF
--- a/app/controllers/DraftMetadataChecksResultsController.scala
+++ b/app/controllers/DraftMetadataChecksResultsController.scala
@@ -24,7 +24,6 @@ class DraftMetadataChecksResultsController @Inject() (
     val keycloakConfiguration: KeycloakConfiguration,
     val consignmentService: ConsignmentService,
     val applicationConfig: ApplicationConfig,
-    val consignmentStatusService: ConsignmentStatusService,
     val draftMetadataService: DraftMetadataService,
     val messages: MessagesApi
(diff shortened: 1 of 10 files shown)


app/controllers/DraftMetadataChecksResultsController.scala
```diff
@@ -37,17 +36,19 @@ class DraftMetadataChecksResultsController @Inject() (
     } else {
       val token = request.token.bearerAccessToken
       for {
-        reference <- consignmentService.getConsignmentRef(consignmentId, request.token.bearerAccessToken)
-        consignmentStatuses <- consignmentStatusService.getConsignmentStatuses(consignmentId, token)
-        draftMetadataStatus = consignmentStatuses.find(_.statusType == DraftMetadataType.id).map(_.value)
+        consignmentDetails <- consignmentService.getConsignmentDetails(consignmentId, token)
+        reference = consignmentDetails.consignmentReference
+        statuses = consignmentDetails.consignmentStatuses
+        uploadedFileName = consignmentDetails.clientSideDraftMetadataFileName.getOrElse("Not Available")
```
thanhz, line 42: The call is unlikely to fail but if it did "Not Available" will be displayed in the UI which might be confusing to the user. Might be best not to show it at all but that's another discussion.

PR #4581, opened 2025-04-24
TDRD-561 View transfer redirects for draftMetadata statuses

thanhz:


app/controllers/ViewTransfersController.scala
```diff
@@ -88,6 +88,10 @@ class ViewTransfersController @Inject() (
       case s if s.containsStatuses(ExportType) => toExportAction(s.find(_.statusType == ExportType.id).get, judgmentType, consignmentId, consignmentRef)
       case s if s.statusValue(ConfirmTransferType).contains(CompletedValue.value) =>
         UserAction(InProgress.value, routes.TransferCompleteController.transferComplete(consignmentId).url, Resume.value)
+      case s if s.statusValue(DraftMetadataType).contains(CompletedValue.value) =>
```
annielh, line 91: Do we not want these cases below the metadata review case?

thanhz: Yeah it should

annielh (approved):

PR #4693, opened 2025-06-13
TDRD-795 Remove unused customMetadata methods

thanhz:


TomJKing (commented):
Looks good. Thanks

I know not really related to this change but could you:

- add a unit test to check the string values of the property `val`s in the `MetadataProperty` object
- remove the unused properties

PR #4734, opened 2025-07-01
TDRD-726 Notification banner for upload pages

thanhz:
**Standard**
image

**Judgment**
image


app/views/standard/upload.scala.html
```diff
@@ -126,7 +126,18 @@ <h2 class="govuk-error-summary__title" id="progress-error-summary-title">
...
-            <p class="govuk-body">Do not close your browser while your records are being uploaded. This may take a few minutes.</p>
+            <div class="govuk-notification-banner govuk-!-margin-bottom-4" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
+                <div class="govuk-notification-banner__header">
+                    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+                        Important
+                    </h2>
+                </div>
+                <div class="govuk-notification-banner__content">
+                    <p class="govuk-notification-banner__heading">
+                        Do not close your browser window while your file is being uploaded. This may take a few minutes.
```
TomJKing, line 137: Should this not be: `... while your file(s) are being ...`? Or something like that, as standard transfers contain can contain 1+ files.

TomJKing (commented):
Looks good. Just one minor comment around the content.

app/views/standard/upload.scala.html
thanhz: Spoke with Shakir and he suggested change the text to pluralise without brackets. so `while your files are being'....`

TomJKing (approved):
Looks good. Thanks. 

Keep an eye on the e2e tests to make sure it doesn't break them.

PR #4803, opened 2025-07-28
delay release restirctions on use download

ian-hoyle:
restrictions on use not in metadata download


TomJKing (approved):
Looks good. Thanks.

I think the E2E tests will also need updating as well?

PR #4813, opened 2025-07-30
Tdrd 863 sorting on metadata download template

vimleshtna:


Tom-Hallett (changes requested):
Needs requirement review because Excel doesn't appear to do a natural sort (but possible that Windows File Explorer does) so need to determine what exact sorting options are.

Tom-Hallett (approved):
All looks fine. No issues with code but suggest documenting idiosyncrasies of Excel such as characters 60 - 62 (`<`, `=` and `>` ) being sorted after other special characters for no obvious reason. Also need to be aware that Excel doesn't like certain characters at the start of a cell (=, @) but agree this is unlikely to occur in metadata.

PR #4841, opened 2025-08-12
update dependencies

thanhz:


build.sbt
```diff
@@ -14,7 +14,7 @@ libraryDependencies += guice
 libraryDependencies += "org.scalatestplus.play" %% "scalatestplus-play" % "7.0.2" % Test
 
 val playVersion = "3.0.3"
-val playPac4jVersion = "12.0.2-PLAY3.0"
+val playPac4jVersion = "13.0.0-PLAY3.0"
```
TomJKing, line 17: As this is a major version upgrade, might be safer to not include it in this PR.

TomJKing (commented):

PR #5105, opened 2026-04-23
correct date

ian-hoyle:
Display time in local time. 
Remove preceding 0 from time  8:30am not 08:30am


TomJKing (approved):
Looks good. Thanks

PR #5110, opened 2026-04-29
TDRD-1314: auth failures in file checks recover with reload

MarkKingsbury:
- remove page refresh interval approach (incl update of tests)
- use a page refresh to recover from a fail in file-checks-progress (wait 10 secs first to mitigate any failure loops)


TomJKing (approved):
Looks good. Thanks

PR #5113, opened 2026-04-30
allow original metadata csv download as excel

ian-hoyle:
### Summary

Adds a new endpoint that allows standard and TNA users to download the original draft metadata CSV file (from S3) converted to Excel format. This supports the workflow where users need to retrieve their submitted metadata for transfers that have been rejected.

### Changes

- **New controller action** `downloadCsvMetadataFile` in `DownloadMetadataController` — fetches the draft metadata CSV from S3 via `DownloadService`, converts it to Excel using a new utility method, and serves it as a `.xlsx` download.
- **New route** `GET /consignment/:id/draft-metadata/download-metadata/csv-as-excel`
- **New utility** `ExcelUtils.convertCsvToExcel` — parses a CSV `Source` using `scala-csv` and writes it to an Excel byte array via the existing `writeExcel` helper.
- **New dependency** `com.github.tototoshi:scala-csv:2.0.0` for robust CSV parsing (handles quoted fields, commas in values, etc.)
- **Tests** — updated `DownloadMetadataControllerSpec` to accommodate new constructor dependencies; added `ExcelUtilsSpec` with coverage for content correctness, custom sheet names, and quoted-field handling.

### Access control

The endpoint uses `tnaUserAction`, so it is accessible to  TNA admin users. TNA user access is audit-logged.


TomJKing (approved):
Looks good. Thanks

PR #5117, opened 2026-05-07
TDRD-1335 Create MetadataReviewHistory page + link to metadata details page

thanhz:


app/controllers/util/DateUtils.scala
```diff
@@ -12,6 +12,14 @@ object DateUtils {
     dateTime.withZoneSameInstant(zoneId).format(formatter)
   }
 
+  def formatWithDaySuffix(dateTime: ZonedDateTime, zoneId: ZoneId = ukTimeZone): String = {
```
ian-hoyle, line 15: can allow this to be also used in def below

TomJKing, line 15: These should all be abstracted to the tdr-common-utils as a "date library". Could you raise a ticket in the backlog for this.

TomJKing (approved):
Looks good. Just one request, which doesn't impact approving this PR. Thanks

app/controllers/util/DateUtils.scala
thanhz: Will create a ticket to extract this out to the tdr-common-utils as a date library
https://national-archives.atlassian.net/browse/TDRD-1465

PR #5138, opened 2026-06-03
Update npm dependency tdr-generated-graphql and fix build

vimleshtna:
`@aws-sdk/lib-storage` pulls in `stream-browserify`, which requires the Node.js `buffer` core module. Webpack 5 no longer auto-polyfills Node core modules.


TomJKing (approved):
Looks good. Thanks

PR #5145, opened 2026-06-15
TDRD-1557 Add skip draft metadata endpoint

thanhz:
Need to create a new endpoint so that we can update the draftmetadata status

### DIFF
--- a/app/controllers/PrepareMetadataController.scala
+++ b/app/controllers/PrepareMetadataController.scala
@@ -2,10 +2,12 @@ package controllers
 
 import auth.TokenSecurity
 import configuration.KeycloakConfiguration
+import graphql.codegen.types.ConsignmentStatusInput
 import org.pac4j.play.scala.SecurityComponents
 import play.api.mvc.{Action, AnyContent, Request}
-import services.Statuses.{DraftMetadataType, InProgressValue}
 import services.{ConsignmentService, ConsignmentStatusService}
+import uk.gov.nationalarchives.tdr.common.utils.statuses.StatusTypes.DraftMetadataType
+import uk.gov.nationalarchives.tdr.common.utils.statuses.StatusValues._
 
 import java.util.UUID
 import javax.inject.Inject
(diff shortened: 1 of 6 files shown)


TomJKing (commented):
Looks good. Thanks

Should update the `ViewTransfersController.scala` to take account of the new "Skipped" status for draft metadata. See here: https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/controllers/ViewTransfersController.scala#L116

I would confirm with Emma about where the user should be taken when they've skipped the metadata upload. Most like the submit for metadata review page?

thanhz:
> Looks good. Thanks
> 
> Should update the `ViewTransfersController.scala` to take account of the new "Skipped" status for draft metadata. See here: https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/controllers/ViewTransfersController.scala#L116
> 
> I would confirm with Emma about where the user should be taken when they've skipped the metadata upload. Most like the submit for metadata review page?

The existing behaviour is that it takes the user to the `download your metadata page` which is the page just before submitting metadata for review page. I think this makes sense as the latter page doens't have a download metadata button but I can still ask Emma.

TomJKing:
> The existing behaviour is that it takes the user to the download your metadata page which is the page just before submitting metadata for review page. I think this makes sense as the latter page doens't have a download metadata button but I can still ask Emma.

@thanhz That makes sense to me. Probably best to confirm with Emma though. Also need to run formatting as the build is failing.

TomJKing (approved):
Looks good. Thanks

thanhz:
> > The existing behaviour is that it takes the user to the download your metadata page which is the page just before submitting metadata for review page. I think this makes sense as the latter page doens't have a download metadata button but I can still ask Emma.
> 
> @thanhz That makes sense to me. Probably best to confirm with Emma though. Also need to run formatting as the build is failing.

Emma said that it's fine to leave it as it is currently

PR #5151, opened 2026-06-24
Tdrd 1251 file checks step function failure

ian-hoyle:


app/views/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,42 @@
+@import views.html.partials.transferReference
+
+@(name: String, title: String, consignmentRef: String)(implicit request: RequestHeader, messages: Messages)
+
+@main(title, name = name, isJudgmentUser = false) {
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-notification-banner file-checks-backend-failure-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
```
vimleshtna: Can you re-use the existing https://github.com/nationalarchives/tdr-transfer-frontend/blob/master/app/views/partials/notificationBannerError.scala.html?

You need to add an argument in `notificationBannerError` for the details.
`@(title: String, heading: String, details: String = "")`

So, if the `details` is not empty, then add a `<p>` tag for the details.

app/views/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,42 @@
+@import views.html.partials.transferReference
+
+@(name: String, title: String, consignmentRef: String)(implicit request: RequestHeader, messages: Messages)
+
+@main(title, name = name, isJudgmentUser = false) {
```
vimleshtna: Redundant default argument - `isJudgmentUser`

Also, could you move this file to - https://github.com/nationalarchives/tdr-transfer-frontend/tree/master/app/views/standard/filechecks folder

app/views/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,42 @@
...
+          <p class="govuk-body">There may be a technical issue with this service.</p>
+        </div>
+      </div>
+
+      <h1 class="govuk-heading-l">What to do next</h1>
+
+      <p class="govuk-body">
+        Contact your transfer advisor at
+        <a class="govuk-link" href="mailto:tdr@@nationalarchives.gov.uk">tdr@@nationalarchives.gov.uk</a>
+        with your consignment reference <strong>@consignmentRef</strong> and they will help you resolve the issue.
```
vimleshtna: You should use `@Messages("nationalArchives.email")` for email.

app/views/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,42 @@
...
+      </p>
+
+      <hr class="govuk-section-break govuk-section-break--m">
+
+      <div class="da-alert da-alert--default">
```
vimleshtna: Can you reuse the `daAlert` template for the alert?

app/views/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,42 @@
...
+          <p class="govuk-body">Once your transfer advisor confirms that the issues are resolved, you will need to start a new transfer.</p>
+        </div>
+      </div>
+    </div>
+
+    @transferReference(consignmentRef, isJudgmentUser = false)
+  </div>
+}
+
+
```
vimleshtna: Unnecessary 2 line breaks.

app/controllers/FileChecksController.scala
```diff
@@ -72,10 +74,15 @@ class FileChecksController @Inject() (
...
 
   def fileCheckProgress(consignmentId: UUID): Action[AnyContent] = secureAction.async { implicit request =>
-    consignmentService
-      .fileCheckProgress(consignmentId, request.token.bearerAccessToken)
-      .map(_.asJson.noSpaces)
-      .map(Ok(_))
+    val token = request.token.bearerAccessToken
+    for {
+      progress <- consignmentService.fileCheckProgress(consignmentId, token)
+      statuses <- consignmentStatusService.getConsignmentStatuses(consignmentId, token)
```
vimleshtna: We're calling this service from the JavaScript every 10 seconds, and with this change, it will make two Consignment API calls, which could impact performance.

Could you update the `getFileCheckProgress` query to include `consignmentStatuses` as well? That would avoid the need for multiple API calls.

app/controllers/FileChecksResultsController.scala
```diff
@@ -40,11 +40,17 @@ class FileChecksResultsController @Inject() (
     val pageTitle = "Results of your checks"
 
     for {
+      consignmentStatuses <- consignmentStatusService.getConsignmentStatuses(consignmentId, request.token.bearerAccessToken)
+      clientChecksStatus = consignmentStatusService.getStatusValues(consignmentStatuses, ClientChecksType).values.headOption.flatten.getOrElse("")
+      serverFFIDStatus = consignmentStatusService.getStatusValues(consignmentStatuses, ServerFFIDType).values.headOption.flatten.getOrElse("")
+      backendChecksFailed = clientChecksStatus == FailedValue.value && serverFFIDStatus == InProgressValue.value
       fileCheck <- consignmentService.getConsignmentFileChecks(consignmentId, request.token.bearerAccessToken)
```
vimleshtna: Same as above related to multiple API calls.

Could you update the `getFileCheckProgress` query to include consignmentStatuses as well? That would avoid the need for multiple API calls.

Also, add the `consignmentRef` field as well in the `getFileCheckProgress` query.

npm/css-src/sass/main.scss
```diff
@@ -102,3 +102,16 @@ address span {
...
+  border-color: #ca3535;
+}
+
+.file-checks-backend-failure-banner .govuk-notification-banner__header {
+  background: #ca3535;
+}
+
+.file-checks-backend-failure-banner .govuk-notification-banner__title {
+  color: #fff;
+}
```
vimleshtna: You don't need to add the above CSS if we reuse the `notificationBannerError.scala.html` template.

vimleshtna (commented):
Looks good. Some minor comments and suggestions. Thanks

app/views/partials/notificationBannerError.scala.html
```diff
@@ -9,5 +9,6 @@ <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title
         <h3 class="govuk-notification-banner__heading">
             @heading
         </h3>
+        @contentDetails.map { content => @content }
```
vimleshtna: If you just write `@contentDetails`, it still works.

app/views/standard/filechecks/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,27 @@
+@import views.html.partials.transferReference
+@import views.html.partials.notificationBannerError
+@import views.html.partials.daAlert
+
+@(name: String, title: String, consignmentRef: String)(implicit request: RequestHeader, messages: Messages)
+
+@main(title, name = name, isJudgmentUser = false) {
```
vimleshtna: Redundant default argument - `isJudgmentUser = false`

app/views/standard/filechecks/fileChecksBackendFailure.scala.html
```diff
@@ -0,0 +1,27 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      @notificationBannerError("File check error", "We have a problem checking your records", Some(Html("""<p class="govuk-body">There may be a technical issue with this service.</p>""")))
+
+      <h1 class="govuk-heading-l">What to do next</h1>
+
+      <p class="govuk-body">
+        Contact your transfer advisor at
+        <a class="govuk-link" href="mailto:tdr@@nationalarchives.gov.uk">tdr@@nationalarchives.gov.uk</a>
+        with your consignment reference <strong>@consignmentRef</strong> and they will help you resolve the issue.
```
vimleshtna, line 17: Could you replace the email with `@Messages("nationalArchives.email")`?

vimleshtna (commented):
Looks good. Just some more minor comments. Thanks

vimleshtna (approved):
Looks good. Thanks.

PR #5153, opened 2026-06-25
TNASupport page

ian-hoyle:


app/controllers/FileChecksResultsController.scala
```diff
@@ -34,38 +36,63 @@ class FileChecksResultsController @Inject() (
...
+              )
+            )
+          }
+        } else {
+          val fileStatusList = fileCheck.files.flatMap(_.fileStatus)
+          val failedResult = Ok(views.html.fileChecksResultsFailed(request.token.name, pageTitle, reference, isJudgmentUser = false, fileStatusList))
+          if (applicationConfig.blockFileChecksFailureV2) {
+            Future.successful(failedResult)
+          } else {
+            fileCheckFailureService.getFileCheckFailureStatusActionTypes(consignmentId).map { statusActions =>
```
thanhz: I feel that it will be more efficient to update the getConsignmentFileChecks graphql to use 'fileStatuses' since we are already calling it above anyway.
https://github.com/nationalarchives/tdr-generated-graphql/blob/master/src/main/graphql/GetFileCheckProgress.graphql#L7
like this:
```      
      files {
            fileStatuses {
                fileId
                statusType
                statusValue
            }
        }
```

then convert that status to an action like this [convertStatusToAction](https://github.com/nationalarchives/tdr-statuses/blob/main/src/main/scala/uk/gov/nationalarchives/services/FileCheckStatusEvaluator.scala#L25) That way we don't need to call s3 and process every object which is expensive. This will affect @vimleshtna work also, so it might be worth liaising with him to avoid conflicts if you decide to go with this approach

thanhz (commented):
This will work, but question about an alternative way to get the correct action

thanhz (approved):
Looks good, thanks for making those changes. Probably want to double check the E2E tests don't break

PR #5154, opened 2026-06-25
TDRD-1435 - Download File Check Failure Excel

thanhz:
Screenshot 2026-06-25 at 15 15 38


ian-hoyle (changes requested):
can you add tests for the content inside the Excel -  DownloadMetadataControllerSpec   has code to do this

ian-hoyle (approved):
looks good thanks

PR #5157, opened 2026-07-07
TDRD-1390 - Update help and faq guides

vimleshtna:


public/images/standard-checking-records.png
TomJKing, line 1: Looks like there is some grey space on the new image, below the web page. Should probably remove that so it is consistent with the other screenshots.

app/views/faq.scala.html
```diff
@@ -361,15 +361,10 @@ <h3 id="progress-checks" class="govuk-heading-s">What checks are carried out?</h
...
+                    Follow the on-screen instructions if an error occurs. Your Transfer Adviser will be notified and can assist with any file issues that cannot be resolved independently.
                 </p>
-                <ul class="govuk-list govuk-list--bullet">
-                    <li>Your internet connection is stable and strong</li>
-                    <li>That your transfer does not contain records we cannot accept.</li>
-                </ul>
                 <p class="govuk-body">
-                    If you can confirm the above then try again. If your transfer has still failed, contact <a href="mailto:@Messages("nationalArchives.email")">
-                @Messages("nationalArchives.email")</a>.
+                    You can contact <a href="mailto:@Messages("nationalArchives.email")">@Messages("nationalArchives.email")</a> if you have any problems with your upload. Quoting your consignment number in your email will help.
```
TomJKing: Should this not say: `... Quoting your consignment reference in your ...`

Not sure how the "consignment reference" is referred to across the user guides, but that's how its referred to on the web pages. Something to check with the business.

TomJKing (commented):

app/views/faq.scala.html
vimleshtna: Good catch. I confirmed it with Shakir and have updated it. Thanks!

TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):
Looks good. Thanks

PR #5159, opened 2026-07-08
TDRD-1662 FCF excel tweaks

thanhz:
Spoke with Emma, and it was suggested that we hide the StatusType & StatusValue columns as they aren't useful to users but can be useful for debugging purposes.


app/controllers/util/ExcelUtils.scala
```diff
@@ -94,7 +94,7 @@ object ExcelUtils {
 
   }
 
-  def writeExcel(worksheetName: String, rows: List[List[String]]): Array[Byte] = {
+  def writeExcel(worksheetName: String, rows: List[List[String]], hiddenColumns: Set[Int]): Array[Byte] = {
```
ian-hoyle, line 100: can this have a default Set.empty so don't need to change existing callers

ian-hoyle (commented):
looks good 
one minor thing about default and do yo need to test the hidden rows

app/controllers/util/ExcelUtils.scala
thanhz: The problem was that Scala doesn't allow default arguments on multiple overloads. What I've done instead is to add another writeexcel that calls the old one so existing callers need no changes. Also added unit tests

ian-hoyle (approved):

PR #5163, opened 2026-07-15
Long file path fix

ian-hoyle:


TomJKing (approved):
Looks good, as far as my knowledge of Typescript goes.

Concern around the impact on performance on the upload for larger numbers of files. Also with memory issues when running this code on a user machine. Will need to test this out.

Should let Mark K know when this gets to staging as could impact the performance tests.

PR #5165, opened 2026-07-16
Use filesystem api

ian-hoyle:


npm/src/upload/form/upload-form.ts
```diff
@@ -164,23 +166,60 @@ export class UploadForm {
...
-      if (isFile(fileWithPath)) {
-        const fileName = fileWithPath.file.name
-        const checkOrError = this.checkForCorrectJudgmentFileExtension(fileName)
-        if (!isError(checkOrError)) {
-          addFileSelectionSuccessMessage(fileName)
-        } else {
-          return checkOrError
+    if (!this.isJudgmentUser && supportsDirectoryPicker()) {
+      try {
+        const dirHandle = await (window as any).showDirectoryPicker()
```
TomJKing, line 174: As noted this API is not supported on Firefox or Safari (https://developer.mozilla.org/en-US/docs/Web/API/Window/showDirectoryPicker#browser_compatibility)

Could you add a comment about this and state will look into Firefox users in a separate ticket.

TomJKing (commented):
Looks good, as far as my knowledge of Typescript goes. Just one minor comment. Thanks

npm/src/upload/form/upload-form.ts
ian-hoyle: added comment

TomJKing (approved):
Looks good. Thanks

PR #5166, opened 2026-07-17
readable folder but not fil fix

ian-hoyle:


TomJKing (approved):
Looks OK, as far as my understanding of Typescript goes. Thanks

PR #5167, opened 2026-07-17
Sort the metadata review log by eventTime before loading actions

thanhz:
Need to sort the metadataReviewLog by eventTime like how the metadataHistory page is done, in order to show the correct actions for TNA users.


ian-hoyle (approved):
looks good to me

PR #5170, opened 2026-07-20
Revert "Revert file path check feature (temporary - reapply after eme…

ian-hoyle:
…rgency fix) (#5169)"

This reverts commit 2b705f2c87e168edd5d506d915e6118f4912aaaf.

### DIFF
--- a/npm/src/upload/form/file-types.ts
+++ b/npm/src/upload/form/file-types.ts
@@ -1,0 +1,29 @@
+import { IFileWithPath } from "@nationalarchives/file-information"
+
+export type IEntryWithPath = IFileWithPath | IDirectoryWithPath
+export interface IDirectoryWithPath {
+  path: string
+  unreadable?: boolean
+}
+
+export function isFile(entry: IEntryWithPath): entry is IFileWithPath {
+  return (entry as IFileWithPath).file !== undefined
+}
+
+export function isDirectory(
+  entry: IEntryWithPath
+): entry is IDirectoryWithPath {
+  return !isFile(entry)
+}
+
+export function withTimeout<T>(
+  promise: Promise<T>,
+  ms: number,
+  msg: string
+): Promise<T> {
+  let timer: ReturnType<typeof setTimeout>
+  const timeout = new Promise<T>((_, reject) => {
+    timer = setTimeout(() => reject(new Error(msg)), ms)
+  })
+  return Promise.race([promise, timeout]).finally(() => clearTimeout(timer))
+}
--- a/npm/src/upload/form/get-files-from-directory-picker.ts
+++ b/npm/src/upload/form/get-files-from-directory-picker.ts
@@ -1,0 +1,55 @@
+import { IEntryWithPath, withTimeout } from "./file-types"
+
+const READ_ENTRIES_TIMEOUT_MS = 5000
+
+interface IFileSystemFileHandle {
+  kind: "file"
+  name: string
+  getFile: () => Promise<File>
+}
+
+interface IFileSystemDirectoryHandle {
+  kind: "directory"
+  name: string
+  entries: () => AsyncIterableIterator<[string, IFileSystemHandle]>
+}
+
+type IFileSystemHandle = IFileSystemFileHandle | IFileSystemDirectoryHandle
+
+export async function getAllFilesFromHandle(
+  dirHandle: IFileSystemDirectoryHandle,
+  pathPrefix: string
+): Promise<IEntryWithPath[]> {
+  const fileInfos: IEntryWithPath[] = []
+  for await (const [name, handle] of dirHandle.entries()) {
+    const fullPath = pathPrefix + "/" + name
+    if (handle.kind === "directory") {
+      const children = await getAllFilesFromHandle(handle, fullPath).catch(
+        (): null => null
+      )
+      if (children === null) {
+        fileInfos.push({ path: fullPath, unreadable: true })
+      } else if (children.length === 0) {
+        fileInfos.push({ path: fullPath })
+      } else {
+        fileInfos.push(...children)
+      }
+    } else {
+      try {
+        const file = await withTimeout(
+          handle.getFile(),
+          READ_ENTRIES_TIMEOUT_MS,
+          `getFile timed out for: ${fullPath}`
+        )
+        fileInfos.push({ file, path: fullPath })
+      } catch {
+        fileInfos.push({ path: fullPath, unreadable: true })
+      }
+    }
+  }
+  return fileInfos
+}
+
+export function supportsDirectoryPicker(): boolean {
+  return "showDirectoryPicker" in window
+}
--- a/npm/src/upload/form/get-files-from-drag-event.ts
+++ b/npm/src/upload/form/get-files-from-drag-event.ts
@@ -1,46 +1,70 @@
 import { IFileWithPath } from "@nationalarchives/file-information"
+import { IEntryWithPath, withTimeout } from "./file-types"
 
-export type IEntryWithPath = IFileWithPath | IDirectoryWithPath
-export type IDirectoryWithPath = Pick<IFileWithPath, "path">
-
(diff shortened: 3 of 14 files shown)


TomJKing (approved):
Looks good. Thanks

npm/src/upload/form/file-types.ts
```diff
@@ -0,0 +1,29 @@
+import { IFileWithPath } from "@nationalarchives/file-information"
+
+export type IEntryWithPath = IFileWithPath | IDirectoryWithPath
+export interface IDirectoryWithPath {
+  path: string
+  unreadable?: boolean
+}
+
+export function isFile(entry: IEntryWithPath): entry is IFileWithPath {
+  return (entry as IFileWithPath).file !== undefined
```
jlhdxw, line 10: You shouldn't need to cast entry to IFileWithPath when you've already said entry is IEntryWithPath in the param type declaration.

npm/src/upload/form/file-types.ts
```diff
@@ -0,0 +1,29 @@
+import { IFileWithPath } from "@nationalarchives/file-information"
+
+export type IEntryWithPath = IFileWithPath | IDirectoryWithPath
+export interface IDirectoryWithPath {
+  path: string
+  unreadable?: boolean
+}
+
+export function isFile(entry: IEntryWithPath): entry is IFileWithPath {
```
jlhdxw, line 9: This type predicate isn't very safe and doesn't seem to be actually checking that `entry` is a `IFileWithPath` - it's checking entry.file is not undefined, which means anything that is passed into this function that happens to have a `file` attribute that isn't undefined will be typed as a `IFileWithPath`

npm/src/upload/form/file-types.ts
```diff
@@ -0,0 +1,29 @@
...
+  return (entry as IFileWithPath).file !== undefined
+}
+
+export function isDirectory(
```
jlhdxw, line 13: Same comment to this type predicate - it's not actually checking it's a `IDirectoryWithPath`, just that it's not a file. So you could pass in `undefined` as entry into this function and entry would be typed as `IDirectoryWithPath`

npm/src/upload/form/get-files-from-directory-picker.ts
```diff
@@ -0,0 +1,55 @@
...
+const READ_ENTRIES_TIMEOUT_MS = 5000
+
+interface IFileSystemFileHandle {
+  kind: "file"
+  name: string
+  getFile: () => Promise<File>
+}
+
+interface IFileSystemDirectoryHandle {
+  kind: "directory"
```
jlhdxw, line 12: It might be worth making these an enum so when you want to find what kind of handle it is, you have a single reference to the "kind"

npm/src/upload/form/get-files-from-directory-picker.ts
```diff
@@ -0,0 +1,55 @@
...
+        const file = await withTimeout(
+          handle.getFile(),
+          READ_ENTRIES_TIMEOUT_MS,
+          `getFile timed out for: ${fullPath}`
+        )
+        fileInfos.push({ file, path: fullPath })
+      } catch {
+        fileInfos.push({ path: fullPath, unreadable: true })
+      }
+    }
```
jlhdxw, line 48: Might be worth splitting some of this out into separate functions for better readability as this nested conditional statements it's hard to follow

npm/src/upload/form/get-files-from-drag-event.ts
```diff
@@ -1,46 +1,70 @@
...
+      READ_ENTRIES_TIMEOUT_MS,
+      `readEntries timed out for: ${dirPath}`
+    )
 
-  let nextBatch = await getEntryBatch(reader)
-
-  while (nextBatch.length > 0) {
-    allEntries = allEntries.concat(nextBatch)
-    nextBatch = await getEntryBatch(reader)
+    while (nextBatch.length > 0) {
```
jlhdxw, line 58: While loops can go on forever - it might be safer to add a boundary and do this in a for loop (i.e. set a max number of batches you would ever expect and run batches until either there are no batches or the max loop is hit.

npm/src/upload/form/file-types.ts
```diff
@@ -0,0 +1,29 @@
...
+  entry: IEntryWithPath
+): entry is IDirectoryWithPath {
+  return !isFile(entry)
+}
+
+export function withTimeout<T>(
```
jlhdxw, line 19: We have this function defined twice - can it not be shared between the 2 places this is used?

npm/src/upload/form/long-path-check.ts
```diff
@@ -0,0 +1,81 @@
...
+        })
+      } else {
+        results.push({ path: dir.path, status: "ok" })
+      }
+    }
+  }
+  return results
+}
+
+export function hasLongPathIssues(results: IFileCheckResult[]): boolean {
```
jlhdxw, line 75: It might be safer to name this function `hasPathIssues` as it doesn't necessarily check there are specifically long path issues. If the code that parses the results to check for issues adds a new check which adds a non-ok status, this would still return those as "long path" issues

npm/src/upload/form/upload-form.ts
```diff
@@ -159,23 +163,63 @@ export class UploadForm {
...
-        if (!isError(checkOrError)) {
-          addFileSelectionSuccessMessage(fileName)
-        } else {
-          return checkOrError
+    // Firefox and Safari do not support the directory picker API, so we need to check if it is supported before calling it.
+    // Can still upload on these browsers but long path check will not be executed
+    // TDRD-1698
+    if (!this.isJudgmentUser && supportsDirectoryPicker()) {
+      try {
+        const dirHandle = await (window as any).showDirectoryPicker()
```
jlhdxw, line 171: How come window is being cast to `any` here? If the typescript compiler you have doesn't have window as a type, you can add it either manually with a declare, or I think there's some other way you can add it via some setting

npm/src/upload/form/upload-form.ts
```diff
@@ -159,23 +163,63 @@ export class UploadForm {
...
+        if (isError(folderCheck)) {
+          return folderCheck
         }
+        this.selectedFiles = filesAndFolders
+        const parentFolder = this.getParentFolderName(this.selectedFiles)
+        addFolderSelectionSuccessMessage(
+          parentFolder,
+          this.selectedFiles.filter(isFile).length
+        )
+      } catch (err: any) {
```
jlhdxw, line 187: Typing `error` values from `catches` should be `unknown` : https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables

npm/test/get-files-from-directory-picker.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        next() {
+          if (index < children.length) {
+            return Promise.resolve({
+              value: children[index++],
+              done: false
+            })
+          }
+          return Promise.resolve({ value: undefined, done: true })
+        }
+      } as AsyncIterableIterator<[string, any]>
```
jlhdxw, line 36: How come we need to cast this here? It would probably be better to use a type annotation rather than casting.

jlhdxw: The same comment to all the places we're casting in the same way in this file.

npm/test/upload-form-folder.test.ts
```diff
@@ -20,6 +20,11 @@ import { displaySelectionSuccessMessage } from "../src/upload/form/update-and-di
 
 jest.mock("uuid", () => "eb7b7961-395d-4b4c-afc6-9ebcadaf0150")
 
+beforeAll(() => {
+  // stop console errors from window.location navigation in JSDOM
+  console.error = jest.fn()
+})
```
jlhdxw, line 26: I'm not sure if you need to, but do you need to reset this after with an `afterAll` call?

PR #5174, opened 2026-07-23
Lfp review

ian-hoyle:
### Summary

Replaces the previous duck-typing approach (`isFile` checked for the presence of a `file` property) with a proper discriminated union using an `EntryKind` enum. This makes file vs directory distinction explicit and type-safe throughout the upload pipeline.

### Changes

**Type system refactor (`file-types.ts`)**
- Added `EntryKind` enum (`File` | `Directory`)
- Replaced `IEntryWithPath` / `IDirectoryWithPath` with `IEntry` = `IFileEntry | IDirectoryEntry`, each carrying a `kind` discriminant
- `isFile()` and `isDirectory()` are now proper type guards on the `kind` field

**Directory picker refactor (`get-files-from-directory-picker.ts`)**
- Exported `IFileSystemFileHandle` and `IFileSystemDirectoryHandle` interfaces for testability
- Extracted `handleDirectoryEntry()` and `handleFileEntry()` helpers for readability
- All entries now include `kind` when pushed to results

**Upload form logic (`upload-form.ts`)**
- Fixed branching so non-judgment users on browsers without `showDirectoryPicker` no longer fall through to judgment-file validation
- Caught non-`AbortError` exceptions in `handleSelectedItems()` to prevent unhandled promise rejections from fire-and-forget calls

**Bug fix (`long-path-check.ts`)**
- Unreadable file entries now report `FileCheckStatus.Unreadable` instead of incorrectly using `LongPathIssue`, consistent with directory handling

**Downstream updates**
- `clientfilemetadataupload`, `clientfileprocessing`, `s3upload`, `clientfileextractmetadata`, and `upload/index.ts` updated to use the new types
- Test mocks updated to satisfy the new `IFileEntry` shape (added `kind` field, added `[Symbol.iterator]` to `FileList`/`DataTransferItemList` mocks)


npm/src/upload/form/upload-form.ts
```diff
@@ -163,12 +169,11 @@ export class UploadForm {
   }
 
   handleSelectedItems: () => any = async () => {
-    // Firefox and Safari do not support the directory picker API, so we need to check if it is supported before calling it.
-    // Can still upload on these browsers but long path check will not be executed
-    // TDRD-1698
-    if (!this.isJudgmentUser && supportsDirectoryPicker()) {
+    if (!this.isJudgmentUser) {
       try {
```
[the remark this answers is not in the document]
ian-hoyle: Different paths for different browsers are not wanted

npm/src/upload/form/upload-form.ts
```diff
@@ -133,11 +131,11 @@ export class UploadForm {
         const webkitEntry = droppedItem.webkitGetAsEntry()
         const resultOrError = this.checkIfDroppedItemIsFolder(webkitEntry)
         if (!isError(resultOrError)) {
-          const filesAndFolders: (IFileWithPath | IDirectoryWithPath)[] =
-            await getAllFiles(webkitEntry as unknown as IWebkitEntry, [])
-          const files = filesAndFolders.filter((f) =>
-            isFile(f)
-          ) as IFileWithPath[]
+          const filesAndFolders: IEntry[] = await getAllFiles(
+            webkitEntry as unknown as IWebkitEntry,
```
jlhdxw: This doesn't look right - `filesAndFolders` is typed as a list of `IEntry`, which is `IFileEntry | IDirectoryEntry`, however we're typing each item as `IWebkitEntry` which aren't linked in any way. Does `IEntry` need to become `type IEntry = IFileEntry | IDirectoryEntry | IWebkitEntry` ?

npm/src/upload/form/upload-form.ts
```diff
@@ -133,11 +131,11 @@ export class UploadForm {
...
+            []
+          )
+          const files = filesAndFolders.filter((f) => isFile(f)) as IFileEntry[]
```
jlhdxw: We shouldn't need to cast the result of this filter to IFileEntry as the predicate function should give the type for you.

Also you can write this like this: `const files = filesAndFolders.filter(isFile)`

npm/src/upload/form/upload-form.ts
```diff
@@ -419,13 +415,20 @@ export class UploadForm {
...
-  private convertFilesToIfilesWithPath(files: File[]): IFileWithPath[] {
+  private convertFilesToEntries(files: File[]): IEntry[] {
     this.checkIfFolderHasFiles(files)
 
-    return [...files].map((file) => ({
-      file,
-      path: (file as FileWithRelativePath).webkitRelativePath
-    }))
+    return [...files].map((file) => {
+      const fileWithPath = file as unknown as { file?: File; path?: string }
```
jlhdxw: `fileWithPath` should already be typed as `File` - does the `File` type not interently have `file` and `path` attributes? We could add another predicate function to ensure it does have a `file` and `path`.

Also `map` doesn't mutate what it's called on, so we shouldn't need to spread/copy the files variable. You should be able to just do `return files.map ...`

npm/src/upload/form/upload-form.ts
```diff
@@ -419,13 +415,20 @@ export class UploadForm {
...
+      return {
+        file: fileWithPath.file || file,
+        path:
+          (file as FileWithRelativePath).webkitRelativePath ||
```
jlhdxw: We could do a type predicate here instead of casting to ensure `file` has `webkitRelativePath`

npm/test/upload-form-utils/mock-files-and-folders.ts
```diff
@@ -50,8 +54,49 @@ export const getDummyFile: (fileName?: string, fileType?: string) => File = (
...
+
+export function createMockDirectoryHandle(
+  folderName: string,
+  files: { name: string; file: File }[]
+): IFileSystemDirectoryHandle {
+  const fileHandles: [string, IFileSystemFileHandle][] = files.map(
+    ({ name, file }) => [
+      name,
+      {
+        kind: "file" as const,
```
jlhdxw, line 71: We should update these instances to use the `EntryKind` that was added

npm/test/get-files-from-directory-picker.test.ts
```diff
@@ -1,9 +1,11 @@
...
-  supportsDirectoryPicker
+  supportsDirectoryPicker,
+  IFileSystemDirectoryHandle,
+  IFileSystemFileHandle
 } from "../src/upload/form/get-files-from-directory-picker"
 
-function createMockFileHandle(name: string, file: File) {
+function createMockFileHandle(name: string, file: File): IFileSystemFileHandle {
   return {
     kind: "file" as const,
```
jlhdxw, line 11: There are some more instances that could do with the `EntryKind` being used

npm/test/upload-form-utils/mock-files-and-folders.ts
ian-hoyle: keep kind: "file" here. This object is an IFileSystemFileHandle, which mirrors the browser File System API and uses literal kinds ("file"/"directory"). EntryKind is for our app-level IEntry/IFileEntry models, so mixing them here would blur that boundary.”

npm/src/upload/form/upload-form.ts
ian-hoyle: IEntry should stay as our normalized domain type (IFileEntry | IDirectoryEntry) and should not include raw browser entries.
 I’ve fixed this at the boundary instead: we now type/narrow the webkitGetAsEntry() result locally (with a directory-entry type guard) before passing it to getAllFiles, and removed the as unknown as IWebkitEntry cast. This keeps browser input types and app output types properly separated

jlhdxw (approved):

TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):
Looks good. Assuming this is the re-reversion of these changes back into intg. Thanks

PR #5179, opened 2026-07-28
TDRD-1701 check draftmetadataupload status before loading draftmetadataupload page

thanhz:


app/controllers/DraftMetadataUploadController.scala
```diff
@@ -13,6 +13,9 @@ import play.api.libs.Files.TemporaryFile
 import play.api.mvc._
 import services.Statuses._
 import services._
+import uk.gov.nationalarchives.tdr.common.utils.statecontrol.{CurrentState, TransferState}
+import uk.gov.nationalarchives.tdr.common.utils.statuses.StatusTypes.{DraftMetadataUploadType => CommonDraftMetadataUploadType}
+import uk.gov.nationalarchives.tdr.common.utils.statuses.StatusValues.{InProgressValue => CommonInProgressValue}
```
TomJKing, line 18: Assuming used the alias for the in progress value to avoid naming conflicts else where until this task is completed: https://national-archives.atlassian.net/browse/TDRD-1346?

TomJKing (commented):
Looks good. Just one minor question. Thanks

app/controllers/DraftMetadataUploadController.scala
thanhz: Yep. There's a method consignmentStatusService.getStatusValues that currently takes a statusValue from the frontend Statuses class instead of the common utils one and it's used in many places so just using an alias for now.

TomJKing (approved):
Looks good. Thanks

PR #5180, opened 2026-07-29
Reapply "Revert file path check feature (temporary - reapply after em…

ian-hoyle:
For a previous bug in sorting reviews file path check code was removed.
After deployed the file path check code was reverted back
This now need to be removed


vimleshtna (approved):
Looks good. Thanks

PR #5182, opened 2026-07-30
TDRD-1731 - Check all consignment statuses to verify backend checks completion

vimleshtna:
Tested this twice on staging by uploading 5,000 files with redacted files, and I haven't seen the issue occur anymore.


npm/src/checks/get-checks-progress.ts
```diff
@@ -70,7 +71,14 @@ export const getFileChecksProgress: () => Promise<
     const response = progress as Consignment & { backendChecksFailed?: boolean }
     if (response) {
       const fileChecks = response.fileChecks
+      const isBackendChecksCompleted =
+        response.consignmentStatuses?.every(
+          ({ value }) =>
+            value === "Completed" || value === "CompletedWithIssues"
```
TomJKing: Can also have a `Failed` value, though not sure if this is being used in relation to the consignment statuses.

If each status is be set to `InProgress` might be safer to change the test to `value !== "InProgress"`

Would need to double check the how the statuses are being set initially to see if that would work. If not to be safe might be worth adding the `Failed` status as part of the check: `value === "Completed" || value === "CompletedWithIssues" || value === "Failed"`

TomJKing (commented):
Looks good. Thanks for looking into this. Just one minor thought.

npm/src/checks/get-checks-progress.ts
vimleshtna: We do set the `Failed` value at the consignment level - https://github.com/nationalarchives/tdr-statuses/blob/main/src/main/scala/uk/gov/nationalarchives/StatusProcessor.scala#L153

Also, when the backend checks start, it does set the `InProgress` value for all the backend checks related status types. So, I think it makes sense to check `value !== "InProgress"`.

I have updated it now. Thanks

TomJKing (approved):
Looks good. Thanks

